PR #17, opened 2023-10-04
Added insight on frontend

KenjiPcx:
We now have a cool AI card thinking of insights
image


web_frontend/src/types.ts
```diff
@@ -1,7 +1,8 @@
 export type Entity = {
-  name: string;
-  summary: string;
+  name?: string;
```
nic-olo, line 2: Shouldn't the name be non-nullable?

KenjiPcx: no, insights have no name in the object right now

KenjiPcx:
Reverted and upscaled the logo

PR #25, opened 2023-10-08
Optimized prompt and set hard limit to agent as guard rail

KenjiPcx:
- Optimized prompt to limit tool usage to just 1, making it ~5s faster
- Added a max iteration for the agent as a guard rail in case it decides to go in an infinite loop


nic-olo:
LGTM

PR #34, opened 2023-10-20
Multi agent explicit

alex1115alex:


server/DatabaseHandler.py
```diff
@@ -202,7 +234,7 @@ def get_new_cse_transcripts_for_user(self, user_id, delete_after=False):
                     backslide_words = ' '.join(backslide_word_list[-(self.backslide-len(backslide_word_list)):])
 
                     words_from_start = first_transcript['text'][start_index:].strip()
-                    first_transcript['text'] = backslide_words + " " + words_from_start
+                    first_transcript['text'] = backslide_words + " " + words_from_start if words_from_start else backslide_words
```
KenjiPcx, line 237: didn't know python would parse "" as false and "123" as true, not really glaring but could be more explicit to improve readability

server/agents/explicit_meta_agent.py
```diff
@@ -20,7 +20,7 @@
 <Transcript start>{conversation_context}<Transcript end>
 
 [Your Task]
-Now use your tools and knowledge to answer the query to the best of your ability. The question or request you are to answer is the last (final) question/request posed by the human to you in the below 'Query'. Make your answer as concise and succinct as possible. Leave out filler words and redundancy to make the answer high entropy and as to-the-point as possible. Never answer with more than 240 characters, and try to make it even less than that. Most answers can be given in under 10 words.
+Now use your knowledge and/or tools (if needed) to answer the query to the best of your ability. Do not use your tools if you already know the answer to the query. The query may accidentally contain some extra speech at the end, you should ignore any noise and try to find the user's inteded query. Make your answer as concise and succinct as possible. Leave out filler words and redundancy to make the answer high entropy and as to-the-point as possible. Never answer with more than 240 characters, and try to make it even less than that. Most answers can be given in under 10 words.
```
KenjiPcx, line 23: line 38 below, I was wondering if we could move the new_expert_agent out of the function and define it somewhere in the file instead, I think it is fine to reuse the same generic agent instead of creating one everytime the explicit agent is called

server/agents/explicit_query_process.py
```diff
@@ -39,11 +39,13 @@ def explicit_query_processing_loop():
...
 
-                # If there has been a pause
-                elif (len(user['final_transcripts']) != 0) and (current_time > user['final_transcripts'][-1]['timestamp'] + pause_query_time):
+                # If there has been a pause (the latest transcript is old)
+                latest_transcript = dbHandler.get_latest_transcript_from_user_obj(user)
+                if latest_transcript and (current_time > latest_transcript['timestamp'] + pause_query_time):
                     is_query_ready = True
 
                 if not is_query_ready: continue
+
```
KenjiPcx, line 48: line 85 below, where the insight gets generated, is it assumed that it would be like "Insight: this is an insight"? because in the other branch i changed insights to return a dict of { insight: str, resource_url: https://<link>, motive: str }

KenjiPcx: ill test the code later, and will add some changes if it doesnt work out of the box

PR #35, opened 2023-10-24
Added examples and history of previous insights into agents

KenjiPcx:


KenjiPcx:
Will open a new PR to dev

PR #36, opened 2023-10-30
Realtime transcripts test new

KenjiPcx:
Added a script that sends transcripts based on timestamps in real time

It sends the transcript in real time and gets the responses to the api calls
image


KenjiPcx:
I am opening a PR into Alex's branch because it had the most updated testing code

KenjiPcx:
I'm going to merge this into #34 because the only thing new is a test file and we can avoid debugging merge conflicts

PR #40, opened 2023-11-04
Explicit query support for Android TPA

alex1115alex:


CaydenPierce:
The smart glasses should say "listening..." when a wake word is detected.

Then they should say "Processing query: {query}"

Then when the answer comes, it should say: "Response: {convoscope_answer}"

CaydenPierce:
Basically 2 things.

1. Small thing - show a "Listening..." screen when wake word is shown.
2. Slightly larger things -  the ui_poll is now too complex to not have memory on the frontend. This leads to annoying things like sometimes a result is only shown for very short time as a new result comes in and overwrites it. We need to hold in memory all of the queries, results, insights, definitions, etc and then implement some timers/buffer in the Convoscope app. Keep it very simple for now should be enough. Either me or @alex1115alex should do it

CaydenPierce:
Also let's not hardcode "Mira" in the frontend - "Mira" is a just a standin name for now...

PR #46, opened 2023-11-13
Frontend Redesign

reeceyang:
image
image
image


nic-olo:
WhatsApp Image 2023-11-19 at 22 44 12_a3611dad


When a card is clickable, the cursor should change to the hand icon. This provides a clear indication to the user that the card contains a link. If the card doesn't have a link, then the cursor should stay as the standard arrow.

alex1115alex:
Hey Reece, first - thanks a ton for doing this for us. It looks **awesome**!

I've only found a couple of small problems:

1. It would be great to be able to highlight a card's text in order to copy and paste. While you **can** technically do this as-is, doing so will trigger the relevant iFrame to open/close. This was already an issue on the old frontend however, so no worries if you can't get to this.

2. The location of the iFrame's "close" button seems a little off. Have you tried sticking it on the left side or above the iFrame (putting the "close" button in a bar above the iFrame would give us a nice space to include a share/copy-link button in the future)? See: 
image
Feel free to disregard this if you've already tried different positioning of the close button... it'll always look a little weird with the iFrame's contents being disconnected from our theme.

KenjiPcx:
Hey @reeceyang, nice work with the new frontend, looks fantastic

For me, I'm having issues with the display of certain logos 
image

image

Maybe we could fix the image aspect ratios so it doesnt gets resized when the read more section appears, if the image is too big, we can also clip images so that they render nicely

KenjiPcx:
Found some more minor things
- Should we stop Mira when we click the stop button, I don't know how to stop it
image
- This scroll thing on all the cards is visible, we could hide it 
image

KenjiPcx (approved):
Code looks good to me, I've added some comments on the visual side

KenjiPcx:
Is it possible to make the font size dynamically shrink based on responsiveness? because when the explore more panel appears the words of the card exceeds its margins

image

The 'y' above looks like its about to exceed the card area

reeceyang:
>     1. It would be great to be able to highlight a card's text in order to copy and paste. While you **can** technically do this as-is, doing so will trigger the relevant iFrame to open/close. This was already an issue on the old frontend however, so no worries if you can't get to this.

I think the challenge here is that it's hard to tell if a user is trying to select text or to open the iFrame. Would adding a separate copy button (probably next to the thumb buttons) be a good workaround? The button could copy all the text on the card.

>     2. The location of the iFrame's "close" button seems a little off. Have you tried sticking it on the left side or above the iFrame (putting the "close" button in a bar above the iFrame would give us a nice space to include a share/copy-link button in the future)? See:

I think putting it on the left side makes a lot of sense! I also added a persistent open iFrame button to stay consistent. A share/copy button could maybe go underneath the close button.


> Code looks good to me, I've added some comments on the visual side

Good catches with the visual issues, I'll push some fixes.

alex1115alex:
Another quick visual thing:

What do you think about centering the explicit query results? They're currently pushed to the top left of the card, which looks a little awkward when compared to the "listening" or "query" card. Here's what I mean.

Listening:
image

Query:
image

Result:
image

CaydenPierce:
- [ ] make the settings panel aesthetic match the rest (mostly color scheme)
- [ ] make the settings panel "set username" look like a button
- [ ] not sure I agree with this new iframe close button - took me a while to find it. Usability matters most here. Let's see what others think though
- [ ] stop button and the background animation look fuzzy/low-resolution
- [ ] "I'm listening" animation/video looks very weird as it's cropped to a circle
- [ ] closing the explore iframe causes a seizure to occur with all the elements of the page. 
- [ ] opening the iframe causes text to distort and such
- [ ] incoming card animation isn't smooth - comes in, slows, then speeds up again - should be smooth/continuous


Animations are important because they're jarring right now - i.e. I'd rather have no animation over animations that distract and cause discomfort.

CaydenPierce:
Screenshot from 2023-11-23 00-44-58

Why is there a newline after "Joe Biden:" - wrapping should happen at end of line

PR #51, opened 2023-11-28
Frontend visual fixes

reeceyang:
- [x] better card and card image aspect ratio sizing
- [x] if there's no image, don't show an image
- [x] only show three cards - size margins/padding using viewport
- [x] make the settings panel aesthetic match the rest (mostly color scheme)
- [x] make the settings panel "set username" look like a button
- [x] "I'm listening" animation/video looks very weird as it's cropped to a circle
- [x] closing the explore iframe causes a seizure to occur with all the elements of the page.
- [x] incoming card animation isn't smooth - comes in, slows, then speeds up again - should be smooth/continuous
    - [x] explicit card should have command label
    - [x] "listening" image should persist
- [x] text sizing on high resolution screens


reeceyang:
We can squash and merge this PR

PR #54, opened 2023-12-02
Move frontend state into Recoil atoms

reeceyang:
This PR moves frontend state into Recoil atoms. The new atoms are the transcription state, the entities state, the explicit insights state, and the explicit listening state. The `useEffect`s for updating the transcription state and the UI update backend poll were also moved into their own hooks to help separate state logic from app layout.

The PR also reorganizes some of the folder structure - moving some functions into `/utils`, and creating a new `/layouts` folder. In addition, a notification sound plays when receiving new entities and/or insights (which can be configured in the new `constants.ts` file).

Everything should behave exactly the same as before + if the explicit agent card is shown, pressing the stop button with hide the card.

I'm hoping to build off of this PR to further refactor entity state handling. In particular, I'd like to introduce a new `Cards` atom that keeps track of each card currently displayed in the app. This will help make the code more modular and makes some features more feasible (such as having the explicit card transform into a reference card).


reeceyang:
Let's use squash and merge for this PR

PR #61, opened 2023-12-18
Study frontend

reeceyang:
image

Stores the current video time (as number of seconds elapsed from the start) in `videoTimeAtom`. If the video hasn't begun yet, the video time is `undefined`.

The video is an mp4 file loaded in from the `/public` directory, and you can configure the video source in `constants.ts`.

The tab changes are tracked in the `tabChangesAtom`


web_frontend/src/App.tsx
```diff
@@ -1,14 +1,12 @@
...
 import { generateRandomUserId, setUserIdAndDeviceId } from "./utils/utils";
 import Cookies from "js-cookie";
-import MainLayout from "./layouts/MainLayout";
+import StudyLayout from "./layouts/StudyLayout";
+import { useTrackTabChange } from "./hooks/useTrackTabChange";
 
 export default function App() {
-  useTranscription();
-  useUiUpdateBackendPoll();
+  useTrackTabChange();
```
nic-olo, line 10: Could you use a variable to disable the backend bits instead of hardcoding the changes please?

reeceyang: That makes sense, just pushed a change for this.

CaydenPierce: This is fine where it's easy, but in general don't lose any time making this backwards compatible - we will make some quick iterations on the frontend during first pilots of the study. Our data and learning at the end will influence what we use/don't use, and we can properly integrate those developments in after the study.

PR #65, opened 2023-12-22
Study backend

nic-olo:


CaydenPierce:
To merge needs:

- images tested and working
- verification def'n delay is ~<4 seconds, fix if not
- specify the name of vtt to use to generate

PR #70, opened 2024-01-02
add custom google search condition

reeceyang:
image

The Google Programmable Search component has to be set to "Two Page" for the component to display correctly.

image


CaydenPierce:
Major issue here is that clicking the links in the search engine results opens them as a page. Can we open them in our own iframe?

reeceyang:
Was able to do it with a hacky workaround.

image

PR #76, opened 2024-01-05
Speed up explicit queries (prompt change)

alex1115alex:
Adjust explicit agent prompt so it does not use agents unless absolutely necessary. Also reduce query pause time.

TODO (perhaps future PR): Inform user that a query will take a long time // has multiple steps

(EXAMPLE, not the only query tested) 

The following query would previously defer to agents + search, thus taking a long time:
"Hey Mira, how much wood could a woodchuck chuck if a woodchuck could chuck wood?"

Now:

https://github.com/TeamOpenSmartGlasses/Convoscope/assets/27019702/55a50fd5-a1e7-4c77-94e6-3da4c7886f63


CaydenPierce:
Not the change we need. Redone in other PR

PR #80, opened 2024-01-07
don't open falsy links

reeceyang:


CaydenPierce:
Just a note for the future - avoid ' in branch name, it makes it unweildy to deal with in the CLI and can break some git scripts.

PR #81, opened 2024-01-08
Quick proactive definer + Cool Stats 

alex1115alex:
- Decrease delay between proactive definer process runs, but also run on shorter transcripts
- Proactive definer Gatekeeper (mad cost savings)
   - GPT3.5 checks if some text is worth running on, as well as returns the term that most heavily impacted its decision 
   - GPT4 select+define as usual
   - If GPT4 doesn't select anything, the term GPT3.5 returned is added to the user's list of irrelevant words. Each user has their own list.
- Add stat tracker for proactive definer, see below

Stats can be found in `definer_stats/stats/{file.json}`. Example:

From an 8 minute transcript... (OG method: GPT4 only, 20s transcripts run every 10s, fat delays)
```
"all": {
        "size": 42,
        "average_time_taken": 10.591442108154297,
        "median_time": 9.107953429222107,
        "cost": 1.6463100000000002
    },
```

From a 50 minute transcript... (Current Gatekeeper system, 4s transcripts run every 2s, much shorter delays)
```
"time_elapsed": "0:50:59",
    "all": {
        "size": 826,
        "average_time_taken": 10.369993638992309,
        "median_time": 9.093019008636475,
        "gatekeeper_accuracy": 52.63157894736842,
        "cost": 1.4182775000000003
    },
```


CaydenPierce:
We should merge this as-is and focus on other things, as the Definer works pretty damn well, and other things need more attention.

In the future, improvements to this would be:
- better gatekeeper prompt. Prompts need to tell the LLM what to do and how. The current gatekeeper prompt doesn't explain that this is a conversation and that we need to define things that people won't know if and only if those definitions will improve their conversation. Calling a term "interesting" means nothing - be more specific.
- the output of the gatekeeper should be well-formed (JSON) and the example in the prompt should follow that form. There should be a score per term, not for the list of terms. If you gave me a string and that prompt, I wouldn't know what you wanted me to output
- the irrelevant terms should decay based on time, not number. We went through this with the other histories. A term that wasn't worth defining in a previous conversation may be worth defining now.

Again, this is a note for learning and if we return to this to improve it. I think we should merge this and move on.


@alex1115alex

CaydenPierce:
Actually, this is missing stuff, hold off on merge.

PR #96, opened 2024-02-06
Update dependencies

nic-olo:


web_frontend/package.json
```diff
@@ -37,10 +40,9 @@
     "@types/regenerator-runtime": "^0.13.1",
     "@typescript-eslint/eslint-plugin": "^6.0.0",
     "@typescript-eslint/parser": "^6.0.0",
-    "@vitejs/plugin-react-swc": "^3.3.2",
+    "@vitejs/plugin-react": "^4.2.1",
```
reeceyang, line 43: I think we were using SWC for faster builds, are you able to get it to work using SWC?

reeceyang (changes requested):
Try `pnpm install` instead of `npm install`; that might solve some of the issues. We should probably delete `package-lock.json` as well. If `plugin-react-swc` works we should try to bring it back but if not it's not a big deal.

We can still merge since it's good to update some of the dependencies.

reeceyang (approved):

PR #102, opened 2024-02-18
Convoscope is now a multi-experience app

CaydenPierce:
Add multi-mode selection

Merge together language learning and existing stuff

TODO after merge: save user mode to backend and don't run any agents/processes that user isn't currently viewing.


alex1115alex:
works on my machine 
AlexFieri

PR #107, opened 2024-03-08
Word rank edits

lars-barkman:


server/agents/helpers/word_frequency_percentiles.py
```diff
@@ -22,21 +22,26 @@ def load_frequency_list(language_code: str) -> pd.DataFrame:
...
     df['rank'] = df['frequency'].rank(ascending=False)
     return df
 
 def get_word_frequency_percentiles(transcript: str, language_code: str) -> dict:
-    df = load_frequency_list(language_code)
+    df = load_frequency_list(language_code, '50k')
     words_dict = pd.Series(df['rank'].values,index=df['word']).to_dict()
     total_words = len(df)  # Get the total number of words in the dataset
 
+    df = load_frequency_list(language_code, 'full')
```
nic-olo, line 37: 50k will have us covered 99% of the time. Therefore, I suggest you only load up the full frequency list as a last resort, i.e. after you can't find a word.

CaydenPierce: I agree, and more so, we should only load this all once. I added that in my commit.

CaydenPierce:
@lars-barkman thanks for this. Try to remember to remove extraneous debug prints before PR

PR #122, opened 2024-06-09
Add languages to ll

SusannaShu:
added french language learning


server/server_config.py
```diff
@@ -40,7 +40,7 @@
 # Prod:
 database_uri = "mongodb://localhost:27017"
 server_port = 8080
-path_modifier = ""
+path_modifier = "/dev4"
```
nic-olo, line 43: Same as above.

CaydenPierce:
Is this PR going to add the other languages as well?

Let's name PRs better to explain what they're doing.

CaydenPierce:
@SusannaShu 

Did you test all the languages to make sure the dictionaries work on the backend?

The build.gradle change is still in here, please remove it.

Otherwise looks good!

CaydenPierce:
Also, there needs to an SGM commit that adds the new languages to the ASR files.

PR #123, opened 2024-06-16
Add IdeaGenerator agent

alex1115alex:


nic-olo:
base branch should be dev, not main

PR #132, opened 2024-07-18
ll upgrade backend + frontend

SusannaShu:
fixed some issues then wrote ll word suggest upgrade backend
Connect front end with backend


server/agents/ll_word_suggest_upgrade_agent.py
```diff
@@ -0,0 +1,223 @@
...
+#Use Pinyin when writing Chinese. Never use Chinese characters, use Pinyin.
+#, use Pinyin if writing Chinese)
+#3.a. If writing Chinese, output exclusively in Pinyin, avoiding Chinese characters entirely.
+#NEVER OUTPUT CHINESE CHARACTERS. USE PINYIN, USE THE LATIN ALPHABET PINYIN FOR CHINESE.
+
+#Use Pinyin when writing Chinese. Never use Chinese characters, use Pinyin.
+
+# in_upgrade_translation must be in Pinyin or Latin characters.
+
+# If writing Chinese, output exclusively in Pinyin, avoiding Chinese characters entirely.
```
nic-olo, line 105: There's a lot of extra comment that I assume come from the original ll file, you can remove them from here they aren't needed

CaydenPierce:
Let's next time remove copious debug logs before PR @SusannaShu 

Vocabulary upgrade on the backend is set as false - when it's set as true on the frontend. Have to toggle on and off to get it to work?

CaydenPierce:
Why do we restart the service on changing the upgrade setttings?

PR #134, opened 2024-07-23
Add gpt 4o mini Option

nic-olo:


CaydenPierce:
Tested for LL rare words:

Average gpt-3.5-turbo times: 0.7661092758178711
Average gpt-4o times: 0.7191376924514771
Average gpt-4o-mini times: 0.8695436477661133
Average Groq times: 3.880943703651428

I'm sure mini is better if you have more in/out tokens. But latency, doesn't seem better. Ran twice, 4o won both times.

PR #145, opened 2024-11-03
Removed gradient from the last card, fixed title sizing, space in between connceted text and glasses name divider

DDev16:


alex1115alex:
epic

PR #183, opened 2025-01-30
Update EvenRealitiesG1SGC.java

fuutott:
adding taps handling for even g1 - untested


CaydenPierce:
Thanks!
Taking a look...

PR #368, opened 2025-04-13
begin mentra live

alex1115alex:


alex1115alex:
At this stage...

StandardBluetoothManager: 

Comms from Phone=>Glasses BROKEN, Comms from Glasses=>Phone WORKS

K900BluetoothManager:

Comms from Phone=>Glasses WORKS, Comms from Glasses=>Phone BROKEN

PR #372, opened 2025-04-14
iOS Bug fixes and improvements

nirbhaysinghnarang:
This PR:

* Makes the sensing switch flow more explicit. Concretely, when sensing is turned off, a warning is displayed so that the user knows that microphone access won't be given (and apps like live captions won't work as expected)

* Ensures that the app falls back to phone mic if glasses aren't connected. Concretely, if glasses (G1s) aren't connected **(assuming that g1manager.isReady is a reliable proxy for this)** , toggling the force phone mic switch has no effect and audio is piped in from the phone mic only

* Fixes iOS jitters on Login Screen.


alex1115alex:
NICE! 🥳

alex1115alex:
Nice! One quick thing, can you please base this off dev and not ios-dev? ios-dev was merged to dev quite a few commits ago, and now there are some conflicts.

nirbhaysinghnarang:
Merged dev in and updated base.
But it seems like most of these changes were already committed to dev by other contributors so might want to close this PR now.

PR #378, opened 2025-04-14
Add support for setting custom backend url in developer settings

OkGoDoIt:
This allows us to test development and staging cloud environments without having to recompile the app.

Screenshot (Apr 13, 2025 11_43_15 PM)

Screenshot (Apr 13, 2025 11_43_31 PM)


alex1115alex:
The reason we haven't implemented this is because it's a little more involved. The .env vars in `augmentos_manager/.env` end up getting passed in to the Android/iOS native modules, and access by them. Those native modules are where the websocket comms are happening. So while changing the serverUrl here would affect the REST requests coming from typescript code, it wouldn't chance where the websockets are pointed to. Would need a way of passing these custom URLs down to the native modules

PR #393, opened 2025-04-17
connection / reconnection updates and fixes

fossephate:


fossephate:
@alex1115alex if you could review this when you get the chance :D

PR #400, opened 2025-04-21
Tpa auto auth

### LINKED ISSUE #390: Auth for webviews
App webviews don't have a way of authenticating with AugmentOS right now, meaning they can't be personalized per-user.

OkGoDoIt:
Closes #390 
Provide a secure and straightforward mechanism for third-party web applications, loaded within the AugmentOS manager application's webview, to identify the current AugmentOS user. This enables personalized experiences within the third-party context without requiring the user to log in separately to that service.

- Authentication flow for web applications loaded via the integrated webview within the AugmentOS manager app.
- Generation of short-lived, temporary tokens within the AugmentOS ecosystem.
- Passing the temporary token via a URL query string parameter.
- A secure server-to-server cloud endpoint for exchanging the temporary token and developer API key for a validated User ID.
- SDK automatically handles this for developers.

For an example implementation from a TPA developer's point of view, see https://github.com/AugmentOS-Community/AugmentOS-Reminders/pull/2

## Webview Query String Token Exchange

This solution leverages a short-lived temporary token passed in the URL, which is then securely exchanged server-side for user identification.

### User Experience

1.  The user interacts with the AugmentOS manager app (e.g., taps the gear, long-presses the app in the app list) 
2.  The manager app opens an integrated webview, loading the specified third-party URL.
3.  The third-party web application loads and, through the mechanism described below, recognizes the AugmentOS user, potentially displaying personalized content. The user does not need to manually enter credentials.

### Detailed Technical Flow

Image

1.  **Trigger:** An action within the AugmentOS manager app initiates the loading of a third-party URL into a webview.
2.  **Temporary Token Request (App -> AugmentOS Backend/SDK):** Before loading the URL, the manager app requests a single-use, temporary authentication token from a secure AugmentOS service. This service generates a cryptographically secure, opaque token associated with the current User ID and the target developer (identified implicitly or explicitly). The token has a short Time-To-Live (TTL).
3.  **Token Generation & Storage (AugmentOS Backend):** The AugmentOS backend generates the token (e.g., a UUID or secure random string), stores it temporarily (e.g., in Redis/Memcached) along with the associated User ID, the intended developer's identifier (e.g., their App ID or API Key ID), and an expiry timestamp (e.g., 60 seconds from creation).
4.  **URL Construction (App):** The manager app receives the temporary token and appends it as a query string parameter to the target URL.
    *Example:* `https://partnerapp.com/dashboard?aos_temp_token=TEMP_TOKEN_ABC123XYZ`
5.  **Webview Load (App):** The manager app loads the modified URL into the integrated webview component.
6.  **Request Reception (Developer Backend):** The developer's web server receives the incoming GET request for the URL.
7.  **Token Extraction (Developer Backend):** The developer's *server-side* application code extracts the value of the `aos_temp_token` parameter from the request URL.
8.  **Token Exchange (Developer Backend -> AugmentOS Cloud API):** The developer's backend makes a *server-to-server* API call to a dedicated AugmentOS cloud endpoint (e.g., `POST https://prod.augmentos.cloud/api/auth/exchange-user-token`). This request MUST include:
    *   The extracted `aos_temp_token`.
    *   The developer's secret API key (provided via a secure mechanism, typically an `Authorization` HTTP header like `Authorization: Bearer <developer_secret_api_key>`).
9.  **Validation (AugmentOS Cloud API):** The AugmentOS `api/auth/exchange-user-token` endpoint performs the following checks:
    *   Authenticates the developer using the provided API key.
    *   Retrieves the temporary token data from its temporary store.
    *   Verifies that the token exists, has not expired, and has not already been used.
    *   Verifies that the token was intended for use by the authenticated developer (matching API key/App ID).
10. **Mark Token Used (AugmentOS Cloud API):** If validation succeeds, the endpoint immediately marks the temporary token as used in the temporary store to prevent replay attacks.
11. **Response (AugmentOS Cloud API -> Developer Backend):** If validation is successful, the API responds with a success status (e.g., `200 OK`) and a JSON payload containing the authenticated AugmentOS User ID.
    *Example Success Response Body:* `{ "user_id": "AUG_USER_987654321" }`
    *Example Error Responses:* `401 Unauthorized` (invalid API key), `404 Not Found` (token doesn't exist), `410 Gone` (token expired or already used), `403 Forbidden` (token not intended for this developer), `400 Bad Request` (malformed request).
12. **Session Creation (Developer Backend):** The developer's backend receives the User ID. It can now trust that the request originated from the specified AugmentOS user within the app's webview. The backend uses this User ID to:
    *   Look up or create a corresponding user account in its own system.
    *   Establish a standard web session for the user (e.g., by setting a secure, HttpOnly session cookie).
13. **Personalized Response (Developer Backend -> Webview):** The developer's server renders and returns the web page, now tailored or authenticated for the identified user.

### Component Responsibilities

*   **AugmentOS Manager App:** Responsible for initiating the flow, requesting the temporary token, constructing the URL, and loading the webview.
*   **AugmentOS Cloud/Backend:** Responsible for generating, storing, and validating temporary tokens; providing the secure `api/auth/exchange-user-token` endpoint; authenticating developers via API keys; returning the User ID.
*   **Developer Backend:** Responsible for extracting the token from the URL, securely storing its API key, calling the `api/auth/exchange-user-token` endpoint, handling the response, managing its own user sessions, and ensuring overall web application security.
*   **AugmentOS SDK:** Handles the developer side automatically (listens for the `aos_temp_token` and exchanges it on a developer's behalf)

## Implementation Details

*   **Temporary Token:**
    *   Format: Securely generated random string (e.g., 32+ bytes of randomness, base64 encoded) or UUID v4.
    *   Lifetime: Configurable, recommended default 60 seconds.
*   **Cloud Endpoint (`/api/auth/exchange-user-token`):**
    *   Standard RESTful API principles.
    *   Requires `POST` method.
    *   Expects `Authorization: Bearer <developer_secret_api_key>` header.
    *   Expects JSON body: `{ "temp_token": "value" }`.
    *   Returns JSON responses with appropriate HTTP status codes.
*   **SDK:**
    *   Handles the developer side automatically (listens for the `aos_temp_token` and exchanges it on a developer's behalf)

## Security Considerations

This section addresses the security implications and required mitigations:

*   **Query String Exposure:**
    *   *Risk:* The temporary token is visible in the URL, potentially logged by browsers, proxies, or server logs.
    *   *Mitigation:* This is acceptable the token meets all the following criteria:
        1.  **Extremely Short-Lived:** Limits the window for interception and use (e.g., 60 seconds).
        2.  **Opaque:** Contains no sensitive user or session information itself. It's merely a pointer.
        3.  **Requires Server-Side Secret:** Useless without the developer's secret API key, which is *never* exposed client-side. The exchange *must* happen server-to-server.
        4.  **HTTPS:** End-to-end TLS encryption protects the token (and the rest of the request/response) in transit.
*   **API Key Security:**
    *   *Risk:* Compromise of a developer's API key would allow an attacker (if they could also obtain temporary tokens) to impersonate users within that developer's application.
    *   *Mitigation:* This is the most critical aspect for the developer to manage.
        1.  **Strict Server-Side Use:** Keys must *never* be embedded in client-side code (JavaScript, mobile app binaries).
        2.  **Secure Storage:** Keys must be stored securely on the developer's backend (e.g., environment variables, secrets management systems).
        3.  **Clear Documentation:** AugmentOS documentation must heavily emphasize these points.
*   **Replay Attacks:**
    *   *Risk:* An attacker intercepts a token and tries to reuse it.
    *   *Mitigation:* Enforced single-use nature of the token (backend invalidation upon first successful use) combined with the short expiry time.
*   **Transport Security:**
    *   *Risk:* Eavesdropping or modification of requests/responses.
    *   *Mitigation:* Mandate HTTPS for the webview load and the server-to-server `api/auth/exchange-user-token` API call.
*   **Token Predictability:**
    *   *Risk:* Attacker guesses valid tokens.
    *   *Mitigation:* Use a cryptographically secure random number generator (CSRNG) to create tokens with sufficient entropy (e.g., >= 128 bits). Do not use sequential or easily guessable identifiers.


OkGoDoIt:
If this looks good @alex1115alex and/or @isaiahb, it's ready to merge.  Please give it a review approval and I can merge it in.

augmentos_cloud/packages/sdk/src/tpa/server/index.ts
```diff
@@ -51,6 +52,24 @@ export interface TpaServerConfig {
...
+   */
+  cloudHostName?: string;
+  
+  /** 
+   * 🔗 Custom cloud URL for API endpoints
+   * Only needed when connecting to a non-default cloud environment or for testing
+   * If not provided, the SDK will use the default production URL
+   */
+  cloudUrl?: string;
+
```
isaiahb, line 67: One thing I previously was trying to do is remove the need for the developer to specify a cloudHostName or a cloudUrl.

For sessions I was able to accomplish this by having the cloud pass what it's public host / url was to the TpaServer through the webhook (making it so we can depricate "augmentOSWebsocketUrl?: string;" 

is there a way we can accomplish something similar for the auth/middleware, so developers can have 1 TPA that seamlessly works in different environments. 
happy to hop on a call to brainstorm this

isaiahb (approved):

augmentos_cloud/packages/sdk/src/tpa/server/index.ts
OkGoDoIt: @isaiahb Are there any scenarios where this would be anything other than the default prod.augmentos.cloud?  I would assume only internal developers like us would need to change this.  If it's something that normal 3rd party app developers don't ever have to do then I don't see it as being a problem. Generally I would expect we have a single cloud API url.  Any load balancing that happens should happen at the dns level.

isaiahb (approved):
LGTM Awesome work!!!

PR #453, opened 2025-05-14
Dashboard uses metric system

nic-olo:
@fossephate see the iOS implementation.

metricSystemEnabled needs to be added to the core_status along with the other augmentos settings (i.e., isConextualDashbaordEnabled, alwaysOnStatusBar etc.)


fossephate:
@nic-olo there's a sendSetUnitSystem function in the ios core communicator that looks like it isn't used / should be removed but otherwise looks good 👍

fossephate (approved):
there's a sendSetUnitSystem function in the ios core communicator that looks like it isn't used / should be removed but otherwise looks good 👍

PR #467, opened 2025-05-20
Add developer console organizations

OkGoDoIt:
Add all developers to organizations.  TPA's are now associated with an org rather than an individual user.  **There's a db migration script that MUST be run before merging this into main**: https://github.com/AugmentOS-Community/AugmentOS/blob/a04b9bb4499f5afc9fd8344e7f9d6ad3d052adb0/augmentos_cloud/packages/cloud/scripts/migrations/001-personal-orgs.ts


isaiahb (approved):
Very solid! Awesome job!

PR #469, opened 2025-05-22
Tpa auth client side

OkGoDoIt:
Add support for TPA webview user auth even in purely client-side webviews, by changing the the token exchange mechanism into a public key verifiable JWT with the user id.

Updates:
1. cloud project to create the token
2. manager app to use the new token in the webview (also send the old token type temporarily for compatibility)
3. sdk to consume the new token (keep the old token type temporarily for compatibility)
4. documentation to explain how to use with react or other frontend webview

**This requires a `TPA_AUTH_JWT_PRIVATE_KEY` to be added to the cloud .env**


isaiahb (approved):
This looks great! Awesome job 😄

OkGoDoIt:
I have fixed the issue with Android web views

> It looks like some older versions of android have a web view that doesn't support the JWT encryption. So I'm going to need to test this more before we merge it.

I also added a new @augmentos/react library that simplifies making react frontends.  The docs are also updated, see https://tpa-auth-client-side.augmentos-docs.pages.dev/react-webviews

isaiahb (approved):

PR #474, opened 2025-05-27
The home page has been updated with a new UI, including new icons and…

markiyurtdas:
… revised text fonts and sizes


fossephate (commented):
We should be replacing all of the styled sheets with themed() styles so that theming works
See the useAppTheme hook

Make sure to use the ignite components / use the tx prop so we get translations as well (you can add new strings in en.ts)

PR #511, opened 2025-06-14
added builder service for reliable docker builds

enley740:
Running "bun run dev" gave me a "cannot find module error" when I set up dev environment for the first time. I found that this was because services like cloud and dashboard-manager depend on shared local packages (e.g., @augmentos/utils, @augmentos/sdk) and the cloud service would try to start before its dependencies were fully built and available within the Docker container. 

This pr introduces a dedicated builder service whose only responsibility is to prepare the environment. It runs bun install to correctly link all workspaces and then builds the shared packages (sdk, utils, agents) in order. Updated the cloud and dashboard-manager services to have a "depends_on" condition so that they'll now explicitly wait for the builder service to exit with a service_completed_successfully status before they attempt to start.

I think this will improve the developer experience by providing a reliable setup for new contributors and it will also hopefully simplify future maintenance because it centralizes the build logic

### DIFF
--- a/augmentos_cloud/docker-compose.dev.yml
+++ b/augmentos_cloud/docker-compose.dev.yml
@@ -1,4 +1,21 @@
 services:
+  # this new service will run first. it installs dependencies and builds the shared packages
+  # that other services like 'cloud' and 'dashboard-manager' need to run.
+  builder:
+    build:
+      context: .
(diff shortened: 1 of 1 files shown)


isaiahb:
Thank you, this looks great! I just set the main branch to dev instead of main. I'll merge this in!

PR #697, opened 2025-07-10
docs: fix incorrect package name from @mentraos/react to @mentra/react

BulbasDanil:
The documentation incorrectly listed the required package as @mentraos/react. The correct package name is @mentra/react, as seen on npm and yarn:

https://www.npmjs.com/package/@mentra/react
https://classic.yarnpkg.com/en/package/@mentra/react

This affects npm, yarn and bun installation instructions, as well as the imports.


BulbasDanil:
The webpage with an error is https://docs.mentra.glass/react-webviews/#installation

OkGoDoIt:
good catch, thanks @BulbasDanil !

PR #723, opened 2025-07-14
Ya/local stt fallback

### LINKED ISSUE #657: Begin infra for local STT fallback
## Part 1

Whenever we currently send `set_microphone_state: true|false` to the client, also send:

`required_data: [pcm_or_transcripts, pcm, transcripts]`

`pcm_or_transcripts`: Apps are only requesting transcripts, so client can send either pcm or transcripts

`pcm`: Cloud requires pcm

`transcripts`: Cloud requires local transcripts (Cloud STT systems are offline)

Possible states: `[pcm], [transcripts], [pcm_or_transcripts], [pcm, transcripts]`

## Part 2

- [ ]  Create a websocket endpoint for `transcripts` that the client can send transcripts to.
- [ ]  Design + publish a JSON format for how these transcripts should be formatted when sent to Cloud
- [ ] Forward received transcripts to Apps that are subscribing to transcripts

yagarwal1307:
closes #657 

- Added logic for sending required data state to clients based on cloud status

### DIFF
--- a/agents/LOCAL_STT_FALLBACK.md
+++ b/agents/LOCAL_STT_FALLBACK.md
@@ -1,0 +1,57 @@
+# Issue:
+- Cloud STT Azure/Soniox might be down or we might run into their limits
+- Client bandwidth might be low
+
+# Logical Changes:
+- MicrophoneManager:
+    - calculateRequiredData:
+        New function to calculate the required data based on the subscriptions and cloud stt status.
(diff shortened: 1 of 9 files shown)


alex1115alex:
Also merge conflicts on this one. Paths changed, should not be difficult to merge dev in.

yagarwal1307:
oh let me fix those just relised

PR #759, opened 2025-07-16
Add Logo to MentraOS app store

ronantakizawa:
Resolves issue #735


alex1115alex:
I think the logo got added already as part of another pr

PR #780, opened 2025-07-18
Photo request specify size

OkGoDoIt:


aisraelov:
I think we've already implemented this in another pr

PR #770, opened 2025-07-17
Make App Store Load Faster

ronantakizawa:
Addresses issue #578. 

After investigating the slow load time of the app store on the mobile app, the bottleneck came from the actual website. 

Here are changes I made: 

  **1. Route-based Code Splitting (App.tsx)**

  - Before: All pages imported synchronously
  - After: Lazy loading with React.lazy()
  const AppStore = React.lazy(() => import('./pages/AppStore'));
  const AppDetails = React.lazy(() => import('./pages/AppDetails'));
  - Benefit: Reduces initial bundle size, loads pages on-demand

  **2. Bundle Optimization (vite.config.ts)**

  - Added: Manual chunk splitting for better caching
  manualChunks: {
    vendor: ['react', 'react-dom', 'react-router-dom'],
    ui: ['@radix-ui/react-dialog', '@radix-ui/react-dropdown-menu', ...],
    auth: ['@supabase/supabase-js', '@supabase/auth-ui-react', ...],
    utils: ['axios', 'clsx', 'tailwind-merge', ...]
  }
  - Benefit: Better browser caching, parallel loading of chunks

  **3. Immediate Header Display (App.tsx)**

  - Before: Full-screen loading spinner blocked entire app during auth
  - After: Removed auth loading block from main routes
  // Removed this blocking logic:
  if (isLoading) {
    return <LoadingSpinner />;
  }
  - Benefit: Header and UI show immediately, auth happens in background

  **4. Image Optimization (AppCard.tsx)**

  - Added: Lazy loading attributes to app logo images
  loading="lazy"
  decoding="async"
  - Added: Immediate placeholder with smooth transitions
  - Benefit: Images don't block card rendering, smooth loading experience

  **5. Component Performance (AppStore.tsx + AppCard.tsx)**

  - Added: useMemo for filtered apps to prevent recalculation
  - Added: useCallback for event handlers to prevent function recreation
  - Created: Memoized AppCard component with React.memo()
  - Benefit: Reduced unnecessary re-renders, better performance with large lists

  **6. Asset Preloading (index.html)**

  - Added: Preload hint for logo
  <link rel="preload" href="/logo.svg" as="image" />
  - Benefit: Faster logo loading

  **7. Suspense Loading (App.tsx)**

  - Added: <Suspense> wrapper for lazy-loaded routes
  - Benefit: Graceful loading states for code-split components

PR #783, opened 2025-07-19
Fix bug for 

ronantakizawa:
@alex1115alex This is a fix necessary for PR #770. 

Currently there are cases where the app store won't update the state of installed apps on the app card after logging in. 

This fix makes the AppStore check if the user is authenticated and the authtoken is ready to immediately update the state of installed apps.


alex1115alex:
Good catch. Didn't notice the issue when first testing, will try around some more.

Can confirm app store loads way faster now.

Issue #585, opened 2025-07-02
Add funnel to install MentraOS on app store


alex1115alex:
The App Store website will be a massive funnel for third-party glasses users. We need the app store to prompt users to install MentraOS.

**Example:** 

1. Someone on Reddit/Discord asks if there is a proactive AI for G1 / Z100 / etc
2. Someone else replies with a link to [https://apps.mentra.glass/package/com.mentra.merge](https://apps.mentra.glass/package/com.mentra.merge)
3. This page drives the person to install MentraOS

PR #785, opened 2025-07-19
Add MentraOS Download Funnel

ronantakizawa:
Addresses issue #585. 

Added iOS and Google Play Store buttons that render next to the logo text on the header of the app store and at the footer of AppDetails. 

If the website is rendered on WebView, the buttons are hidden. 

Screenshot 2025-07-19 at 4 17 41 PM
Screenshot 2025-07-19 at 4 18 19 PM
Simulator Screenshot - iPhone 11 Pro Max - 2025-07-19 at 16 17 51


alex1115alex:
I think the play here would be to replace the typical app store badges w/ a properly styled "Get MentraOS" button (next to the header's "Sign In" button) that deeplinks to the appropriate app store for that platform.

Potentially even do the same thing for the "Get App" button if the user is on mobile.

What do you think?

PR #788, opened 2025-07-20
Mentra live ble image

alex1115alex:


asg_client/app/src/main/assets/test.jpg
OkGoDoIt, line 1: maybe we should remove this?

PR #897, opened 2025-07-31
Make Simulated Glasses the default / initial app state

fossephate:


aisraelov:
Let's revisit after android_core refactor. This would cause issues with the current Android setup.

PR #982, opened 2025-08-08
Ahmad/file sync

aisraelov:


aisraelov:
Need to test:
* Sending photos directly via WIFI (was failing both here and on dev so need another round of testing)
* Streaming (untested)
* Gallery (untested)

PR #1132, opened 2025-09-01
Ya/fix local transcription

yagarwal1307:
- Revert @fossephate this [commit](https://github.com/Mentra-Community/MentraOS/commit/0a541c1a66aa2ba2faef5ff499d2221ea1b137cf#diff-aa58e9c1d6c11194a98966c17ab29600481c9386acaa102389442aa42f4bb74a). (Issue: In case of local transcription mic will be enabled but we won't be sending pcm to the cloud. Hence that logic is wrong) 
- Add a listener for restartTranscriber in the new CommandBridge.

### DIFF
--- a/android_core/app/src/main/java/com/augmentos/augmentos_core/smarterglassesmanager/SmartGlassesManager.java
+++ b/android_core/app/src/main/java/com/augmentos/augmentos_core/smarterglassesmanager/SmartGlassesManager.java
@@ -928,7 +928,7 @@ public class SmartGlassesManager extends Service {
             Log.w(TAG, "PhoneMicrophoneManager is null, skipping setRequiredData call");
         }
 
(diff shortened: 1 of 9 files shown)


fossephate:
We shouldn’t revert that commitment; there should only be a single source of truth on whether the mic should be on and the required data is enough info to go off of


having both a micEnabled field and requiredData field is confusing because if required data is transcription but micEnabled is set to off then those are conflicting arguments

lmk if I’m missing something but I’m pretty sure if there’s an issue with this it should be fixed server side

yagarwal1307:
No you are right. Earlier it was there for backward compatibility.
But in your last commit it broke because we were assuming that mic should be on only when pcm is required which isn't true. We will need mic to be on in any case whether pcm or transcription. Will revert my changes and modify the logic.

PR #1153, opened 2025-09-04
chore: handle button events and send them to cloud

yagarwal1307:


yagarwal1307:
@aisraelov 
The reason I didn't do that was because it seemed to have defined some enums which aren't true for the nex glasses. 
So I wasn't sure if something would break in the chain if I do that.
But you are right maybe we should modify and use the same infra.
Let me look into it again.

```
export interface ButtonPress extends BaseMessage {
  type: GlassesToCloudMessageType.BUTTON_PRESS;
  buttonId: string;
  pressType: 'short' | 'long';
}
```
In nex we have 4 different press types for example

cc: @nic-olo

PR #1193, opened 2025-09-10
New sdk version message

Drakonheart:
If a developer is using an older version of the Mentra SDK, then every time they run bun run dev to start their app, a large Mentra logo will pop up in the terminal to remind them to update the SDK.

This check targets the SDK version inside node_modules, not the project’s root package.json.

Img :
image

Extra doc:
https://docs.google.com/document/d/1bGdLyKOIHrxg5eAdIJgFkUd0j-QlDsbMqTF9vyFMvm0/edit?usp=sharing


Drakonheart:
The npm is changed to bun

PR #1237, opened 2025-09-17
Mentra Live bug fixes

aisraelov:
* user connection tracking to report connection failures early in sdk
* tiny improvements to gallery sync flow
* remove some dead asg code
* faster asg ble fallback for images


isaiahb (approved):
Looks good, just need to ensure

PR #1266, opened 2025-09-25
Removed all permissions; tested and verified it should be good to go

Drakonheart:
image


aisraelov:
what about old apps that have the ALL permission? what happens to those?

Drakonheart:
Updated: only the UI is affected. Migration script coming tomorrow morning.

PR #1274, opened 2025-09-26
cloud es-lint / prettier init refactor, cleanup dead code

isaiahb:


aisraelov:
@isaiahb If this is good and tested you can merge this through so it doesn't go stale

PR #1278, opened 2025-09-27
Mentra live improve photo requests

nic-olo:


aisraelov:
Before we merge this or go further with iOS, need to do some actual testing as to if cutting out the MTK<=>BES packet ACK system is actually a good idea.

Since we had to bump up the between-packet delay to 20ms it might not be worth it.

PR #1292, opened 2025-09-30
Add new env var for Mira

Drakonheart:
Added a new env var with dynamic support for Mira package names.


Drakonheart:
For future reference, if we want to bring this back, add the following env var for everything to work:

```
ALLOWED_MIRA_API_PACKAGES=test.augmentos.mira,cloud.augmentos.mira,com.augmentos.mira,dev.mira.local  # Make sure there are no spaces here
```

---

PR #1315, opened 2025-10-06
Ya/china cloud cd

yagarwal1307:


cloud/.env.example
```diff
@@ -2,6 +2,7 @@
 # cloud/.env
 # This file contains the base configuration for the AugmentOS Cloud backend.
 
+DEPLOYMENT_REGION=china
```
aisraelov: `.env.example`

Can we add a hint here of what the other deployment region might be?

Maybe do smth like:

```
# DEPLOYMENT_REGION=china
DEPLOYMENT_REGION = global
```

yagarwal1307: Sure let me do that

mobile/.env.example
```diff
@@ -1,3 +1,6 @@
+# Deployment region
+DEPLOYMENT_REGION=china
```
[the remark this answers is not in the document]
yagarwal1307: Sure

PR #1398, opened 2025-10-26
Centralize web auth

rajeevrajeshuni:


yagarwal1307:
This will affect our CI/CD pipelines because currently our deployments for account/store/console are very independent. 
cc: @isaiahb

cloud/websites/shared/src/auth/components/ForgotPassword.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    {errorMessage}
+                  </p>
+                </div>
+              )}
+            </div>
+
+            {/* Sign In Link */}
+            <div className="text-center text-sm text-gray-500 mt-6">
+              <p
+                onClick={() => console.log('Navigate to sign in')}
```
yagarwal1307: We need to implement this

PR #1399, opened 2025-10-27
Fix bugs caused by Android refactor

aisraelov:
This PR fixes some issues due to Android refactor.
* Mic handling is now improved, G1 mic + Phone Mic audio switching works again
* Fixes a bug preventing playing audio through Mentra Live via SCO mode
* Fixes a bug launching live captions vs offline live captions
* Mentra Live i2s system somewhat works
* Fixes issues with unmanaged RTMP streams


aisraelov:
@codex review

PR #1420, opened 2025-11-02
Tpa can trigger wifi setup

aisraelov:
this pr:
1. Makes it so that pairing Mentra Live on iOS prompts you to connect the BT Classic
2. Fixes an issue where we see "DISCONNECTED" and "CONNECTING" flicker at the top of the screen every minute or so
3. Phone reports the glasses wifi connection status to cloud. Calls to start streaming now check if WiFi is enabled before sending to phone. If not, the TPA is informed.
5. Adds an endpoint in SDK that allows TPAs to trigger a popup on the phone to configure glasses wifi


aisraelov:
@codex review

PR #1445, opened 2025-11-07
Low battery camera actions

adeeteya:
Prevent clicking photos, videos or starting rtmp streams when battery is below 10%, by providing appropriate feedback as well.

Also handle cases where rtmp stream is ongoing and battery drops below 10% to automatically stop the stream.

### DIFF
--- a/asg_client/app/src/main/java/com/augmentos/asg_client/service/core/handlers/K900CommandHandler.java
+++ b/asg_client/app/src/main/java/com/augmentos/asg_client/service/core/handlers/K900CommandHandler.java
@@ -52,12 +52,10 @@ public class K900CommandHandler {
 
             switch (command) {
                 case "cs_pho":
-                    extractAndUpdateBattery(bData);
                     handleCameraButtonShortPress();
                     break;
 
                 case "cs_vdo":
-                    extractAndUpdateBattery(bData);
                     handleCameraButtonLongPress();
                     break;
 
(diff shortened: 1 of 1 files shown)


aisraelov:
@codex review

aisraelov (changes requested):
Some immediate issues that need fixed:

  1. MediaCaptureService: Registers in constructor, cleanup() exists but never called → memory leak
  2. RtmpStreamingService: Has @Subscribe method but never registers at all → dead code that does nothing
  3. Both stay subscribed even when not recording/streaming → wasting resources

  But generally I think we should not use EventBus for this.

  ## The Real Problem

  EventBus is over-engineered here. You already have `stateManager.getBatteryLevel()` working perfectly for the proactive checks. Just use
  polling for reactive monitoring too:

  ```java
  // In video recording loop - check every 10 seconds
  while (isRecording) {
      if (stateManager.getBatteryLevel() < 10) {
          stopRecording();
          playBatteryLowSound();
          break;
      }
      Thread.sleep(10000);
  }
```

  Why polling is better:
  - No lifecycle management = no memory leaks
  - Self-contained logic
  - Easy to debug
  - Battery changes slowly enough that 10s polling is fine

  Other Fixes Needed

  1. Inconsistent threshold: Use <= 10 everywhere, not < 10 in some places
  2. Magic number: Extract MIN_BATTERY_LEVEL = 10 to a shared constant instead of defining in 3 files

PR #1450, opened 2025-11-09
Livekit FIX... Options For Audio Mix And Stop Other Audio

AryanFP:
From the SDK, you now have the option to set:

- stopOtherAudio: false → multiple audios can play and overlap with each other.
- stopOtherAudio: true → the current audio stops any other audio before playing its own.

### DIFF
--- a/cloud/bun.lock
+++ b/cloud/bun.lock
@@ -87,7 +87,7 @@
         "ali-oss": "^6.23.0",
         "axios": "^1.7.4",
         "bun-types": "1.0.17",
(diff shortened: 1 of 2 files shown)


cloud/packages/sdk/src/logging/logger.ts
```diff
@@ -1,38 +1,38 @@
-import pino from 'pino';
+import pino from "pino"
 
 // Constants and configuration
```
Drakonheart: No Changes here ESLINT...

PR #1465, opened 2025-11-11
Add “Install” Button for Third-Party Apps — Because Magic

AryanFP:
This PR makes the developer console dramatically more seamless and intuitive by introducing a clear **Install** button for third-party apps.

### Highlights

* Added an **Install** button when the app isn’t installed on the developer’s account.
* Removed redundant “API Key” and “Publish” buttons from the Actions tab.
* Improved the Share dialog to explain two installation options: via app link or package name.
* Clarified that install/uninstall only affects the logged-in user’s account.

### Why

Developers often need to reinstall or share their apps after uninstalling, switching accounts, or collaborating with others.
This change makes that process instant, obvious, and just a little bit magical.

image

image

image


isaiahb (approved):
lgtm

PR #1472, opened 2025-11-11
Update issue templates (bug reports)

aaviator42:
This PR adds a template for bug reports so users are more likely to provide us with all the information that we might need to actually address their issues.


aisraelov:
@codex explain what this actually does in practice

PR #1482, opened 2025-11-12
[Cloud]: Add device state REST API & consolidate state in DeviceManager

isaiahb:
## Problem
Display requests fail with `GLASSES_DISCONNECTED` error even when glasses are connected. Root cause: mobile never syncs device state to cloud, and device state is scattered across UserSession and DeviceManager with no clear source of truth.

## Solution
1. Add REST endpoint `POST /api/client/device/state` for device state updates
2. Create `GlassesInfo` type in `@mentra/types` package
3. Refactor DeviceManager to store `deviceState: Partial<GlassesInfo>` as single source of truth
4. Remove duplicate state from UserSession (`glassesConnected`, `glassesModel`, etc.)
5. Update ConnectionValidator to read from DeviceManager

## Implementation
- **Type**: `GlassesInfo` in `@mentra/types` (connection, model, WiFi, battery, metadata)
- **Endpoint**: `POST /api/client/device/state` with `Partial<GlassesInfo>` payload
- **DeviceManager**: Add `updateDeviceState()`, `getDeviceState()`, `isConnected()`, `getModel()` methods
- **UserSession**: Remove glasses state, keep only phone connection state
- **ConnectionValidator**: Use `deviceManager.isConnected()` instead of `userSession.glassesConnected`

### DIFF
--- a/cloud/packages/cloud/src/services/layout/tests/harness/MockUserSession.ts
+++ b/cloud/packages/cloud/src/services/layout/tests/harness/MockUserSession.ts
@@ -79,12 +79,14 @@ export class MockUserSession implements Partial<UserSession> {
   bufferedAudio: ArrayBufferLike[] = [];
   whatToStream: any[] = [];
 
-  // Connection state properties needed for ConnectionValidator
-  phoneConnected: boolean = true; // Simulate connected phone
-
   // Mock DeviceManager for tests
   deviceManager: any = {
-    isConnected: () => true,
+    get isPhoneConnected() {
+      return true;
+    },
+    get isGlassesConnected() {
+      return true;
+    },
     getModel: () => "Even Realities G1",
     getDeviceState: () => ({
       connected: true,


cloud/packages/cloud/src/services/layout/tests/harness/MockUserSession.ts
Drakonheart: Question how did you test this? As in how do you run this test?

cloud/issues/device-state/device-state-architecture.md
```diff
@@ -0,0 +1,383 @@
...
+    payload.connected = true
+    this.logger.debug("Inferred connected=true from modelName")
+  } else if (
+    (payload.modelName === null || payload.modelName === "") &&
+    payload.connected === undefined
+  ) {
+    payload.connected = false
+    this.logger.debug("Inferred connected=false from empty modelName")
+  }
+
```
Drakonheart: Could you explain the logic to me for this section?

cloud/packages/cloud/src/services/layout/tests/harness/MockUserSession.ts
isaiahb: it's a pre-existing test from months ago, had to adopt it so it passes. it's automatically run as part of the ci/cd for the cloud.
can run it locally by running `bun test` from /cloud folder to run all the tests.
I believe bun automatically finds and runs files names <x>.test.ts

PR #1484, opened 2025-11-13
Fix various native module issues

aisraelov:
* Fix battery status (G1) after reconnecting
* Fix forgetting mentra live glasses not actually disconnecting from them on iOS


aisraelov:
@codex review

PR #1582, opened 2025-11-20
stt-ios-fixes

fossephate:


aisraelov:
This takes 9 minutes to extract. We need to go back to old lib (or your idea, reupload as .zip, decompress .zip instead)

Issue #1498, opened 2025-11-14
Microphone stays on after forgetting or disconnecting glasses


aisraelov:
<img src="https://uploads.linear.app/55865761-8742-4aa0-be4e-71cf49ead33e/59744e9d-c553-42c5-bf50-752286211439/1d34ae6d-1380-4dd2-a5ff-53d5df8763fc?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzU1ODY1NzYxLTg3NDItNGFhMC1iZTRlLTcxY2Y0OWVhZDMzZS81OTc0NGU5ZC1jNTUzLTQyYzUtYmY1MC03NTIyODYyMTE0MzkvMWQzNGFlNmQtMTM4MC00ZGQyLWE1ZmYtNTNkNWRmODc2M2ZjIiwiaWF0IjoxNzYzMTQxMDkzLCJleHAiOjE3OTQ3MTE2NTN9.WkGEbjQ0XtSLR_GBokpqJFyMaGIYwVIAQbGO0trFIAg " alt="image.png" width="950" data-linear-height="71" />

PR #1585, opened 2025-11-20
Fix microphone staying on after disconnecting or forgetting glasses

AltuisticIsopod:
Issue: https://github.com/Mentra-Community/MentraOS/issues/1498

Added microphone stop call to handle_disconnect() and handle_forget() in both Android and iOS CoreManager to fix issue where microphone stays on after disconnecting or forgetting glasses.


aisraelov:
Thanks for the PR! Could you update the PR with the fixes the Codex bot flagged? Once those are in, I can approve and merge.

PR #1624, opened 2025-11-28
Ya/clean china cicd

yagarwal1307:


mobile/src/stores/settings.ts
```diff
@@ -65,7 +65,7 @@ export const SETTINGS: Record<string, Setting> = {
     key: "store_url",
     defaultValue: () =>
       process.env.EXPO_PUBLIC_DEPLOYMENT_REGION === "china"
-        ? "https://store.mentraglass.cn"
+        ? "https://dev-store.mentraglass.cn"
         : "https://apps.mentra.glass",
```
isaiahb: nitpick: this could also be an env var so in the future it's easier to separate prod / dev env resources 
especially now that we are moving towards having both apps.mentraglass.com and appsbeta.mentraglass.com

isaiahb (approved):
LGTM!
left a few nitpicks, but not crucial so feel free to ignore those for now

PR #1744, opened 2025-12-19
Gallery n camera improvements

nic-olo:


aisraelov:
Syncing works but photos are still too low, and scrolling left/right is laggy and only works for videos; not pictures.

aisraelov:
[] needs some small bottom padding for old android navbar when watching a video
[] need to fix sometimes cannot scroll up/down in the gallery
[] When you scroll to a video (or back to a video) it should restart at the beginning of the video

PR #1839, opened 2026-01-07
Install Mini Apps to the home screen

fossephate:


aisraelov:
Closing due to staleness. It's a cool and great idea though. Would love to add this properly at some point.

PR #1856, opened 2026-01-09
Create new Hono-based store API routes under /api/store for the Mentr…

AryanFP:
# MentraOS Store API Update

## Summary
This update refactors the MentraOS Store backend and frontend to use a cleaner, more scalable API structure based on Hono. It improves code organization, clarity of endpoints, and overall maintainability.

## Backend Changes
- Create new Hono-based store API routes under:

- Add `store.service.ts` to separate business logic from HTTP routing, improving testability and code structure.
- Implement core store functionality:
- App browsing
- App installation
- User management endpoints

## API Restructuring
- Public app browsing (no authentication required):

- Authenticated app browsing (includes installation status):

## Frontend Changes
- Update the store frontend to use the new API structure.
- Migrate API calls from:

## Benefits
- Cleaner and more intuitive API design
- Clear separation of concerns between routing and business logic
- Improved maintainability and test coverage
- Better distinction between public and authenticated store views


cloud/packages/cloud/src/services/core/store.service.ts
```diff
@@ -0,0 +1,197 @@
...
+  // Get full app details for each installed app
+  const installedApps = await Promise.all(
+    (user.installedApps || []).map(async (installedApp) => {
+      const appDetails = await appService.getApp(installedApp.packageName);
+      return {
+        ...appDetails,
+        installedDate: installedApp.installedDate,
+      };
+    }),
+  );
```
isaiahb: This is the worse possible way to do this. 

we should try to not use appService(we want to prevent this service from having dependancies on app service, because appService sucks and we want to depreciate it, this file is a change to rewrite the db queries to be optimized and not tied to the constraints that appService is, 
we use the App model from db specifically so we can actually create optimized endpoints for this.

i.e we can get list of all packageNames we need and make 1 query to Mongodb, instead of spamming our db with multiple queries getting 1 app at a time.

PR #1930, opened 2026-01-19
mobile/true-glasses-mirror-display

isaiahb:


fossephate:
I'll need to double check the padding changes don't mess with the related screens because getting that all to line up correctly was tricky

Also if we keep them we should let parth know since it's different from the current figma design

Otherwise looks good

PR #1988, opened 2026-01-31
fix: auto-install app for developer when created via CLI/console

isaiahb:
When a developer creates an app via the CLI or console, the app is now automatically installed in their user.installedApps array. This allows them to immediately test their app without manual installation.

The autoInstallAppForDeveloper function was already in developer.routes.ts but was missing from the console.apps.service used by CLI/console routes.


aisraelov:
@codex review

PR #1990, opened 2026-02-01
Fixes 29 3

fossephate:


aisraelov:
hey so onboarding still has lots of issues in this branch and I dont have time to fix it again monday, let's use the old branch

PR #2035, opened 2026-02-14
feat: re-apply display height after reconnection in CoreManager

yagarwal1307:


fossephate:
hmm definitely a few ways to go about it
I think probably the cleanest thing is probably making the sgc grab the core setting (from the glasses store) and sending it on connect

PR #2108, opened 2026-02-25
fix(console): org creation infinite loop + slug collision + VersionError

isaiahb:
Users with common name prefixes (alex, chris, david) got stuck in an infinite loop on the Organization Settings page because:

1. generateSlug() produced non-unique slugs → E11000 duplicate key
2. createPersonalOrg() was not idempotent → blind insert every call
3. Frontend useEffect retried on every failure (unstable deps + no guard)
4. user.save() VersionError when 5 concurrent paths raced on same doc
5. Frontend didn't re-list orgs after failed create

Fixes:
- generateSlug: check DB, append random suffix on collision
- createPersonalOrg: idempotency check + E11000 catch
- All user.save() calls → atomic User.updateOne($addToSet/$set)
- OrganizationContext: useCallback + ref mutex on ensurePersonalOrg
- OrganizationSettings: single-attempt ref guard, stable deps
- Frontend catch: re-list orgs before erroring (concurrent recovery)

Confirmed via Better Stack logs: 7015 errors for one user, 1175 for another.

Issue: 043-org-creation-bug

## Release Notes

* **Bug Fixes**
  * Fixed organization creation race condition causing infinite loops and failed initialization
  * Resolved concurrent request errors during org assignment
  * Eliminated infinite retry loops and duplicate toasts in settings
  * Enhanced error handling for failed organization bootstrap

* **New Features**
  * Added automatic retry with backoff logic for organization operations

* **Refactor**
  * Improved organization settings UI layout and responsiveness


isaiahb:
@codex review

PR #2326, opened 2026-03-27
[WIP][Do Not Merge] SDK V3 

isaiahb:


isaiahb:
## Addressing previous review comments

**Codex comment 1: "Preserve legacy session ID format in start webhooks"**
Fixed in commit cd91c2a. The webhook now passes the legacy `userId-packageName` format as `sessionId` to AppSession, so v2 SDKs can parse it with `sessionId.split('-')[0]` to recover the userId. The v3 `sessionToken` (UUID) is a separate field for v3 SDKs.

**Codex comment 2: "Keep app session IDs resolvable to user sessions"**
Fixed in the same commit cd91c2a. AppSession receives the legacy format instead of defaulting to `randomUUID()`. Code paths like `transcripts.routes.ts` that do `appSessionId.split('-')[0]` continue to work. The UUID-based sessionId is only used internally by the v3 runtime.

Both fixes are verified and deployed to the debug environment.

---

@codex review

isaiahb:
lgtm!!

PR #2483, opened 2026-04-08
Restore audio level after it was changed by audio playing

PhilippeFerreiraDeSousa:
Behaviour verified.
Note that Mentra AI does the bubbly sound for a long time before answering, and this is still forced to be on high volume

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth (A2DP) volume handling to avoid repeated volume bumps during playback and to reliably restore original device volume when playback ends, fails, or is interrupted.
  * Prevents premature volume restoration when a new playback request may reuse the temporarily raised volume.


aisraelov:
@codex review

aisraelov:
Nice work!

PR #2487, opened 2026-04-08
Mentra live throttles bitrate instead of overheating

PhilippeFerreiraDeSousa:
## Summary by CodeRabbit

* **New Features**
  * Thermal-aware bitrate management that caps/adapts bitrate based on device temperature.
  * Periodic real-time stream metrics reporting (bitrate, FPS, resolution, dropped frames, duration, temperature) exposed to apps/viewers and included in status updates.
  * Safer camera shutdown handling to avoid teardown/reconnect crashes.
  * Console log text now selectable in the developer UI.

* **Bug Fixes**
  * Improved resilience during camera teardown to prevent crashes.

* **Tests**
  * New unit tests validating thermal controller behavior and temperature smoothing.


aisraelov:
@codex review

aisraelov (commented):
I reviewed the diff and checked the CodeRabbit/Codex feedback. The important items I think should be fixed before merge are:

1. `WhipStreamingService.applyTemperatureControl` only compares bitrate before returning. If width/height/FPS/audio settings change while bitrate stays the same, `mStreamConfig` remains stale, and camera/audio setup still reads from `mStreamConfig`. The overlapping CodeRabbit/Codex comments are valid.

2. Thermal sysfs reads are happening on the main looper. `startSysfsThermalPolling()` posts to `mMainHandler` and calls `WhipThermalUtils.readCpuTemperature()` there every 2s. That blocking file I/O should move to a background handler/executor. While touching it, guard `reader.readLine()` for null/empty input so an empty sysfs read does not crash the polling path.

3. Managed terminal statuses lose final telemetry. `stream.latestStats` is cleared before `sendManagedStreamStatus()` sends `stopped`/`error`, but the outbound payload reads `stream.latestStats`. Send the terminal status first, then clear/cleanup.

4. `normalizeStreamTelemetry()` chooses either `status.stats` or compact `status.m` wholesale. If `stats` is present but partial, valid compact fields are discarded and stats can become `undefined`. Merge fallbacks per field instead.

5. New 2s WHIP metrics messages use `status: \"streaming\"`; the managed handler currently calls `updateStreamUrls()` for every streaming status. That can turn telemetry into repeated Cloudflare `getStreamDetails` polling. Gate it to first-active/missing-URL cases or throttle it.

Lower-priority bot notes like test naming/Javadocs can be cleaned up separately. The CodeRabbit note about a WHIP default bitrate increase looks stale on current head; `DEFAULT_VIDEO_BITRATE` is still 1 Mbps.

PhilippeFerreiraDeSousa:
Paused because more work was needed for:
1. addressing reviews
2. behavior was not working on SRT and RTMP when verifying

PR #2510, opened 2026-04-11
Add A6N software depth distance mapping

wangli65535:
## Summary
This PR adds software-controlled A6N depth distance adjustment.

The display distance can now be configured from 2.0m to 5.0m, with 2.5m as the default 0px offset. The firmware converts the requested distance into a left/right stereo pixel offset using a trigonometry-based model.

Legacy app values are still supported:
- `1` maps to 2.0m
- `2` maps to 2.5m
- `3` maps to 5.0m

## Changes
- Add distance-to-offset conversion helper in the A6N driver
- Apply `DisplayDistanceConfig` as either legacy tier values or real centimeter values
- Add `display distance <meters>` shell command
- Add `display distance_test` shell command to print the distance-to-pixel table
- Remove unused A6N write performance counters

## Testing
- Ran `display distance_test`
- Tested `display distance 2.0`
- Tested `display distance 2.5`
- Tested `display distance 5.0`
- Confirmed stereo direction appears correct on device
- Build not run per local workflow

## Notes
The current mapping uses calibrated software constants. Final distance accuracy should be validated with optical measurement equipment when available.

### DIFF
--- a/mcu_client/nrf5340_ble_simulator/custom_driver_module/drivers/display/lcd/a6n.c
+++ b/mcu_client/nrf5340_ble_simulator/custom_driver_module/drivers/display/lcd/a6n.c
@@ -1,7 +1,7 @@
 /*
  * @Author       : Cole
  * @Date         : 2025-07-28 11:31:02
- * @LastEditTime : 2026-03-25 09:52:59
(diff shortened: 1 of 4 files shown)


yagarwal1307:
LGTM

PR #2550, opened 2026-04-14
Fix OTA flow

nic-olo:
Test
ASG_Client 27 => 37
ASG_Client 36 => 37
ASG_Client 37 => 38

Track whether a WiFi status update has been received from glasses to distinguish "not yet known" from "definitely disconnected", preventing premature wifi-gating before the first status event arrives.

Made-with: Cursor

## Summary by CodeRabbit

* **New Features**
  * OTA sessions persist across restarts for reliable resume and completion reporting; new unified OTA status and query flow with simplified progress UI.
  * OTA progress screen reworked with watchdogs, retries, and reconnect-aware behavior.
  * Saved Wi‑Fi networks refresh on screen focus; Wi‑Fi tile shows “Loading…” until status known; connection-overlay suppression during expected reboot/install.

* **Bug Fixes**
  * Stronger timeouts, retry/reset and stall detection for OTA; improved retry/ack handling and clearer failure messages.
  * Unified, reliably delivered OTA status to reduce stale/incorrect progress and improve resume behavior.

* **Documentation**
  * Added comprehensive OTA rearchitecture specification.


aisraelov:
@codex review

aisraelov (commented):
How does OtaService read persisted session? If OtaService is part of asg_client, and we install an update asg_client, OtaService also gets restarted. We either need a dedicated OTA APP or handle this another way.

Previously, we handled this in a very dumb (but reliable way): asg_client tracked its build number to sharedpreferences, and when asg_client started, it would check if its current build number is higher than its previous known build number from sharedpreferences. If that was true, it meant we had just come from an asg_client update, in which we would then trigger MTK/BES updates.

docs/ota-rearchitecture-spec.md
```diff
@@ -0,0 +1,836 @@
...
+ota_start ──────────────────►  startOtaFromPhone()
+                               ├─ Create OTA session (persisted to SharedPreferences)
+                               ├─ Compute step sequence from version manifest
+◄────────────────────────────  ota_status {step:1/3, phase:download, overall:0%, status:in_progress}
+                               ├─ Download APK
+◄────────────────────────────  ota_status {step:1/3, phase:download, overall:15%, ...}
+                               ├─ Install APK (sends FINISHED, then process dies)
+◄────────────────────────────  ota_status {step:1/3, phase:install, overall:33%, status:step_complete}
+
+                               ── App restarts ──
```
aisraelov, line 217: How does OtaService read persisted session? If OtaService is part of asg_client, and we install an update asg_client, OtaService also gets restarted. We either need a dedicated OTA APP or handle this another way.

Previously, we handled this in a very dumb (but reliable way): asg_client tracked its build number to sharedpreferences, and when asg_client started, it would check if its current build number is higher than its previous known build number from sharedpreferences. If that was true, it meant we had just come from an asg_client update, in which we would then trigger MTK/BES updates.

docs/ota-rearchitecture-spec.md
```diff
@@ -0,0 +1,836 @@
...
+
+## Detailed Design
+
+### A. Glasses: OTA Session Persistence
+
+**New class:** `OtaSessionManager.java`
+
+**SharedPreferences key:** `"ota_session"`
+
+**Persisted fields:**
```
aisraelov, line 265: Is there a recovery mechanism for what happens if we get to an awkward state? Like what happens if the glasses crash / run out of battery / something weird happens mid-update? Is there a way to clear update state?

Remember right now asg_client is pretty stateless. This is how we've managed to avoid needing a factory reset option for so long.

docs/ota-rearchitecture-spec.md
```diff
@@ -0,0 +1,836 @@
...
+  "total_steps": 3,
+  "current_step": 2,
+  "step_type": "mtk",
+  "phase": "install",
+  "step_percent": 60,
+  "overall_percent": 53,
+  "status": "in_progress",
+  "error_message": null
+}
+```
```
aisraelov, line 345: this is a large payload and we don't have a good message chunking system for going asg_client => phone. please ensure this fits in mtu with a real uuid.

aisraelov:
TLDR of my comments. I agree w/ general premise. Move logic to glasses. This makes everything easier and more maintainable long term when considering how OTAs will work in future glasses models (if G1 style will likely be handled mostly internally within SGC, not RN layer in a giant progress.tsx file). 

HOWEVER:

I'm a little worried about persisting this giant blob of OTA progress information. This makes asg_client more stateful and increases surface area for bugs that can persist through a reboot. Need to be very careful this gets wiped away after 30 minutes.

Keep in mind on time: Mentra Live syncs its time with the internet some time after you connect to a wifi network. (This is just handled automatically by underlying android os not our APP). Need to be sure  we don't save an initial time when the clock hasn't been accurately synced yet, this time would be too early. Also need to make sure there isn't a way we can set initial time on a synced clock then somehow regress.

nic-olo:
@coderabbitai review

nic-olo:
@codex review

nic-olo:
@codex review

aisraelov (approved):

PR #2555, opened 2026-04-15
Add unified display brightness control with ambient AUTO mode

wangli65535:
## Summary

- Add `mos_brightness` component for unified manual and AUTO brightness control
- Move shell commands to `display brightness auto ...` to make AUTO brightness ownership explicit
- Tune AUTO brightness mapping for the installed OPT3006 position, where effective lux is much lower than room lux
- Add optional AUTO sample logging controlled by shell

## Changes

- Added shared brightness APIs for manual brightness, AUTO enable/disable, multiplier, sample logging, and status query
- Added shell commands:
  - `display brightness <0-100>`
  - `display brightness auto on/off/status`
  - `display brightness auto log on/off/status`
  - `display brightness auto multiplier <0.10-3.00>`
- AUTO brightness samples OPT3006 every 1500 ms when enabled
- AUTO brightness applies EMA filtering, hysteresis, and minimum apply interval to reduce flicker
- Brightness writes go through A6N Bank0 `0xE2`
- Protobuf brightness control now uses the shared `mos_brightness` module
- Initial display brightness now uses the shared brightness path

## AUTO Brightness Tuning

The OPT3006 is installed near the frame with the light inlet facing downward, so measured values are effective device lux rather than normal room lux.

Current curve:

| Effective lux | Brightness |
| --- | --- |
| 0 | 30% |
| 10 | 30% |
| 30 | 50% |
| 80 | 58% |
| 150 | 68% |
| 300 | 85% |
| 500+ | 100% |

## Logging

AUTO sample logs are disabled by default to avoid flooding the console.

### DIFF
--- a/mcu_client/nrf5340_ble_simulator/CMakeLists.txt
+++ b/mcu_client/nrf5340_ble_simulator/CMakeLists.txt
@@ -28,6 +28,7 @@ add_subdirectory(src/mos_components/mos_font_storage)
 add_subdirectory(src/mos_components/mos_lvgl_display)
 add_subdirectory(src/mos_components/mos_jlink_usb_switch)
 add_subdirectory(src/mos_components/mos_dfu_progress)
+add_subdirectory(src/mos_components/mos_brightness)
 add_subdirectory(src/protobuf)
 add_subdirectory(src/proto)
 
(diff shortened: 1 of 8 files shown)


yagarwal1307 (approved):
Rather than running the computation on the MCU we should try to run on the sensor and use interrupts. 
For now let's merge it, but lets try to modify this in the future.

PR #2579, opened 2026-04-21
Add incidents and alerts to the e2e captions tester

PhilippeFerreiraDeSousa:
image

image

image

image

image

## Summary by CodeRabbit

* **New Features**
  * Live React monitoring dashboard for E2E tests (Overview, Incidents, Alerts, Latency) with build/run workflow.
  * Incident config support and runtime controls for dropped captions, audio-device mismatch, app-foreground, and high-latency; read-only archived mode.
  * Audio output device option with optional auto-switching; internal Android build variant and scripts.

* **Documentation**
  * Updated E2E README with run, rebuild, incident config, and operational guidance.

* **Chores**
  * Replaced legacy local history server with prebuilt UI serving; ESLint/TS resolver tweaks; improved incident/alert logging.


aisraelov:
This is an Expo app, so the receiver/test-only native bits belong in `modules/core`, not in committed `mobile/android/`.

the repo should encode the `internal` lane completely from committed config/plugin code.

Concretely, I think the missing pieces are:

1. `mobile/plugins/android.ts`

   In `withAppBuildGradleModifications`, add the app-side `internal` build type to the generated `android/app/build.gradle`.

   Right now the PR adds `internal` in `mobile/modules/core/android/build.gradle`, which is the library side, and then calls `:app:assembleInternal` / `assembleInternal`. What I do not see is the corresponding app build type being created from committed Expo/plugin logic.

   This is the place I would expect to inject something along the lines of:

   ```gradle
   internal {
       initWith release
       matchingFallbacks = ['release']
       signingConfig signingConfigs.release
   }
   ```

   I am not saying to commit `mobile/android/`; I am saying the Expo app plugin should generate the app variant.

2. `mobile/eas.json`

   The local `android:internal` script sets:

   ```bash
   EXPO_PUBLIC_ENABLE_E2E_METRICS=true
   ```

   but the EAS `build.internal` profile only sets `gradleCommand`.

   So I think `mobile/eas.json` should also add:

   ```json
   "env": {
     "EXPO_PUBLIC_ENABLE_E2E_METRICS": "true"
   }
   ```

   under `build.internal`, so local `bun android:internal` and EAS `internal` builds actually have the same instrumentation behavior.

If those two pieces are added, my main Android/build-lane concern mostly goes away. `internal` then becomes a properly reproducible “release-like build with test hooks,” which seems like the right end state for this harness.

PR #2581, opened 2026-04-21
Mentra live usb camera

aisraelov:


aisraelov:
@nic-olo the camera is flipped horizontally

PR #2666, opened 2026-04-28
Fix pre-commit formatting scope

PhilippeFerreiraDeSousa:
## Summary

Fix the pre-commit formatting scope so root-only lint-staged patterns do not also run on nested files by basename.

This also applies the cloud Prettier config to the two cloud issue docs that were being reformatted during dev merges.

## Why

When `origin/dev` was merged into the Bluetooth SDK rollout branch, newly staged cloud Markdown files were picked up by both the cloud Markdown formatter and the supposed root-only Markdown formatter. That produced unrelated formatting noise in the SDK PR.

The root patterns now use `./*` so they only match files at the repository root, matching the existing comment and avoiding duplicate formatting of nested cloud docs.

## Validation

- `lint-staged --diff=origin/dev --verbose` with the repository dependencies on PATH
- `git diff --check origin/dev..HEAD`

### DIFF
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -12,8 +12,8 @@ module.exports = {
   "!(cloud)/**/*.{js,jsx,ts,tsx,json,css,md,html,yml,yaml}": "prettier --write",
 
   // Root-level config files: prettier only (root eslint needs expo which may not be installed)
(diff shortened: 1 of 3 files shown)


PhilippeFerreiraDeSousa:
pre-commit linting is no more

PR #2687, opened 2026-04-29
Fix miniapp starter template

PhilippeFerreiraDeSousa:
## Summary
- add the missing `create-mentra-miniapp` generator bin for portable starter projects
- update the starter template so it wraps apps in `MentraProvider`, uses `session.display.showTextWall`, declares a usable manifest/icon/port, and typechecks build/dev scripts
- keep Bun dev/build bundles on the app React copy and lazy-load mock transport so iPhone WebViews do not crash on first load
- re-register local miniapp runtimes on CONNECT so re-scans/reloads recover after ping timeout unregisters

## Verification
- `bun x tsc --noEmit -p sdk/create-mentra-miniapp/tsconfig.json`
- `bun run build` in `sdk/miniapp`
- `bun run test` in `sdk` (110 passing)
- fresh scaffold in `/tmp/mentra-template-portable-smoke`; generated `package.json` keeps registry semver deps with no local `file:` paths or postinstall symlink script
- earlier iPhone scan of fresh template dev QR showed `Connected to MentraOS` with no Bun mock-module error or React hook dispatcher crash

## Notes
- The template now assumes published `@mentra/miniapp` / `@mentra/miniapp-cli` packages are installable by Bun; no local monorepo sibling checkout is required.


PhilippeFerreiraDeSousa:
### Mock transport lazy loading

`sdk/miniapp/src/transport/auto.ts` no longer statically imports `MockTransport`. Instead it keeps the mock fallback behavior but loads the mock transport lazily only when the mock path is actually selected.

This avoids the Bun WebView crash we saw where the generated dev bundle tried to load the mock transport module even though the real runtime path was the mobile postMessage/socket transport. The mock still works for explicit mock mode and no-WebSocket fallback, but it is no longer pulled into the normal starter app path as a static bundled module.

sdk/create-mentra-miniapp/bin/index.ts
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env bun
```
PhilippeFerreiraDeSousa, line 1: This adds the missing generator entrypoint that `create-mentra-miniapp/package.json` already advertised via its `bin` field. The generator is intentionally portable: it copies the template, prompts for project path / package name / display name, validates reverse-DNS package names, replaces placeholders, and leaves the generated app with normal registry dependencies. It does not rewrite deps to local `file:` paths or assume this SDK repo exists next to the generated project.

sdk/create-mentra-miniapp/template/src/App.tsx
```diff
@@ -1,19 +1,29 @@
-import { useSession } from '@mentra/miniapp/react';
+// eslint-disable-next-line import/no-unresolved -- resolved after the template installs dependencies
+import {MentraProvider, useConnected, useSession} from "@mentra/miniapp/react"
```
PhilippeFerreiraDeSousa, line 2: The starter app now uses the React provider expected by `@mentra/miniapp/react`. The old template called `useSession` directly at the top level, which could fail once loaded in the mobile WebView. This also updates the sample button to the current SDK API, `session.display.showTextWall(...)`, instead of the older `session.layouts.showTextWall(...)` shape.

sdk/create-mentra-miniapp/template/miniapp.json
```diff
@@ -1,8 +1,11 @@
 {
+  "$schema": "./node_modules/@mentra/miniapp-cli/schema/miniapp.schema.json",
```
PhilippeFerreiraDeSousa, line 2: The generated manifest now reflects the generated app instead of hardcoded `My Miniapp` metadata. The schema, icon, port, and required `DISPLAY` hardware declaration let the mobile app treat the starter as a real display miniapp immediately after `bun install` / `bun run dev`.

sdk/create-mentra-miniapp/template/build.ts
```diff
@@ -25,7 +29,16 @@ const result = await Bun.build({
   entrypoints: ["./index.html"],
   outdir: distDir,
   target: "browser",
-  plugins: [],
+  plugins: [
```
PhilippeFerreiraDeSousa, line 32: The `app-react-singleton` resolver forces Bun's HTML bundler to resolve `react`, `react-dom`, and `scheduler` from the generated app's own `node_modules`. That matters because the SDK exposes React hooks; if Bun bundles one React copy for the app and another through the SDK package path, the WebView can hit the invalid hook dispatcher failure (`resolveDispatcher().useState` is null).

sdk/miniapp/src/transport/auto.ts
```diff
@@ -36,7 +35,7 @@ export function createTransport(options: CreateTransportOptions = {}): Transport
 
   // Explicit opt-in to mock — useful for stories / framework-level testing.
   if (isMockExplicitlyRequested()) {
-    return new MockTransport()
+    return new LazyMockTransport()
```
PhilippeFerreiraDeSousa, line 38: This keeps mock fallback behavior but loads `MockTransport` only when the mock path is actually selected. The normal mobile WebView path should use postMessage/socket transport, and a static mock import could make Bun include/load the mock module in the dev bundle unnecessarily, which is what produced the bundled-module crash on device.

mobile/src/components/miniapp/MiniappHost.tsx
```diff
@@ -412,10 +421,21 @@ export default function MiniappHost() {
...
-    const data = event.nativeEvent.data
-    localMiniappRuntime.handleRawMessage(packageName, data)
-  }, [])
+  const handleMessage = useCallback(
+    (packageName: string, event: WebViewMessageEvent) => {
+      const data = event.nativeEvent.data
+      const envelope = parseEnvelope(data)
+      const requestType = envelope?.payload ? (envelope.payload as Record<string, unknown>).type : undefined
+
+      if (requestType === MiniappRequestType.CONNECT) {
```
PhilippeFerreiraDeSousa, line 430: On a fresh `CONNECT` from the WebView, the host re-registers the local runtime with the mounted manifest. This is needed for dev reloads and QR re-scans: if the runtime unregisters after missed pings while the WebView is still alive or reconnecting, the next `CONNECT` restores the runtime and its permission/hardware manifest instead of being treated as an unknown app.

sdk/create-mentra-miniapp/bin/index.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  await mkdir(targetDir, {recursive: true})
+  await cp(templateDir, targetDir, {recursive: true})
+
+  const replacements = {
+    packageName: String(packageAnswer),
+    projectName,
+    appName: String(nameAnswer),
+  }
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Addressed in 3905095b1. The generator now escapes scaffold substitutions by context: package.json receives JSON string content, index.html receives HTML text escaping, and App.tsx uses a parseable string-literal placeholder with JSON string-content escaping. The generated miniapp.json is now written from a manifest object with JSON.stringify, which also moves the DISPLAY hardware requirement into generator output instead of the raw template.

PR #2700, opened 2026-05-04
chore: write shell scripts for firmware cli flashing

yagarwal1307:


yagarwal1307:
moved to the new repo

PR #2704, opened 2026-05-05
2.10 hotfixes

fossephate:


aisraelov:
Review pass on `2-10-hotfixes` / PR #2704:

- **Blocker, 10/10:** This likely does not typecheck. `GlassesStatus` adds required `leftMacAddress` / `rightMacAddress`, but the Zustand `initialState` never supplies them. See `mobile/modules/core/src/Core.types.ts` and `mobile/src/stores/glasses.ts`.
- **Blocker, 9/10:** `DeviceStatus` is passed `glassViewClassName`, but `DeviceStatusProps` does not define it. That should be another TS compile error. See `mobile/src/components/home/DeviceStatus.tsx`.
- **High, 8/10:** The Android G2 notification-threading concern looks valid. This PR removes `mainHandler.post` around `handleNotifyData`, so left/right GATT callbacks can now mutate shared `receiveManager.partials` off BLE callback threads. The source key avoids cross-merging payloads, but the shared mutable map is still unsynchronized. See `mobile/modules/core/android/src/main/java/com/mentra/core/sgcs/G2.kt`.
- **Medium, 6/10:** Android `GlassView` now drops flex/size/layout props from the outer wrapper and moves them into nested children. That can collapse or mis-size cards on Android, especially callers using `flex-1`, `h-full`, row layout, etc. See `mobile/src/components/ui/GlassView.tsx`.
- **Medium, 5/10:** iOS G2 auth flags are not reset on disconnect/forget, unlike Android. A stale `leftAuthenticated` or `rightAuthenticated` can make a later reconnect mark fully booted after only one fresh auth response. See `mobile/modules/core/ios/Source/sgcs/G2.swift`.
- **Low/medium, 4/10:** R1 logs the raw `advStart` payload including the glasses MAC on Android and iOS. Not a release blocker by itself, but bad for device identifier exposure. See `mobile/modules/core/android/src/main/java/com/mentra/core/controllers/R1.kt` and `mobile/modules/core/ios/Source/controllers/R1.swift`.

Net: the direction looks useful, but I would not merge until the two TS blockers and the Android G2 notification serialization issue are fixed.

PR #2725, opened 2026-05-06
Add Bluetooth SDK native helper APIs

PhilippeFerreiraDeSousa:
## Summary

This PR is the third stacked PR after #2707 and #2726. It contains the remaining native Bluetooth SDK helper APIs used by the example apps.

It intentionally does not include the RGB LED brightness cleanup anymore. That review belongs in #2726.

## What changed

- Exposes gallery-mode state and control through the Bluetooth SDK surface used by the example apps.
- Defers iOS SDK store callbacks so listener delivery does not synchronously mutate while applying store updates.
- Adds default-device helpers for saving, reading, clearing, and connecting the preferred glasses device.
- Adds typed Android media-volume helpers for reading and setting glasses media volume.
- Updates SDK docs to describe these helper APIs.

## Stack / review order

1. Base PR: #2707 (`philippe/os-1178-bluetooth-sdk-example-app-changes`)
2. Intermediate PR: #2726 (`codex/remove-broken-rgb-led-brightness`)
3. This PR: `codex/remove-rgb-led-brightness-workaround`

## Review notes

- The branch name is historical. The current diff is about gallery mode, default-device helpers, iOS callback ordering, and Android media volume helpers.
- RGB LED brightness observations and behavior changes are documented on #2726 instead.

## Validation

- `cd mobile/android && ./gradlew :mentra-bluetooth-sdk:compileDebugKotlin`
- `cd asg_client && ./gradlew :app:compileDebugJavaWithJavac`

PR #2707, opened 2026-05-05
Expose typed Bluetooth SDK request models for examples

PhilippeFerreiraDeSousa:
## Summary

This PR contains the Bluetooth SDK request/model cleanup needed by the redesigned partner/example apps. It is stacked on top of `codex/remove-button-mode-sdk-api`.

The main goal is to make native Android, native iOS, and React Native callers use a clearer typed SDK surface for camera, streaming, keep-alive, microphone, and RGB LED control instead of loose maps and unconstrained strings. The Expo bridge still accepts the existing MentraOS payloads and normalizes them before forwarding to native, so this should not force a broad MentraOS TypeScript call-site rewrite.

## What changed

- Exposes RGB LED control through the Android and Swift SDK facades so partner apps can call it directly.
- Adds typed Android, Swift, and TypeScript request/result models for photo requests, stream requests, stream keep-alives, RGB LED actions/colors, photo sizes, compression values, and related events.
- Makes photo upload auth tokens optional instead of requiring callers to pass an empty string when no auth header is needed.
- Passes iOS microphone configuration through the Expo bridge, including PCM data, transcript, and VAD flags.
- Normalizes MentraOS cloud/socket payloads before forwarding them into the typed SDK surface.
- Keeps RGB LED brightness out of the public SDK/example-app surface because the brightness behavior was found to be unsupported in practice.
- Updates the native API planning docs with the tightened value boundaries.

## Stack / review order

1. Base: `codex/remove-button-mode-sdk-api`
2. This PR: `philippe/os-1178-bluetooth-sdk-example-app-changes`
3. Follow-up: #2726 removes the broken ASG-client RGB LED brightness command path.
4. Follow-up: #2725 adds unrelated native helper APIs used by the example apps.

## Review notes

- This PR is the typed SDK surface cleanup. It should not contain the ASG-client brightness workaround removal; that is isolated in #2726.
- The partner kit/example app UI and app logic live in the separate partner-kit repository.
- The RGB LED API still supports color and timing/pattern controls. Brightness is intentionally not exposed as a partner-facing control.

## Validation

- `cd mobile/android && ./gradlew :mentra-bluetooth-sdk:compileDebugKotlin`
- `cd asg_client && ./gradlew :app:compileDebugJavaWithJavac`

PR #2726, opened 2026-05-06
Stop pre-setting RGB LED brightness

PhilippeFerreiraDeSousa:
## Summary

This PR isolates the Mentra Live RGB LED brightness behavior fix from the broader Bluetooth SDK example-app API work in #2707.

We concluded RGB LED brightness should not be exposed as a partner-facing SDK feature right now. The visible problem we still need to fix is narrower: before sending RGB LED on/flash/solid commands, the ASG client pre-called the brightness setter, and that caused an initial red blink before the requested LED behavior. This PR stops doing that pre-call.

## What changed

- Stops calling the internal RGB LED brightness setter before RGB LED on/flash/solid commands.
- Keeps RGB LED color and timing control intact.
- Keeps the internal brightness setter and hardware-manager plumbing in place; this PR does not delete internal APIs.
- Does not document RGB LED brightness as a public SDK capability.

## Stack / review order

1. Base PR: #2707 (`philippe/os-1178-bluetooth-sdk-example-app-changes`)
2. This PR: `codex/remove-broken-rgb-led-brightness`
3. Follow-up: #2725 (`codex/remove-rgb-led-brightness-workaround`) contains unrelated native helper APIs and should be reviewed after this PR.

## Review notes

- This is intentionally not a public SDK brightness feature.
- The public SDK/example-app RGB LED API remains color + timing/pattern only.

## Validation


PhilippeFerreiraDeSousa:
Ok what we merge about this PR is very much up in the air. The current observed behavior was undesirable

PhilippeFerreiraDeSousa:
I converted that to draft. We don't want the red blink when doing other led controls but for the pure brightness setter feature that is broken, we don't want public sdk functions but no need to delete internal setters.

PhilippeFerreiraDeSousa:
Ok so for the latest version here is what is happening : we just remove the setBrightness preflight calls when doing led control operations because this does not set the brightness (broken) so each time we do an led operation it retries to set it and it just blinks once red

PhilippeFerreiraDeSousa:

PR #2747, opened 2026-05-11
[codex] Fix gallery mode during first 30s after ASG connect

PhilippeFerreiraDeSousa:
## Summary

Fixes gallery mode for the first 30 seconds after ASG connects to the phone.

Before this change, ASG always initialized its phone connection state as `isConnected=false` and waited for the first phone heartbeat to flip it to true. The phone heartbeat interval is 30 seconds, but the phone sends user settings immediately after the `phone_ready` -> `glasses_ready` handshake. That meant ASG received `save_in_gallery_mode=false` while still believing the phone was disconnected.

The camera-button handler only suppresses local capture when both conditions are true:

```java
!isSaveInGalleryMode && isConnected
```

Because `isConnected` was still false during that first 30-second window, the handler took the disconnected fallback path and saved media locally even though gallery mode was inactive.

## Pre-existing behavior

1. ASG starts heartbeat monitoring and sets `isConnected=false`.
2. Phone sends `phone_ready`.
3. ASG responds with `glasses_ready`.
4. Phone starts the heartbeat scheduler, but the first ping is delayed by 30 seconds.
5. Phone immediately sends user settings, including `save_in_gallery_mode`.
6. For the first 30 seconds after connect, ASG can have `save_in_gallery_mode=false` and `isConnected=false` at the same time.
7. A camera-button press during that window saves locally via the disconnected fallback path.

For reconnects, this same first-30-seconds behavior happens whenever ASG has reset or expired its connection state before the new handshake. If ASG is still inside the previous 35-second heartbeat timeout window and still has `isConnected=true`, then the specific false-state window is already avoided by the old state.

## New behavior

ASG marks the phone connection active as soon as the explicit handshake completes: after handling `phone_ready` and successfully sending `glasses_ready`.

Heartbeat still owns ongoing liveness after that point. It continues to refresh the timeout on each ping and marks the phone disconnected if pings stop.

The result is that `save_in_gallery_mode=false` suppresses local button capture immediately after connection, without waiting for the first scheduled heartbeat.

## Other cleanup

- Tightened `AsgClientServiceManager` so its `AsgClientService` reference is a required constructor invariant instead of carrying unreachable null branches.

## Validation

- Built ASG debug APK with `./gradlew assembleDebug --no-daemon`.
- Installed the updated ASG APK on a connected Mentra Live.
- Connected the Partner Kit Android example on a Samsung Note 20.
- Verified before the first 30-second heartbeat that ASG logs `save_in_gallery_mode` inactive and a synthetic camera-button press reports `Gallery Mode: INACTIVE, Connection State: CONNECTED`.
- Ran `git diff --check` clean before committing.

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/service/core/AsgClientService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/service/core/AsgClientService.java
@@ -1261,6 +1261,14 @@ public class AsgClientService extends Service implements NetworkStateListener, B
     }
 
     /**
+     * Handle the phone_ready/glasses_ready handshake completing over Bluetooth.
+     */
+    public void onPhoneReadyHandshakeComplete() {
+        Log.d(TAG, "📱 Phone ready handshake complete - marking phone connection active");
+        resetHeartbeatTimeout();
+    }
+
+    /**
      * Handle service heartbeat received from MentraLiveSGC
      */
     public void onServiceHeartbeatReceived() {
@@ -1599,4 +1607,4 @@ public class AsgClientService extends Service implements NetworkStateListener, B
             Log.e(TAG, "🗑️ Error cleaning up orphaned BLE transfers", e);
         }
     }
-} 
+}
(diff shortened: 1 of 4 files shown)


PhilippeFerreiraDeSousa:

PR #2752, opened 2026-05-12
Refresh Bluetooth signal strength every 10 seconds

PhilippeFerreiraDeSousa:
## Summary

This updates the Mentra Bluetooth SDK so the glasses-side `signalStrength` status is refreshed by the phone BLE stack on a 10-second timer instead of only being populated opportunistically during connection setup.

## What changed

- Starts periodic RSSI reads after the Mentra Live BLE GATT setup finishes on Android and iOS.
- Stops RSSI polling and resets `signalStrength` on disconnect/cleanup.
- Adds typed `signalStrengthUpdatedAt` to the Android, iOS, and TypeScript SDK surfaces so apps can tell when the value was last refreshed.
- Ensures status callbacks include the current `signalStrength` whenever `signalStrengthUpdatedAt` changes, so subscribers receive a real signal-strength refresh event even if Android/iOS report the same dBm value twice.

## Verification

- `./gradlew :mentra-bluetooth-sdk:publishToMavenLocal` from `mobile/android` completed successfully.
- Rebuilt and installed the Android partner-kit example on the connected Note 20 (`R5CN80W6MZA`).
- Verified the Device tab updated from `-54 dBm / updated 11:48:52` to later readings including `-61 dBm / updated 11:52:30`.
- `bun run build` in `mobile/modules/bluetooth-sdk` completed successfully and regenerated the SDK declaration output with `signalStrengthUpdatedAt`.

## Notes

- `bun run compile` in `mobile` still fails on existing repo-wide TypeScript errors outside this change, beginning with `e2e-tests/ui` dynamic import config and missing `bun:test`/Vite/Recharts types.
- The iOS example Swift compile path got through, but simulator linking still fails because the installed GStreamer framework is built for iOS device, not iOS Simulator.

### DIFF
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/DeviceStore.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/DeviceStore.kt
@@ -82,6 +82,8 @@ object GlassesStore {
         store.set("glasses", "hotspotPassword", "")
         store.set("glasses", "hotspotGatewayIp", "")
         store.set("glasses", "bluetoothName", "")
+        store.set("glasses", "signalStrength", -1)
+        store.set("glasses", "signalStrengthUpdatedAt", 0L)
         store.set("glasses", "controllerConnected", false)
         store.set("glasses", "controllerFullyBooted", false)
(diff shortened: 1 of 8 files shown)


PhilippeFerreiraDeSousa:

PR #2749, opened 2026-05-12
Add typed Bluetooth SDK status models

PhilippeFerreiraDeSousa:
## Summary

This PR adds typed status models to the Bluetooth SDK after the gallery-mode connection fix branch.

It replaces public native SDK access to string-keyed status dictionaries for the stable glasses/Bluetooth status surfaces with typed Android and iOS models, while keeping map serialization internal for the React Native bridge.

## Changes

- Add typed Android status snapshots and partial update models for glasses and Bluetooth state.
- Add typed iOS status accessors and update helpers for the same status surfaces.
- Keep React Native bridge payloads compatible by converting typed models back to maps internally.
- Add typed Wi-Fi scan/device search result models used by status APIs.

## Validation

- `cd mobile && bun install --frozen-lockfile`
- Android SDK Kotlin compile was run before pushing this stack.
- Partner Kit Android example was rebuilt, installed, and launched on the Note 20 after the typed SDK changes.

PR #2738, opened 2026-05-08
bluetooth-sdk: declare FOREGROUND_SERVICE_DATA_SYNC permission

isaiahb:
## Summary

Adds the missing `android.permission.FOREGROUND_SERVICE_DATA_SYNC` to `mobile/modules/bluetooth-sdk/android/src/main/AndroidManifest.xml`. Without it, every host app targeting Android 14+ (targetSdk 34+) crashes on first launch.

## Why

The library's `ForegroundService` is declared with three foreground-service types:

```xml
<service android:name="com.mentra.core.services.ForegroundService"
         android:foregroundServiceType="dataSync|connectedDevice|microphone"/>
```

But the manifest only declares two of the three matching `uses-permission` entries — `FOREGROUND_SERVICE_DATA_SYNC` is missing. On Android 14+, `startForeground()` validates that the **runtime type the service starts with** has a matching app-level permission, regardless of which sibling types are also declared. The SDK calls `startForegroundWithAutoDetectedType(...)` which can pick `dataSync`, and that path throws:

```
java.lang.SecurityException: Starting FGS with type dataSync
    callerApp=ProcessRecord{...:com.example.host/u0a117} targetSDK=36
    requires permissions: all of the permissions [android.permission.FOREGROUND_SERVICE_DATA_SYNC]
  at android.app.Service.startForeground(Service.java:862)
  at com.mentra.core.services.ForegroundService.startForegroundWithAutoDetectedType(Foreground.kt:52)
  at com.mentra.core.services.ForegroundService.onCreate(Foreground.kt:24)
```

…which propagates up as `RuntimeException: Unable to create service com.mentra.core.services.ForegroundService` and kills the host process.

## Repro

Running the partner kit's `examples/react-native` app on a Pixel 8 (Android 15, targetSdk 36):

1. `expo run:android` builds and installs cleanly.
2. App opens to the dev launcher, JS bundle loads (727 modules in 312ms).
3. App immediately crashes — `RuntimeException: Unable to create service ... Caused by SecurityException`.
4. Stuck on "Reloading…" forever because RN reload tries to recreate the service.

With this one-line manifest change applied, same APK boots through to the Device tab and shows the disconnected Mentra Live, Scan/Connect/Display Hello/Clear actions, and the Camera/Stream/System/Console tabs.

## Risk

- One additional `<uses-permission>` declaration. No code change.
- Permission is normal-protection (granted at install time). No runtime prompt impact.
- Brings the manifest in line with what the service already declares it needs at runtime.

[...]

PR #2688, opened 2026-04-29
Reapply Bluetooth SDK naming cleanup

PhilippeFerreiraDeSousa:
## Summary

This PR reverts `f8ab48f19` (`Restore Core naming for review separation`) so the Bluetooth SDK naming cleanup can be reviewed and merged separately after the main feature branch lands.

It intentionally keeps the file/module move from the feature branch and re-applies only the code-facing rename layer, including the SDK-facing module/status/store naming that was temporarily undone to reduce review and merge-conflict noise.

## Why

The base feature branch now preserves the old Core-facing names to make the current PR easier to review and easier to merge alongside the parallel TypeScript work. This follow-up branch gives us a clean, isolated PR that can bring the SDK-facing names back later without mixing that rename churn into the larger refactor review.

## Validation

- `cd mobile && bun run test`
- `cd mobile/android && ./gradlew :mentra-bluetooth-sdk:compileDebugKotlin`

PR #2683, opened 2026-04-29
Handle notification titles as CharSequence

PhilippeFerreiraDeSousa:
## Summary

- Read `android.title` from notification extras as a `CharSequence` instead of a `String`.
- Preserve notification titles from apps that provide `SpannableString` titles.
- Avoid Android `Bundle` cast warnings while keeping the moved Crust notification listener behavior unchanged.

## Root Cause

During Android notification testing with the Notification Test app, Crust received and processed the notification, but Android logged:

```text
Key android.title expected String but value was a android.text.SpannableString
java.lang.ClassCastException: android.text.SpannableString cannot be cast to java.lang.String
```

The old listener already used `getCharSequence("android.text")` for body text, but title still used `getString("android.title")`. Some Android notifications expose title as styled text, so `getString` returns the default value and logs a cast warning.

## Validation

- `./gradlew :crust:compileDebugKotlin`
- `./gradlew :app:installDebug` on connected Android device `SM_N986U`
- Re-tested with Notification Test app:
  - Crust logged `Received notification from com.mand.notitest`
  - Crust logged `Buffering notification (200ms delay)`
  - Crust logged `Processing buffered notification from Notification Test`
  - No `Bundle`, `ClassCastException`, or `android.title expected String` warning appeared in the filtered log after the fix.

PR #2765, opened 2026-05-12
Bluetooth SDK aggregate review branch

PhilippeFerreiraDeSousa:
## Scope

This PR is an aggregate review branch on top of `mentra-miniapp-sdk-2`. It merges the Bluetooth SDK stack and the relevant side fixes into one branch so the combined result can be reviewed and tested without modifying the original stacked branches.

Base:
- `mentra-miniapp-sdk-2`

Merged into this branch, in order:
- `philippe-fix/fgs-data-sync-permission` / #2738: already present in `mentra-miniapp-sdk-2`; confirmed included for the Android 14+ foreground-service permission fix.
- `codex/remove-button-mode-sdk-api` / #2681: removes the stale button-mode Bluetooth SDK API surface.
- `philippe/os-1178-bluetooth-sdk-example-app-changes` / #2707: exposes typed request models and SDK APIs needed by the example apps.
- `codex/remove-broken-rgb-led-brightness` / #2726: stops pre-setting RGB LED brightness before LED commands, avoiding the red blink caused by the broken brightness path.
- `codex/remove-rgb-led-brightness-workaround` / #2725: adds the native helper APIs used by the Bluetooth SDK examples while keeping the broken public LED brightness setter out of the API surface.
- `philippe-fix-gallery-mode` / #2747: fixes gallery-mode behavior during the first 30 seconds after ASG connection.
- `codex/bluetooth-sdk-typed-status` / #2749: adds typed Bluetooth SDK status models.
- `codex/bluetooth-sdk-rssi-refresh` / #2752: refreshes Bluetooth signal strength periodically.
- `codex/fix-notification-title-charsequence` / #2683: handles Android notification titles as `CharSequence`.

Explicitly excluded:
- `codex/reapply-bluetooth-sdk-renames` / #2688. The naming cleanup remains separate and should not be part of this aggregate review branch.

## Why

The individual Bluetooth SDK PRs are stacked and some of them depend on earlier branches. This aggregate branch gives reviewers one place to inspect the final combined behavior on top of the Island/Main App SDK 2 base while preserving the original PR branches unchanged.

## Conflict Resolution Notes

There was one conflict while merging `philippe/os-1178-bluetooth-sdk-example-app-changes`:

- `mobile/modules/bluetooth-sdk/ios/BluetoothSdkModule.swift`

The resolution keeps the Island `JSCExperiment.maybeAutoBenchmark()` hook and the Bluetooth SDK initialization in the later `OnCreate` block, while removing the obsolete duplicate early `OnCreate` block. This preserves Island benchmark behavior and Bluetooth SDK startup behavior.

## Validation

Local verification performed on this aggregate branch:

- Confirmed the worktree is clean after merging.
- Confirmed all intended branch heads are ancestors of this branch.
- Confirmed `codex/reapply-bluetooth-sdk-renames` is not an ancestor of this branch.
- Ran a strict conflict-marker scan with `rg -n "^(<<<<<<<|=======|>>>>>>>|\\|\\|\\|\\|\\|\\|\\|)( |$)" mobile asg_client docs`.
- Ran `git diff --check origin/mentra-miniapp-sdk-2..HEAD`.

Full mobile builds/tests were not run for this aggregation branch yet; the intent here is to provide a combined review target first.


mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/BluetoothSdkModule.kt
```diff
@@ -360,5 +499,162 @@ class CoreModule : Module() {
         AsyncFunction("extractTarBz2") { sourcePath: String, destinationPath: String ->
             com.mentra.core.stt.STTTools.extractTarBz2(sourcePath, destinationPath)
         }
+
+        // MARK: - Settings Navigation
+
+        AsyncFunction("openBluetoothSettings") {
```
fossephate: these functions have been moved to the crust module

mobile/modules/bluetooth-sdk/ios/Source/Bridge.swift
```diff
@@ -230,12 +242,14 @@ class Bridge {
     }
 
     static func sendWifiStatusChange(connected: Bool, ssid: String?, localIp: String?) {
-        let event: [String: Any] = [
-            "connected": connected,
-            "ssid": ssid,
-            "local_ip": localIp,
-        ]
-        Bridge.sendTypedMessage("wifi_status_change", body: event)
+        guard let status = MentraWifiStatus.fromStoreFields(
```
fossephate: we shouldn't be naming types with "Mentra" as it doesn't add any contextual value and just lengthens the name

mobile/src/stores/glasses.ts
```diff
@@ -10,7 +10,14 @@ export function isGlassesLinkLayerBusy(connectionState: string | undefined): boo
 
 interface GlassesState extends GlassesStatus {
   wifiStatusKnown: boolean
-  setGlassesInfo: (info: Partial<GlassesStatus>) => void
+  wifiConnected: boolean
```
fossephate: double check there are no duplicate keys being added here

mobile/src/stores/glasses.ts
```diff
@@ -45,6 +126,13 @@ export const getGlasesInfoPartial = (state: GlassesStatus) => {
 interface GlassesStore extends GlassesStatus {
   mtkUpdatedThisSession: boolean
   wifiStatusKnown: boolean
+  wifiConnected: boolean
```
fossephate: same as previous comment about duplicate keys

mobile/src/services/SocketComms.ts
```diff
@@ -18,6 +23,29 @@ import restComms from "@/services/RestComms"
...
+}
+
+function normalizeRgbLedAction(value: unknown): RgbLedAction {
+  return typeof value === "string" && RGB_LED_ACTIONS.has(value as RgbLedAction) ? (value as RgbLedAction) : "off"
+}
+
+function normalizeRgbLedColor(value: unknown): RgbLedColor | null {
+  return typeof value === "string" && RGB_LED_COLORS.has(value as RgbLedColor) ? (value as RgbLedColor) : null
+}
+
```
fossephate: this code should probably be moved to BluetoothSdkModule.types.ts / or something similar

mobile/src/app/miniapps/settings/glasses.tsx
```diff
@@ -41,7 +41,9 @@ function DeviceSettings() {
   const applets = useApps()
   const features: Capabilities = getModelCapabilities(defaultWearable)
 
-  const wifiLocalIp = useGlassesStore((state) => state.wifiSsid)
```
PhilippeFerreiraDeSousa, line 44: Little bug here. I think the time we have an ssid but not local ip is small. But having a local ip is what is needed for features to work

fossephate (approved):

PR #2770, opened 2026-05-13
Make Bluetooth SDK status getters synchronous

PhilippeFerreiraDeSousa:
## Summary

It makes the React Native Bluetooth SDK homogeneous across Android and Swift for cached status reads:

- Android already exposes `getGlassesStatus`, `getCoreStatus`, and `getDefaultDevice` as synchronous Expo `Function`s.
- Swift now does the same instead of exposing cached snapshots through `AsyncFunction`.
- The TypeScript SDK surface returns plain values again, rather than wrapping the native result in `Promise.resolve(...)`.

## How it works

The iOS issue was not that the data itself required async work. These getters read cached store snapshots. The async shape came from routing those reads through `MainActor.run` because the whole store was main-actor isolated.

This PR makes the cached store safe to read synchronously by protecting `ObservableStore` with a lock. `GlassesStore` itself is no longer globally main-actor isolated, but `apply(...)` and the dashboard hardware side effects remain `@MainActor`, because those paths can drive CoreManager / glasses side effects.

That keeps the real control plane actor-isolated while allowing snapshot getters to be synchronous on both Android and iOS.

## MentraOS impact

MentraOS previously used `await CoreModule.getCoreStatus()` and `await CoreModule.getGlassesStatus()` even when Android returned plain values. This PR removes those awaits to match the SDK contract. This is a pre-existing MentraOS call-site shape change, which is why it is separated into this PR instead of staying bundled into `codex/bluetooth-sdk-device-target-api`.

## Verification

- `bun run build` in `mobile/modules/bluetooth-sdk`
- `./gradlew :mentra-bluetooth-sdk:compileDebugKotlin --console=plain` in `mobile/android`
- `xcodebuild -workspace ios/Mentra.xcworkspace -scheme MentraBluetoothSDK -configuration Debug -sdk iphonesimulator -destination generic/platform=iOS\ Simulator build CODE_SIGNING_ALLOWED=NO` in `mobile`
- `git diff --check`

Note: repo-wide `bun compile` still fails with the existing broad mobile typecheck error set; no SDK getter-specific errors were observed.

### DIFF
--- a/mobile/modules/bluetooth-sdk/ios/BluetoothSdkModule.swift
+++ b/mobile/modules/bluetooth-sdk/ios/BluetoothSdkModule.swift
@@ -73,22 +73,16 @@ public class CoreModule: Module, MentraBluetoothSDKDelegate {
 
         // MARK: - Observable Store Functions
 
-        AsyncFunction("getGlassesStatus") {
-            await MainActor.run {
-                self.bluetoothSdk().glassesStatus.dictionary
-            }
+        Function("getGlassesStatus") {
+            MentraGlassesStatus(values: GlassesStore.shared.store.getCategory("glasses")).dictionary
         }
 
-        AsyncFunction("getCoreStatus") {
-            await MainActor.run {
-                self.bluetoothSdk().bluetoothStatus.values
-            }
+        Function("getCoreStatus") {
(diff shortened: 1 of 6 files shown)


PhilippeFerreiraDeSousa:
To be discussed with @fossephate

PhilippeFerreiraDeSousa:
Better to make them both asynchronous

PR #2784, opened 2026-05-14
[codex] Fix mobile type checking

PhilippeFerreiraDeSousa:
## Summary
- Fix the mobile TypeScript compile errors while preserving existing runtime behavior where possible.
- Add narrow type declarations and missing module shims for untyped/native packages and broken internal imports.
- Add `bun run compile` to the mobile PR quality workflow so type checking runs before Jest on pull requests.
- Document type errors that look like possible behavior bugs in `mobile/TYPECHECK_BEHAVIOR_NOTES.md`.

## Validation
- `bun compile`
- `bun run test`
- `git diff --check`

Note: `bun test` was intentionally not used as the validation command because it invokes Bun's test runner directly and crashes on React Native Flow syntax in `node_modules/react-native/index.js`; the repo's CI path uses `bun run test` / Jest.


.github/workflows/augmentos-manager-jest.yml
```diff
@@ -44,6 +44,10 @@ jobs:
         working-directory: ./mobile
         run: bun install
 
+      - name: Run TypeScript type check
```
PhilippeFerreiraDeSousa, line 47: This CI change is here so the same compiler command that this PR fixes now runs on mobile PRs before Jest.

Original failure class:
```text
bun compile
# tsc --noEmit -p . --pretty
# failed with 244 TypeScript errors before this PR
```
This workflow step does not correspond to one source-line compiler error; it prevents this class of mobile type errors from landing again.

mobile/tsconfig.json
```diff
@@ -32,5 +32,11 @@
     }
   },
   "include": ["**/*.ts", "**/*.tsx", ".expo/types/**/*.ts", "expo-env.d.ts"],
-  "exclude": ["node_modules", "test/**/*"]
+  "exclude": [
+    "node_modules",
+    "test/**/*",
+    "e2e-tests/ui/**/*",
```
PhilippeFerreiraDeSousa, line 37: I excluded the standalone dashboard and module test sources from the mobile app typecheck because they have different toolchains/dependencies and were being pulled into `mobile/tsconfig.json` as if they were app code.

Original errors included:
```text
e2e-tests/ui/src/App.tsx(9,32): error TS1323: Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node16', 'node18', 'node20', or 'nodenext'.
modules/miniapp/src/envelope.test.ts(2,38): error TS2307: Cannot find module 'bun:test' or its corresponding type declarations.
```
I also moved the module target to `esnext` so mobile code using dynamic import-compatible settings is checked under the same modern module mode Expo expects.

mobile/src/types/untyped-modules.d.ts
```diff
@@ -0,0 +1,177 @@
+declare module "bzip2" {
```
PhilippeFerreiraDeSousa, line 1: This declaration file covers runtime dependencies that are present but do not ship usable TypeScript declarations. I typed only the APIs used by this app instead of adding broad `any` declarations.

Original errors included:
```text
src/services/TarBz2Extractor.ts(1,19): error TS2307: Cannot find module 'bzip2' or its corresponding type declarations.
src/services/TarBz2Extractor.ts(3,17): error TS7016: Could not find a declaration file for module 'tar-js'.
src/components/mirror/GlassesDisplayMirror.tsx(3,44): error TS7016: Could not find a declaration file for module 'react-native-canvas'.
src/effects/TranscriptionsListener.tsx(1,36): error TS2307: Cannot find module 'expo-speech-transcriber' or its corresponding type declarations.
```

mobile/src/types/untyped-modules.d.ts
```diff
@@ -0,0 +1,177 @@
...
+  export function assetModelPath(path: string): ModelPathConfig
+  export function fileModelPath(path: string): ModelPathConfig
+  export function getDefaultModelPath(): string
+  export function getAssetPackPath(packName: string): Promise<string | null>
+  export function listAssetModels(): Promise<SherpaModelInfo[]>
+  export function listModelsAtPath(path: string): Promise<SherpaModelInfo[]>
+  export function resolveModelPath(path: ModelPathConfig): Promise<string>
+}
+
+declare module "react-native-sherpa-onnx/stt" {
```
PhilippeFerreiraDeSousa, line 109: The Sherpa ONNX declarations are intentionally local because the package is used by this app but was untyped from the compiler's perspective. The shape here matches the calls in `SherpaTest` closely enough to remove implicit `any`s without changing that feature.

Original errors included:
```text
src/effects/SherpaTest.tsx(24,8): error TS2307: Cannot find module 'react-native-sherpa-onnx' or its corresponding type declarations.
src/effects/SherpaTest.tsx(39,8): error TS2307: Cannot find module 'react-native-sherpa-onnx/stt' or its corresponding type declarations.
src/effects/SherpaTest.tsx(986,43): error TS7006: Parameter 'samples' implicitly has an 'any' type.
```

mobile/src/components/ignite/Icon.tsx
```diff
@@ -39,13 +40,18 @@ import {
 import {ShoppingBagIcon, HomeIcon, GridIcon} from "@/components/icons"
 import {useAppTheme} from "@/contexts/ThemeContext"
 
-export type IconTypes = keyof typeof iconRegistry
+export type IconTypes = keyof typeof iconRegistry | (string & {})
```
PhilippeFerreiraDeSousa: I widened `IconTypes` instead of adding new icon registry entries because adding registry entries changes rendering behavior. The existing runtime behavior for unknown names is preserved, and the behavior concern is documented separately.

Original errors included:
```text
src/app/miniapps/settings/data-export.tsx(342,42): error TS2322: Type '"copy"' is not assignable to type ...
src/app/miniapps/settings/data-export.tsx(350,42): error TS2820: Type '"share-2"' is not assignable to type ... Did you mean '"share"'?
src/components/glasses/Gallery/GalleryScreen.tsx(895,19): error TS2322: Type '"download-circle-outline"' is not assignable to type ...
```

mobile/src/components/ignite/Icon.tsx
```diff
@@ -39,13 +40,18 @@ import {
...
   /**
    * The name of the icon
    */
-  name: IconTypes
+  name?: IconTypes
+
+  /**
+   * Legacy prop accepted by older call sites.
+   */
+  icon?: IconTypes
```
PhilippeFerreiraDeSousa: The legacy `icon` prop is typed but not wired into rendering on purpose. Some call sites already passed `icon`, but changing it to render would be a behavior change beyond type cleanup.

Original errors included:
```text
src/components/ignite/ListItem.tsx(176,9): error TS2322: Type '{ size: number; icon: ... }' is not assignable to type 'IntrinsicAttributes & Omit<ViewProps, "style"> & BaseIconProps'.
  Property 'icon' does not exist on type 'IntrinsicAttributes & Omit<ViewProps, "style"> & BaseIconProps'.
src/components/settings/SelectSetting.tsx(89,19): error TS2322: Type '{ icon: string; size: number; color: ... }' is not assignable to type ...
  Property 'icon' does not exist on type ...
```

mobile/src/theme/typography.ts
```diff
@@ -94,6 +94,9 @@ const fonts = {
   },
 }
 
+const primaryFonts = Platform.select({ios: fonts.sfProRounded, android: fonts.redHatDisplay}) ?? fonts.redHatDisplay
```
PhilippeFerreiraDeSousa, line 97: `Platform.select` can return `undefined` in the type system, which then poisoned the Text style types. I added explicit fallback font maps instead of loosening the Text component types.

Original errors included:
```text
src/components/ignite/Text.tsx(158,42): error TS2769: No overload matches this call.
  Argument of type '{ light: string | undefined; ... } | undefined' is not assignable to parameter of type '{}'.
src/components/ignite/TextField.tsx(252,15): error TS18048: 'typography.primary' is possibly 'undefined'.
```

mobile/src/components/glasses/OtaProgressSection.tsx
```diff
@@ -3,18 +3,34 @@ import {View} from "react-native"
 import {Text} from "@/components/ignite"
 import {useAppTheme} from "@/contexts/ThemeContext"
 import {ThemedStyle} from "@/theme"
-import {OtaProgress} from "@/utils/CoreStatusParser"
+
+type LegacyOtaStep = {
```
PhilippeFerreiraDeSousa: The old import target does not exist, so I inlined the legacy progress shape that this component already reads. I kept the prop as `unknown` and cast locally to avoid pretending the current SDK `OtaProgress` really has this nested shape.

Original error:
```text
src/components/glasses/OtaProgressSection.tsx(6,27): error TS2307: Cannot find module '@/utils/CoreStatusParser' or its corresponding type declarations.
```
This also explains why `TYPECHECK_BEHAVIOR_NOTES.md` calls out the OTA progress shape as a possible runtime issue.

mobile/src/services/MiniSockets.ts
```diff
@@ -452,3 +452,13 @@
 
 // const miniSockets = MiniSockets.getInstance()
 // export default miniSockets
+
+export type MiniSocketTextHandler = (clientId: number, text: string) => void
+
+const miniSockets = {
```
PhilippeFerreiraDeSousa: The file had its implementation commented out but `MantleManager` still imports a default export and calls methods on it. I added a no-op export surface instead of uncommenting/reimplementing socket behavior in a type-checking PR.

Original errors:
```text
src/services/MantleManager.ts(9,8): error TS1192: Module '.../src/services/MiniSockets' has no default export.
src/services/MantleManager.ts(251,34): error TS7006: Parameter 'clientId' implicitly has an 'any' type.
src/services/MantleManager.ts(251,44): error TS7006: Parameter 'text' implicitly has an 'any' type.
```
The no-op behavior is called out in `TYPECHECK_BEHAVIOR_NOTES.md`.

mobile/src/components/ui/InfoSection.tsx
```diff
@@ -0,0 +1,59 @@
...
+  value?: string | number | null
+}
+
+type InfoSectionProps = {
+  title: string
+  items: InfoSectionItem[]
+  style?: ViewStyle
+}
+
+export default function InfoSection({title, items, style}: InfoSectionProps) {
```
PhilippeFerreiraDeSousa: This module was imported by `DeviceInformation` but did not exist. I added the smallest component matching the call shape rather than redesigning that screen.

Original error:
```text
src/components/glasses/info/DeviceInformation.tsx(3,25): error TS2307: Cannot find module '@/components/ui/InfoSection' or its corresponding type declarations.
```

mobile/src/app/applet/settings.tsx
```diff
@@ -301,12 +301,11 @@ export default function AppSettings() {
     }))
 
     // Build an array of settings to send.
-    restComms
-      .updateAppSetting(packageName, {key, value})
+    ;(restComms.updateAppSetting(packageName, {key, value}) as unknown as Promise<void>)
```
PhilippeFerreiraDeSousa: `RestComms.updateAppSetting` returns `AsyncResult`, but the existing code uses it like a catchable Promise. I cast through `unknown` to preserve the current `.then(...).catch(...)` behavior instead of rewriting the async flow in this PR.

Original errors:
```text
src/app/applet/settings.tsx(305,25): error TS2345: Argument of type 'string | string[]' is not assignable to parameter of type 'string'.
src/app/applet/settings.tsx(309,8): error TS2339: Property 'catch' does not exist on type 'PromiseLike<void>'.
src/app/applet/settings.tsx(309,15): error TS7006: Parameter 'error' implicitly has an 'any' type.
```

mobile/src/stores/glasses.ts
```diff
@@ -9,6 +9,7 @@ export function isGlassesLinkLayerBusy(connectionState: string | undefined): boo
 }
 
 interface GlassesState extends GlassesStatus {
+  wifiConnected?: boolean
```
PhilippeFerreiraDeSousa: This optional legacy field is typed because the OTA check screen already reads it. I did not remap the screen to `wifi.state === "connected"` because that would change behavior, so the possible runtime issue is documented separately.

Original errors:
```text
src/app/ota/check-for-updates.tsx(26,65): error TS2551: Property 'wifiConnected' does not exist on type 'GlassesState'. Did you mean 'connected'?
src/app/ota/check-for-updates.tsx(284,37): error TS2551: Property 'wifiConnected' does not exist on type 'GlassesState'. Did you mean 'connected'?
```

mobile/src/components/glasses/NexDeveloperSettings.tsx
```diff
@@ -8,14 +8,19 @@ import ToggleSetting from "@/components/settings/ToggleSetting"
...
-import {translate} from "@/i18n/translate"
+import {translate, TxKeyPath} from "@/i18n"
 import {useGlassesStore} from "@/stores/glasses"
 import {SETTINGS, useSetting} from "@/stores/settings"
 import {ThemedStyle} from "@/theme"
 import showAlert from "@/utils/AlertUtils"
 import {MOCK_CONNECTION} from "@/utils/Constants"
 import GlobalEventEmitter from "@/utils/GlobalEventEmitter"
 
+const nexCoreModule = CoreModule as typeof CoreModule & {
```
PhilippeFerreiraDeSousa: These native methods are used by this screen but are missing from the Bluetooth SDK TypeScript surface. I added a local narrowed cast at the call site rather than extending global SDK types, because that keeps the assumption scoped to this screen.

Original errors:
```text
src/components/glasses/NexDeveloperSettings.tsx(367,24): error TS2339: Property 'displayImage' does not exist on type 'CoreModule'.
src/components/glasses/NexDeveloperSettings.tsx(396,24): error TS2339: Property 'setLc3AudioEnabled' does not exist on type 'CoreModule'.
src/components/glasses/NexDeveloperSettings.tsx(430,32): error TS2345: Argument of type '"settings:screenSettings"' is not assignable to parameter of type 'TxKeyPath'.
```

mobile/src/app/miniapps/settings/super.tsx
```diff
@@ -8,6 +8,11 @@ import {useNavigationStore} from "@/stores/navigation"
 import {SETTINGS, useSetting} from "@/stores/settings"
 import {RouteButton} from "@/components/ui/RouteButton"
 
+const debugCoreModule = CoreModule as typeof CoreModule & {
```
PhilippeFerreiraDeSousa: Same pattern as the Nex native calls: these debug methods are used here but are not part of the typed SDK export, so the cast is local to the debug settings screen.

Original errors:
```text
src/app/miniapps/settings/super.tsx(48,67): error TS2339: Property 'dbg1' does not exist on type 'CoreModule'.
src/app/miniapps/settings/super.tsx(49,67): error TS2339: Property 'dbg2' does not exist on type 'CoreModule'.
```

PhilippeFerreiraDeSousa (commented):
Added inline notes for the non-trivial type-checking fixes. Each comment includes the original compiler error that drove the change and why I chose the narrower behavior-preserving fix.

mobile/src/components/ignite/Toggle/Switch.tsx
```diff
@@ -318,7 +318,7 @@ function _SwitchAccessibilityLabel(props: SwitchInputProps & {role: "on" | "off"
       {accessibilityMode === "icon" && shouldLabelBeVisible && (
         <Image
           style={[$switchAccessibilityIcon, {tintColor: color}]}
-          source={role === "off" ? iconRegistry.hidden : iconRegistry.view}
+          source={(iconRegistry as unknown as Record<string, number | undefined>)[role === "off" ? "hidden" : "view"]}
```
PhilippeFerreiraDeSousa: `hidden` and `view` are not registered icons. I used a cast here to preserve the existing runtime lookup, which currently resolves to `undefined`, instead of adding assets or changing the accessibility icon rendering.

Original errors:
```text
src/components/ignite/Toggle/Switch.tsx(321,49): error TS2339: Property 'hidden' does not exist on type '{ ... }'.
src/components/ignite/Toggle/Switch.tsx(321,71): error TS2339: Property 'view' does not exist on type '{ ... }'.
```

mobile/src/components/ignite/ListView.tsx
```diff
@@ -26,9 +26,11 @@ export type ListViewProps<T> = PropsWithoutRef<FlashListProps<T>>
  * @returns {JSX.Element} The rendered `ListView` component.
  */
 const ListViewComponent = forwardRef(<T,>(props: ListViewProps<T>, ref: ForwardedRef<ListViewRef<T>>) => {
```
PhilippeFerreiraDeSousa, line 28: FlashList and FlatList expose incompatible ref and viewability callback types. The component already conditionally renders one or the other, so this keeps the public prop surface FlashList-compatible and casts only at the branch where we hand props to FlatList.

Original errors included:
```text
src/components/ignite/ListView.tsx(7,30): error TS2749: 'FlashList' refers to a value, but is being used as a type here. Did you mean 'typeof FlashList'?
src/components/ignite/ListView.tsx(31,11): error TS2769: No overload matches this call.
```

mobile/src/utils/storage/storage.test.ts
```diff
@@ -1,7 +1,16 @@
 import {storage} from "@/utils/storage/storage"
+import {Result} from "typesafe-ts"
 
 const VALUE_OBJECT = {x: 1}
 
+function expectOk<T>(result: Result<T, Error>): T {
```
PhilippeFerreiraDeSousa, line 6: The storage API returns `Result<T, Error>`, so tests cannot read `.value` until they prove the result is ok. This helper keeps the test assertions readable while satisfying the discriminated union.

Original errors:
```text
src/utils/storage/storage.test.ts(23,25): error TS2339: Property 'value' does not exist on type 'Result<object, Error>'.
src/utils/storage/storage.test.ts(27,25): error TS2339: Property 'value' does not exist on type 'Result<string, Error>'.
```

mobile/patches/react-native-draggable-masonry@1.3.0.patch
```diff
@@ -24,11 +24,26 @@ index cacb84dbd769c001852547c720696ff2772cbdca..86a91158ceb7772df7eaa34f806148f4
...
-@@ -238,6 +238,13 @@ const MasonryItemComponent = ({
-             runOnJS(onDragEnd)(id);
-         });
+--- a/src/MasonryItem.tsx
++++ b/src/MasonryItem.tsx
+@@ -26,6 +26,7 @@ interface MasonryItemProps {
+     dragStartScrollY: SharedValue<number>;
+     // New props
+     sortEnabled: boolean;
++    swapMode?: boolean;
```
PhilippeFerreiraDeSousa, line 33: The dependency patch now includes the type-only fixes needed for reinstall durability. I changed the patch file because editing `node_modules` alone would make CI and future installs regress.

Original errors were from `react-native-draggable-masonry`'s source under `node_modules`; they included a missing `swapMode` prop on `MasonryItemProps` and a no-implicit-returns error in the mount effect.

Representative compiler errors:
```text
node_modules/react-native-draggable-masonry/src/MasonryItem.tsx: Property 'swapMode' does not exist on type 'MasonryItemProps'.
node_modules/react-native-draggable-masonry/src/MasonryItem.tsx: error TS7030: Not all code paths return a value.
```

mobile/src/app/mirror/fullscreen.tsx
```diff
@@ -46,7 +46,7 @@ export default function GlassesMirrorFullscreen() {
   const [defaultWearable] = useSetting(SETTINGS.default_wearable.key)
 
   const cameraRef = useRef<CameraView | null>(null)
-  const recordingTimerRef = useRef<number | null>(null)
+  const recordingTimerRef = useRef<NodeJS.Timeout | null>(null)
```
PhilippeFerreiraDeSousa: we are not in a NodeJS env, we actually use numbers

mobile/patches/react-native-draggable-masonry@1.3.0.patch
PhilippeFerreiraDeSousa: Investigate why this created an error in compiling our repo. PR from Fosse created the swapMode on some objects: https://github.com/Megane14916/react-native-draggable-masonry/pull/3

mobile/src/components/ui/InfoSection.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <View key={item.label} style={themed($row)}>
+          <Text text={item.label} style={themed($label)} />
+          <Text text={item.value == null ? "" : String(item.value)} style={themed($value)} />
+        </View>
+      ))}
+    </View>
+  )
+}
+
+const $container: ThemedStyle<ViewStyle> = ({colors, spacing}) => ({
```
fossephate: ThemedStyles should no longer be used, we should use tailwind instead

mobile/src/i18n/ar.ts
```diff
@@ -1,6 +1,6 @@
-import en, {Translations} from "./en"
+import en from "./en"
 
-const ar: Translations = {
+const ar = {
```
PhilippeFerreiraDeSousa, line 3: not sure untyping the translation data is the best. this is work in itself

mobile/src/components/ui/InfoSection.tsx
fossephate, line 1: this file should be deleted

mobile/src/components/ui/InfoSection.tsx
PhilippeFerreiraDeSousa: Delete DeviceInformation and InfoSection

mobile/src/app/miniapps/settings/super.tsx
PhilippeFerreiraDeSousa: This is in swift and Kotlin interface but missing from the typescript interface

mobile/src/app/mirror/fullscreen.tsx
fossephate: preferably no nodejs types as this is not a nodejs runtime

mobile/src/app/miniapps/settings/transcription.tsx
```diff
@@ -115,6 +115,7 @@ export default function TranscriptionSettingsScreen() {
         const backHandler = BackHandler.addEventListener("hardwareBackPress", handleBackPress)
         return () => backHandler.remove()
       }
+      return undefined
```
PhilippeFerreiraDeSousa: Are we allowed to return undefined? What's the behaviour. do we return a cleanup callback or undefined

mobile/src/app/pairing/scan-controller.tsx
```diff
@@ -1,4 +1,4 @@
-import CoreModule, {MentraDevice} from "@mentra/bluetooth-sdk"
+import CoreModule, {type Device as MentraDevice} from "@mentra/bluetooth-sdk"
```
PhilippeFerreiraDeSousa: Nope, use it as Device in this file

mobile/src/components/glasses/NexDeveloperSettings.tsx
```diff
@@ -393,7 +394,8 @@ export default function NexDeveloperSettings() {
   const onLc3AudioToggle = async (enabled: boolean) => {
     setLc3AudioEnabled(enabled)
     if (glassesConnected) {
-      await CoreModule.setLc3AudioEnabled(enabled)
+      console.log("setLc3AudioEnabled", enabled)
+      console.warn("setLc3AudioEnabled not yet implemented in Bluetooth SDK API")
```
PhilippeFerreiraDeSousa, line 398: @nic-olo Please check here, the method does not exist

mobile/src/components/glasses/NexDeveloperSettings.tsx
```diff
@@ -364,7 +364,8 @@ export default function NexDeveloperSettings() {
 
   const onSendImageClick = async () => {
     if (glassesConnected) {
-      await CoreModule.displayImage(selectedImageType, selectedImageSize)
+      console.warn("sendDisplayImage not yet implemented in Bluetooth SDK API")
+      console.log("Would display image:", selectedImageType, selectedImageSize)
```
PhilippeFerreiraDeSousa, line 368: @nic-olo Please check here, the method does not exist

mobile/src/app/pairing/prep-controller.tsx
```diff
@@ -13,6 +13,12 @@ import {useState} from "react"
 import GlassesTroubleshootingModal from "@/components/glasses/GlassesTroubleshootingModal"
 import CoreModule from "@mentra/bluetooth-sdk"
 
+const androidPermissions = PermissionsAndroid.PERMISSIONS as typeof PermissionsAndroid.PERMISSIONS & {
+  BLUETOOTH?: string
+  BLUETOOTH_ADMIN?: string
+  NEARBY_DEVICES?: string
+}
```
PhilippeFerreiraDeSousa: Ok this will have to be checked at runtime on old android phones because it does not seem that those permissions are ever not null in this code

PhilippeFerreiraDeSousa: @fossephate Do you know if there is a trick here, it seems `PermissionsAndroid.PERMISSIONS.BLUETOOTH`, `PermissionsAndroid.PERMISSIONS.BLUETOOTH_ADMIIN` and `PermissionsAndroid.PERMISSIONS.NEARBY_DEVICES` below is always undefined.

PhilippeFerreiraDeSousa: I'd like to simplify the code

mobile/src/components/ignite/Toggle/Switch.tsx
```diff
@@ -316,10 +316,7 @@ function _SwitchAccessibilityLabel(props: SwitchInputProps & {role: "on" | "off"
       )}
 
       {accessibilityMode === "icon" && shouldLabelBeVisible && (
-        <Image
-          style={[$switchAccessibilityIcon, {tintColor: color}]}
-          source={role === "off" ? iconRegistry.hidden : iconRegistry.view}
-        />
+        <Icon name={role === "off" ? "hidden" : "view"} size={14} color={color} />
```
PhilippeFerreiraDeSousa: Needs to be verified on a phone

PhilippeFerreiraDeSousa: The old code was using iconRegistry as if it contained image sources:

  <Image source={role === "off" ? iconRegistry.hidden : iconRegistry.view} />

  But iconRegistry.hidden and iconRegistry.view were not real image assets. The registry mostly
  describes icon names / icon components / glyphs, not React Native ImageSourcePropType values.
  In the old code those keys were also missing, so the source could be undefined.

  The replacement:

  <Icon name={role === "off" ? "hidden" : "view"} size={14} color={color} />

  routes rendering through the app’s actual icon component. In this PR, hidden and view are
  aliases:

  hidden -> eye-off
  view -> eye

  So the switch accessibility icon now renders through the same path as the rest of the app
  icons, instead of reading non-asset values from iconRegistry and feeding them to <Image>.

mobile/src/app/pairing/prep-controller.tsx
fossephate: could be that they existed in older versions of react-native, but not anymore
should be safe to remove / clean

mobile/src/stores/core.ts
```diff
@@ -18,6 +18,7 @@ const initialState: CoreStatus = {
   wifiScanResults: [],
   lastLog: [],
   otherBtConnected: false,
+  gallery_mode: false,
```
PhilippeFerreiraDeSousa: I think this should be initialized to true?

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/DeviceStore.kt
```diff
@@ -120,7 +120,7 @@ object GlassesStore {
         store.set("core", "dashboard_depth", 2)
         store.set("core", "head_up_angle", 30)
         store.set("core", "contextual_dashboard", true)
-        store.set("core", "gallery_mode", false)
+        store.set("core", "gallery_mode", true)
```
PhilippeFerreiraDeSousa, line 123: The native Android store is the value `MentraLive.sendGalleryMode()` reads when device-ready sync sends `save_in_gallery_mode` to the glasses. Keeping this true prevents the phone from immediately turning off the ASG client default capture behavior on first connect.

mobile/src/stores/core.ts
```diff
@@ -18,6 +18,7 @@ const initialState: CoreStatus = {
   wifiScanResults: [],
   lastLog: [],
   otherBtConnected: false,
+  gallery_mode: true,
```
PhilippeFerreiraDeSousa, line 21: Default this to true to match the ASG client contract: `save_in_gallery_mode` defaults true on the glasses so button presses still capture until the phone has an app state reason to disable capture. A false phone-side default could override that during startup before `GalleryModeSync` runs.

mobile/modules/bluetooth-sdk/ios/Source/DeviceStore.swift
```diff
@@ -80,7 +80,7 @@ class GlassesStore {
         store.set("core", "dashboard_depth", 2)
         store.set("core", "head_up_angle", 30)
         store.set("core", "contextual_dashboard", true)
-        store.set("core", "gallery_mode", false)
+        store.set("core", "gallery_mode", true)
```
PhilippeFerreiraDeSousa, line 83: Same default on iOS for parity with Android: Mentra Live reads this store value when sending `save_in_gallery_mode`, so the first native sync should preserve capture-enabled behavior unless the app state later disables it.

mobile/modules/bluetooth-sdk/ios/Source/MentraBluetoothSDK.swift
```diff
@@ -447,7 +447,7 @@ public struct BluetoothStatus: CustomStringConvertible {
     public var dashboardDepth: Int { intValue(values["dashboard_depth"]) ?? 2 }
     public var headUpAngle: Int { intValue(values["head_up_angle"]) ?? 30 }
     public var contextualDashboard: Bool { boolValue(values, "contextual_dashboard") ?? true }
-    public var galleryModeAuto: Bool { boolValue(values, "gallery_mode") ?? false }
+    public var galleryModeAuto: Bool { boolValue(values, "gallery_mode") ?? true }
```
PhilippeFerreiraDeSousa, line 450: This mirrors the Android status fallback: if `gallery_mode` is absent from the status values, expose the glasses default behavior (`true`) instead of treating missing data as capture-disabled.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothModels.kt
```diff
@@ -358,7 +358,7 @@ data class BluetoothStatus(
                 dashboardDepth = numberValue(values, "dashboard_depth") ?: 2,
                 headUpAngle = numberValue(values, "head_up_angle") ?: 30,
                 contextualDashboard = boolValue(values, "contextual_dashboard") ?: true,
-                galleryModeAuto = boolValue(values, "gallery_mode") ?: false,
+                galleryModeAuto = boolValue(values, "gallery_mode") ?: true,
```
PhilippeFerreiraDeSousa, line 361: This fallback is for status maps that do not contain `gallery_mode`. It should decode as capture-enabled for the same reason as the store default; otherwise omitted data would silently become false.

mobile/src/stores/settings.ts
```diff
@@ -474,7 +474,7 @@ export const SETTINGS: Record<string, Setting> = {
     saveOnServer: false,
     persist: false,
   },
-  gallery_mode: {key: "gallery_mode", defaultValue: () => false, writable: true, saveOnServer: true, persist: true},
+  gallery_mode: {key: "gallery_mode", defaultValue: () => true, writable: true, saveOnServer: true, persist: true},
```
PhilippeFerreiraDeSousa, line 484: This setting default needs to match the core/native default. `GalleryModeSync` treats no foreground app as capture-enabled, so a new local setting should start true rather than persisting/sending false before app state has been evaluated.

mobile/src/components/glasses/settings/BrightnessSettings.tsx
```diff
@@ -43,7 +43,6 @@ export function BrightnessSettings({
           min={0}
           max={100}
           onValueSet={onBrightnessChange}
-          containerStyle={{paddingHorizontal: 0, paddingTop: 0, paddingBottom: 0}}
```
PhilippeFerreiraDeSousa, line 46: This had no effect by mistake but fixing it would change the UI look

mobile/src/components/ignite/TextField.tsx
```diff
@@ -200,7 +200,6 @@ export const TextField = forwardRef(function TextField(props: TextFieldProps, re
 
         <TextInput
           ref={input}
-          underlineColorAndroid={colors.transparent}
```
PhilippeFerreiraDeSousa, line 203: this was undefined

mobile/src/components/ignite/Text.tsx
```diff
@@ -160,7 +160,6 @@ const $fontWeightStyles = Object.entries(typography.primary).reduce((acc, [weigh
 }, {}) as Record<Weights, TextStyle>
 
 const $baseStyle: ThemedStyle<TextStyle> = (theme) => ({
-  ...$sizeStyles.sm,
```
PhilippeFerreiraDeSousa, line 163: this was ...undefined

mobile/src/utils/LogoutUtils.ts
```diff
@@ -0,0 +1,189 @@
+import CoreModule from "@mentra/bluetooth-sdk"
+import {Session} from "@supabase/supabase-js"
+
+import restComms from "@/services/RestComms"
+import {useSettingsStore} from "@/stores/settings"
+import GlobalEventEmitter from "@/utils/GlobalEventEmitter"
+import mentraAuth from "@/utils/auth/authClient"
+import {storage} from "@/utils/storage"
+
```
PhilippeFerreiraDeSousa, line 9: Original typecheck error: `src/app/miniapps/settings/profile.tsx(16,27): error TS2307: Cannot find module '@/utils/LogoutUtils' or its corresponding type declarations.`

The earlier pass made this file re-export `LogoutUtils` from `AuthContext`. I moved the existing class implementation here instead because this is the import path the profile settings code already uses, and it avoids making the auth context module the public utility module. The logout method bodies are unchanged.

PhilippeFerreiraDeSousa (commented):
Typecheck investigation note for the latest cleanup.

PhilippeFerreiraDeSousa:
Icon prop note, moved out of the repo artifact:

The original typecheck issue here was that some legacy Ignite call sites pass `icon` to `<Icon />`, while this app's `Icon` implementation renders from `name`. I intentionally kept `icon?: IconTypes` in the prop type but did not read it in `Icon`, so the PR does not change the current UI. In other words, places that were visually relying on `name` still behave the same, and places that passed `icon` but were not rendered before remain unchanged.

The live affected call sites I found are `ListItem`, `SelectSetting`, and `SelectWithSearchSetting`. `Header` is not affected because it already maps its action icon value into `name` before rendering `PressableIcon`. If we want to make those legacy `icon` call sites render visibly later, that should be a deliberate UI change: migrate them to `name` and choose the concrete icon names, for example replacing the old `caretRight` value with the current registry's `chevron-right` if that is the desired visual.

fossephate (approved):

PR #2806, opened 2026-05-15
Detect captive portal WiFi on Mentra Live and warn in mobile app

nic-olo:
## Summary

Users often connect Mentra Live to captive-portal WiFi (hotels, flights, cafés). The glasses cannot complete a browser sign-in flow, which leads to broken cloud/streaming/OTA behavior and support noise.

## Changes

- **asg_client**: On Android 10+, detect `NET_CAPABILITY_CAPTIVE_PORTAL` on the active Wi‑Fi network via `BaseNetworkManager.isWifiCaptivePortal()`. Include `captive_portal` (boolean) on existing `wifi_status` BLE JSON alongside `connected`, `ssid`, and `local_ip`.
- **bluetooth-sdk (Android/iOS/TS)**: Extend `WifiStatus` with optional `captivePortal`, plumb through `wifi_status` / `wifi_status_change`, `GlassesStore` (`wifiCaptivePortal`), and legacy `wifiFromLegacyFields` in the mobile glasses store.
- **mobile app**: New `CaptivePortalWifiBanner` shown when `wifi.state === "connected"` and `wifi.captivePortal === true` on WiFi scan, WiFi connection success, and device info screens. English copy under `wifi:captivePortalTitle` / `wifi:captivePortalMessage`.

## Testing

- Added a unit test for legacy `wifiCaptivePortal` mapping in `glasses-ota.test.ts`.
- Local `asg_client` Gradle compile was not run here (no Android SDK in the agent environment). Mobile `tsc`/`jest` were not run for the same reason; CI should cover them.

## Notes

- Detection relies on the Android connectivity stack reporting the captive-portal capability; networks that behave like portals without that flag may not be flagged.
- Older Android (&lt; 10) paths report `captive_portal: false` only.

<div><a href="https://cursor.com/agents/bc-2aab38ad-f389-4ae6-afb5-0f0195297fef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2aab38ad-f389-4ae6-afb5-0f0195297fef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


asg_client/app/src/main/java/com/mentra/asg_client/io/network/core/BaseNetworkManager.java
```diff
@@ -240,6 +241,36 @@ public boolean isConnectedToWifi() {
...
+        }
+
+        if (!capabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI)) {
+            return false;
+        }
+
+        boolean captive = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_CAPTIVE_PORTAL);
+        Log.d(TAG, "WiFi captive portal capability: " + captive);
+        return captive;
+    }
```
rampageservices, line 280: **Detection-reliability question** — this implementation depends entirely on the OS-level `NET_CAPABILITY_CAPTIVE_PORTAL` flag, which is only set when Android's connectivity stack actually performs the captive-portal probe (`Settings.Global.captive_portal_https_url` / `captive_portal_http_url`) **and** succeeds in identifying the network as captive.

On many stripped-down AOSP builds used for low-power AR/wearable devices, the probe is disabled (`Settings.Global.CAPTIVE_PORTAL_MODE = 0`) or pointed at an unreachable host, and the capability flag is never set — even on networks that are clearly captive.

Before merge, please confirm on real Mentra Live hardware that this flag actually fires against a known captive portal (phone hotspot with redirect, or `iptables`-based test rig). If it doesn't, the user-facing banner never triggers and the PR delivers no value in the field.

If reliability turns out to be a problem, the lowest-effort fallback is a single HTTPS HEAD against a known endpoint (your own `/healthz` works) on WiFi-connect, with `captive_portal` set when the response signature is wrong (HTTP redirect to a non-HTTPS host, TLS failure where TLS should succeed, captive-portal-style HTML body, etc.).

asg_client/app/src/main/java/com/mentra/asg_client/io/network/core/BaseNetworkManager.java
```diff
@@ -240,6 +241,36 @@ public boolean isConnectedToWifi() {
...
+        if (capabilities == null) {
+            return false;
+        }
+
+        if (!capabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI)) {
+            return false;
+        }
+
+        boolean captive = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_CAPTIVE_PORTAL);
+        Log.d(TAG, "WiFi captive portal capability: " + captive);
```
rampageservices, line 278: nit: this `Log.d` fires every time `sendWifiStatusOverBle` is called. Per my read that's event-driven (WiFi state-change), so spam is bounded — but worth confirming there's no periodic poller calling this. Consider downgrading to `Log.v` or gating behind a debug flag.

asg_client/app/src/main/java/com/mentra/asg_client/service/communication/managers/CommunicationManager.java
```diff
@@ -77,16 +77,20 @@ public void sendWifiStatusOverBle(boolean isConnected) {
...
+                    Log.d(TAG, "🔄 🧾 Captive portal: " + captivePortal);
                     
                     wifiStatus.put("ssid", ssid != null ? ssid : "unknown");
                     wifiStatus.put("local_ip", localIp != null ? localIp : "");
+                    wifiStatus.put("captive_portal", captivePortal);
                 } else {
                     Log.d(TAG, "🔄 ❌ WiFi not connected or network manager unavailable");
                     wifiStatus.put("ssid", "");
                     wifiStatus.put("local_ip", "");
+                    wifiStatus.put("captive_portal", false);
```
rampageservices, line 88: nit: emitting `captive_portal: false` in the disconnected branch is harmless (and consistent with how `ssid`/`local_ip` are emitted as empty), just semantically redundant since disconnected ≡ no portal. Take or leave.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/Bridge.kt
```diff
@@ -371,8 +371,8 @@ public class Bridge private constructor() {
 
         /** Send WiFi status change */
         @JvmStatic
-        fun sendWifiStatusChange(connected: Boolean, ssid: String?, localIp: String?) {
-            val status = WifiStatus.fromStoreFields(connected, ssid, localIp) ?: return
+        fun sendWifiStatusChange(connected: Boolean, ssid: String?, localIp: String?, captivePortal: Boolean = false) {
```
rampageservices, line 508: Defaulting `captivePortal: Boolean = false` here lets future call sites compile silently without making a deliberate decision about portal state. There aren't other call sites in this diff, but if you anticipate more entry points, dropping the default forces an explicit choice at each one. Same applies to the iOS bridge.

mobile/modules/bluetooth-sdk/ios/Source/Bridge.swift
```diff
@@ -242,11 +242,12 @@ class Bridge {
         Bridge.sendTypedMessage("glasses_serial_number", body: body)
     }
 
-    static func sendWifiStatusChange(connected: Bool, ssid: String?, localIp: String?) {
+    static func sendWifiStatusChange(connected: Bool, ssid: String?, localIp: String?, captivePortal: Bool = false) {
```
rampageservices, line 388: Same nit as Bridge.kt: defaulting `captivePortal: Bool = false` here lets future call sites compile silently without making a deliberate decision. Consider dropping the default to force an explicit choice at each site.

mobile/modules/bluetooth-sdk/src/BluetoothSdk.types.ts
```diff
@@ -57,7 +57,7 @@ export type LogEvent = {
 
 export type WifiStatus =
   | {state: "disconnected"}
-  | {state: "connected"; ssid: string; localIp?: string}
+  | {state: "connected"; ssid: string; localIp?: string; captivePortal?: boolean}
```
rampageservices, line 111: Minor inconsistency: `captivePortal` is optional on the TS `connected` variant, but every other platform (Kotlin `Connected(... captivePortal: Boolean = false)`, Swift `case connected(..., captivePortal: Bool)`) defaults it to `false`. Consumers here have to write `wifi.captivePortal === true` to handle `undefined`. Consider making it required (with construction sites defaulting to `false`) for parity.

mobile/src/stores/glasses.ts
```diff
@@ -196,7 +208,9 @@ export const useGlassesStore = create<GlassesState>()(
         if (connected && !trimmedSsid) {
           return {}
         }
-        const wifi: WifiStatus = connected ? {state: "connected", ssid: trimmedSsid} : {state: "disconnected"}
+        const wifi: WifiStatus = connected
+          ? {state: "connected", ssid: trimmedSsid, captivePortal: false}
```
rampageservices, line 234: Since the TS type marks `captivePortal` optional, the literal `captivePortal: false` here is equivalent to omitting it. If you address the type-optionality comment on `BluetoothSdk.types.ts` by making it required, keep this; otherwise drop for less noise.

mobile/src/i18n/en.ts
```diff
@@ -294,6 +294,9 @@ const en = {
     failedDescription: "Make sure the password was entered correctly",
     connectingDescription: "This may take up to 20 seconds",
     connectingToNetwork: "Connecting to {{network}}...",
+    captivePortalTitle: "Sign-in WiFi detected",
+    captivePortalMessage:
+      "This network needs a browser sign-in (common on flights, hotels, and cafés). Mentra Live cannot open that page, so many features will not work. Use your home WiFi or phone hotspot instead.",
```
rampageservices, line 299: These new strings only exist in `en.ts`. The other locale files — `ar.ts`, `es.ts`, `fr.ts`, `hi.ts`, `ja.ts`, `ko.ts` — will display the raw key (e.g. `wifi:captivePortalTitle`) instead of falling back to English unless the i18n config does fallback explicitly. For a string this user-facing (it explains why glasses don't work), key-strings are worse than the English copy. Either add translations now, or confirm the i18n layer falls back to English for missing keys and file a follow-up i18n ticket. Which is the norm in this repo?

mobile/src/components/wifi/CaptivePortalWifiBanner.tsx
```diff
@@ -0,0 +1,29 @@
...
+import {Text} from "@/components/ignite"
+import {useAppTheme} from "@/contexts/ThemeContext"
+import {useGlassesStore} from "@/stores/glasses"
+
+/**
+ * Shown when glasses report a captive-portal WiFi network (Android NET_CAPABILITY_CAPTIVE_PORTAL).
+ */
+export function CaptivePortalWifiBanner() {
+  const {theme} = useAppTheme()
+  const show = useGlassesStore((s) => s.wifi.state === "connected" && s.wifi.captivePortal === true)
```
rampageservices, line 12: The store mapping is covered by `glasses-ota.test.ts`, but the banner component itself isn't. A 3-case test would be cheap insurance:

1. no banner when `wifi.state === "disconnected"`
2. no banner when connected & `!captivePortal`
3. banner rendered when connected & `captivePortal === true`

The project already uses `react-native-testing-library` (see `mobile/src/components/ignite/Text.test.tsx` for a pattern).

mobile/src/app/wifi/scan.tsx
```diff
@@ -275,6 +276,9 @@ export default function WifiScanScreen() {
       )}
 
       <View className="flex-1">
+        <View className="px-5 mb-2">
+          <CaptivePortalWifiBanner />
```
rampageservices, line 235: Sanity-check: the banner shows when *currently connected* to a captive portal. On the scan screen, that means the user is back here because the captive WiFi-A wasn't working, with WiFi-A still reported as the connected state until they pick WiFi-B. I think that's actually the right behavior — it explains *why* they're back at scan — but worth confirming the state clears immediately on true disconnect. `MentraLive.swift` and the Android equivalent appear to reset `wifiCaptivePortal` to `false` on disconnect, so this should be fine. Just flagging to double-check the no-stale-state path.

rampageservices (commented):
## Review — captive-portal WiFi detection

Thanks for tackling this. The plumbing is clean, additive, version-gated correctly at API 29, and consistent across all three SDK targets (asg_client Java, bluetooth-sdk Kotlin/Swift/TS). I left inline comments on specifics; two higher-level things below.

### [BLOCKER] Detection reliability on Mentra Live firmware

`NET_CAPABILITY_CAPTIVE_PORTAL` is only set when Android's connectivity stack actually runs its portal probe (`captive_portal_https_url` / `captive_portal_http_url` in `Settings.Global`) **and** that probe successfully identifies the network as captive. On stripped-down AOSP builds typical for low-power AR devices, the probe is sometimes disabled outright (`Settings.Global.CAPTIVE_PORTAL_MODE = 0`) or pointed at an unreachable host, in which case this capability flag is never set — even on networks that are obviously captive.

Can you confirm — on real Mentra Live hardware against a known captive portal (phone hotspot with redirect, or an `iptables`-based test rig) — that `NET_CAPABILITY_CAPTIVE_PORTAL` actually fires? A 5-minute manual test settles whether this PR delivers value at all in the field.

If detection turns out unreliable, the lowest-effort fallback is a single HTTPS HEAD against your own endpoint on WiFi-connect, with `captive_portal` set when the response signature is wrong (HTTP redirect to a non-HTTPS host, TLS where TLS should succeed, etc.). Happy to file a follow-up issue if needed. See inline comment on `BaseNetworkManager.isWifiCaptivePortal()`.

### [BLOCKER] No end-to-end validation performed

The PR body notes: *"Local asg_client Gradle compile was not run here … Mobile tsc/jest were not run for the same reason; CI should cover them."*

CI catches type/compile errors but not the runtime question above. Could you post a screenshot or short video of the banner triggering on real Mentra Live + real captive portal before this merges? This is a user-facing warning — if the detection path is wrong we'll find out from a support ticket, not from CI.

### [NIT] Other observations (see inline)

- Missing translations in 6 locales (`ar`, `es`, `fr`, `hi`, `ja`, `ko`)
- No component test for `CaptivePortalWifiBanner`
- TS `captivePortal` is optional but every other platform defaults it — minor inconsistency
- Minor log-noise / redundant-default nits

### Note on PR provenance

The PR body indicates this was generated by a Cursor background agent — fine, but worth a human-author pass to confirm the testing claims and that the description matches what was actually exercised. Per the repo's `AGENTS.md`, commits should not carry AI co-author attribution; same spirit probably applies to the description prose where convenient.

Overall: leaving as "Comment" rather than "Request changes" since the blockers are about evidence/validation rather than code defects. Once the e2e test confirms the OS capability flag fires in practice, this is good to merge with the smaller items addressed.

aisraelov:
@codex review

aisraelov (approved):
Good to merge!

PR #2844, opened 2026-05-19
Reject concurrent SDK photo requests for full capture-to-upload pipeline

nic-olo:
## Summary

- Collapse `isCapturingPhoto` and `isUploadingPhoto` into a single `isPhotoJobInFlight` flag on the ASG client so only one SDK photo job (capture **or** upload/BLE handoff) runs at a time.
- Second `take_photo` commands during an in-flight job now return immediate `CAMERA_BUSY` instead of queuing in CameraNeo until the SDK 30s timeout (OS-947).
- Extend the photo-job safety timeout from 15s to 45s so slow webhook uploads do not prematurely release the flag.

## Test plan

- [ ] Rapid-fire 3x `camera.requestPhoto()` from a cloud SDK app — expect first succeeds, others get `CAMERA_BUSY` in under 1s
- [ ] Single photo request over WiFi webhook — still completes end-to-end
- [ ] Photo with BLE transfer path — still completes end-to-end
- [ ] Webhook fail → BLE fallback — job stays single-flight through handoff
- [ ] Physical button / gallery capture — unaffected (non-SDK path)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/media/core/MediaCaptureService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/media/core/MediaCaptureService.java
@@ -45,6 +45,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


aisraelov:
@codex review

PR #2876, opened 2026-05-25
Mentra console mcp server

nic-olo:


nic-olo:
@codex review

isaiahb:
Is it really necessary to expose generating new api keys via the cli / mcp? since the point of generating the api keys is to authenticate the cli/mcp servers this is an action it might make sense to limit to the user to do via the dashboard ui only,  but I'm curious if there is another reason for it that i'm missing?

isaiahb:
Also please fix the tests & build errors
Screenshot 2026-05-26 at 10 06 56 AM

aisraelov:
Now is not a good time to build this. Console will be almost completely rewritten for new SDK, and we have a new CLI coming for that new SDK. The best play for us is to finish the local JS SDK (including accompanying CLI for building new local JS miniapps), then we can upgrade the new CLI with new developer console specific things.

nic-olo:
> Now is not a good time to build this. Console will be almost completely rewritten for new SDK, and we have a new CLI coming for that new SDK. The best play for us is to finish the local JS SDK (including accompanying CLI for building new local JS miniapps), then we can upgrade the new CLI with new developer console specific things.

It's already done and I mostly need it to query user incidents.

PR #2877, opened 2026-05-25
Enhance GallerySyncService with desync recovery logic and improve tes…

nic-olo:
- Introduced a new function to check for empty sync responses from the server.
- Updated the GallerySyncService to retry syncing when a desync is detected, specifically when the glasses report content but the API returns no data.
- Enhanced unit tests to cover various scenarios for the new desync recovery logic, ensuring robust handling of sync states and responses.
- Added additional mock implementations for localStorageService and asgCameraApi to support new test cases.


aisraelov:
Hey so this seems fine as a hotfix but can you please try to find the root cause for why this happened? Surely there is *some* reason why the BLE gallery status check would return 8 items, but the HTTP request returns 0. There's something wrong with the HTTP request.

aisraelov (approved):
Approving this for now as a hotfix, however, we still need to fix the root cause of the issue.

aisraelov:
cherry picked to new branch

PR #2989, opened 2026-05-27
Ya/mentra display vad toggle

yagarwal1307:


mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/utils/NexSGCUtils.kt
```diff
@@ -202,15 +200,9 @@ object NexProtobufUtils {
...
-            .build()
-        val phoneToGlasses = PhoneToGlasses.newBuilder()
-            .setMsgId(msgId)
-            .setVersionRequest(versionRequest)
-            .build()
-        return generateProtobufCommandBytes(phoneToGlasses)
+        // VersionRequest/VersionResponse removed from the BLE schema; fw_version now comes via DeviceInfo.
+        Bridge.log("Nex: generateVersionRequestCommandBytes is a no-op after schema removal")
+        return ByteArray(0)
     }
```
PhilippeFerreiraDeSousa, line 206: So why are we keeping this function and
```
                // Query glasses protobuf version from firmware
                Bridge.log("=== SENDING GLASSES PROTOBUF VERSION REQUEST ===")
                val versionQueryPacket = NexProtobufUtils.generateVersionRequestCommandBytes()
                sendDataSequentially(versionQueryPacket, 100)
                Bridge.log("Sent glasses protobuf version request")
```

mobile/src/components/glasses/NexDeveloperSettings.tsx
```diff
@@ -279,6 +279,9 @@ export default function NexDeveloperSettings() {
   // LC3 Audio Control state
   const [lc3AudioEnabled, setLc3AudioEnabled] = useState(true)
 
+  // VAD (Voice Activity Detection) — sourced from native DeviceStore via glasses status.
+  const vadEnabled = useGlassesStore((state) => state.voiceActivityDetectionEnabled)
+
```
PhilippeFerreiraDeSousa, line 284: So what is this file about, because when I fixed the MentraOS type checking errors last week, this file was calling 2 methods that did not exist, I replaced the calls by warning logs https://github.com/Mentra-Community/MentraOS/pull/2784/changes#r3244903493

mobile/src/services/MantleManager.ts
```diff
@@ -170,17 +170,12 @@ class MantleManager {
       sendDisplayEvent: (event) => BluetoothSdk.displayEvent(event),
       subscribeGlassesStatus: (onChange) => BluetoothSdk.onGlassesStatus(onChange),
       restartTranscriber: () => BluetoothSdk.restartTranscriber(),
-      setMicRequirements: (requirements) => {
-        const values: Record<string, unknown> = {
+      setMicRequirements: (requirements) =>
+        BluetoothSdk.update("core", {
```
PhilippeFerreiraDeSousa, line 174: The key "core" was renamed to "bluetooth" even though we support it as an alias for now
```
    companion object {
        const val BLUETOOTH_CATEGORY = "bluetooth"
        private const val LEGACY_CORE_CATEGORY = "core"

        fun normalizeCategory(category: String): String =
                if (category == LEGACY_CORE_CATEGORY) BLUETOOTH_CATEGORY else category
```

mobile/src/services/MantleManager.ts
```diff
@@ -170,17 +170,12 @@ class MantleManager {
...
           should_send_pcm: requirements.shouldSendPcm,
           should_send_lc3: requirements.shouldSendLc3,
           should_send_transcript: requirements.shouldSendTranscript,
-        }
-        if (requirements.voiceActivityDetectionEnabled !== undefined) {
-          values.voice_activity_detection_enabled = requirements.voiceActivityDetectionEnabled
```
PhilippeFerreiraDeSousa, line 180: I had rewired the cloud bypassVad to control the glasses vad instead of the defunct phone-side vad. But there wasn't a good reason to and I didn't know if this flow was triggered really so this was not tested anyway, and I think was already dead code.

PhilippeFerreiraDeSousa (approved):

mobile/src/services/MantleManager.ts
yagarwal1307: Its not triggered anymore. Earlier we used to have a toggle in the ui for bypass vad. And the logic used to flow from glasses to the cloud back to glasses again. (Earlier we had a lot of sphagetti code.)
If we wanna have that button again there are simpler and better ways to implement that.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/utils/NexSGCUtils.kt
yagarwal1307: All the confusion here has been caused because until recently we had two different protobuf files and one supported version request and another didn't. In the recent efforts to reconcile those files we lost the ability for the version request. Will add that again soon. So keeping this function.

mobile/src/components/glasses/NexDeveloperSettings.tsx
yagarwal1307: This is for mentra displays.
There are a few test functions which can interact with the mentra displays directly. Kinda like dev interface for mentra displays.

mobile/src/services/MantleManager.ts
yagarwal1307: Sure Ill update that.

PR #3051, opened 2026-06-01
Harden Mentra Live BLE transport for large photo commands

PhilippeFerreiraDeSousa:
## Scope

- Adds reverse BLE chunk reassembly for large JSON messages sent from Mentra Live glasses back to the phone SDK on Android and iOS.
- Chunks large phone-to-glasses JSON commands through the K900/BES serial path so oversized photo commands can survive BLE transport limits.
- Hardens BLE photo transfer failure reporting from glasses to phone, including packet packing failures, timeouts, repeated BLE TX failures, and exhausted retries.
- Sends iOS BLE MTU readiness configuration and uses BLE transport state for the button capture gate.

## Why split

This branch isolates BLE transport robustness for large photo commands from the larger aggregate branch. It intentionally excludes the aggregate merge commit `94011642f` and keeps BES tracing and other neighboring work out of this draft PR so review can focus on transport behavior and photo-transfer failure handling.

## Commit list

- `16c7982d9` Add reverse BLE chunk reassembly
- `a8c6ef5ac` Harden BLE photo transfer failures
- `95e6865a9` Send iOS BLE MTU config during readiness
- `ce4448b0b` Use BLE transport for button capture gate
- `5c13449eb` Remove split-only BLE trace dependency

No requested cherry-pick was empty.

## Conflict resolution notes

- Resolved `mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.java` by preserving chunked-message detection/reassembly before normal JSON message handling.
- Removed `BleTraceLogger` calls that came from the aggregate branch context but belong to the separate BLE/BES tracing split, keeping this PR self-contained against `dev`.

## Validation

- Passed: `git diff --check origin/dev..HEAD`
- Ran: `cd asg_client && ./gradlew :app:compileDebugJavaWithJavac`
  - The in-scope `BleTraceLogger` compile error was removed.
  - The task still fails on untouched existing streaming-service `VideoConfig` constructor mismatches in:
    - `asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java`
    - `asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/SrtStreamingService.java`
- Confirmed there are no remaining `BleTraceLogger` references in `asg_client` or the Android Bluetooth SDK paths included in this split.

## Dependencies/ordering notes

- This branch is based on `dev`.
- It does not depend on the aggregate merge commit `94011642f`.
- It should not require the separate BLE/BES tracing split because the trace logger references were removed from this PR.

---

> [!NOTE]
> **Medium Risk**
> Changes core BLE framing, chunk protocols, and photo transfer failure paths on both glasses and mobile SDKs; regressions could break pairing, large commands, or photo delivery without obvious UI errors.
> 
> **Overview**
> This PR hardens the Mentra Live BLE path for **large JSON (especially photo commands)** and **photo file transfers** between phone and glasses.
> 
> **Glasses (`asg_client`)** now **chunks outbound JSON** that exceeds the K900/BES packed limit (`MessageChunker` + paced sends in `K900BluetoothManager`), with a shared **C-wrapper helper** and a **safer default file packet size (~221 bytes)** aligned to BLE MTU instead of assuming 400-byte payloads. Incoming chunk reassembly is **stricter** (invalid metadata dropped, `totalChunks` mismatches reset sessions, thread-safe chunk maps). Failed BLE photo transfers **notify the phone** via `transfer_failed` with reasons (pack failure, packet timeout, stuck TX, max retries). Any phone command (not only handshake) **refreshes connection heartbeat**.
> 
> **Phone Bluetooth SDK (Android + iOS)** gains **reverse chunk reassembly** for glasses→phone JSON, **measured chunk sizing** so escaped JSON still fits under the 253-byte BES limit, and handling of **chunk envelopes inside C-wrapped** traffic. iOS sends **`set_ble_mtu`** after `glasses_ready`. Photo flows emit **`photo_response` errors** on transfer timeout/failure and **preserve `requestId` across BLE photo retries** when the glasses ask to resend.
> 
> README adds **phone↔glasses compatibility expectations** (new phone must tolerate old `asg_client`; new client need not support old phone apps beyond cloud min-version).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f3b69a3a9d45c7cf56f00525fd9166b14c64d4b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.java
```diff
@@ -6186,9 +6224,10 @@ private void processFilePacket(K900ProtocolUtils.FilePacketInfo packetInfo) {
                     Log.e(TAG, "❌ BLE photo transfer incomplete after final packet. Missing " + missingPackets.size() + " packets: " + missingPackets);
                     Log.e(TAG, "❌ Telling glasses to retry entire transfer");
 
-                    // Tell glasses transfer failed, they will retry
+                    // Tell glasses transfer failed, they will retry. Keep the photo transfer
+                    // entry so the retry still maps back to the original requestId.
                     sendTransferCompleteConfirmation(packetInfo.fileName, false);
-                    blePhotoTransfers.remove(bleImgId);
+                    photoTransfer.session = null;
```
PhilippeFerreiraDeSousa, line 6247: From Codex:
> Why: transfer_complete(success=false) causes the glasses-side K900BluetoothManager.handlePhoneConfirmation(..., false) to retry the same current file transfer from packet 0. It keeps currentFileTransfer.fileName and restarts with currentPacketIndex = 0, so the phone will receive the retry under the same bleImgId / file base name.
> 
> On the phone SDK side, blePhotoTransfers is the only thing mapping that bleImgId back to the original requestId. If we remove it on incomplete final packet, the retry arrives as an “unknown transfer” and the SDK loses the request context. Keeping the entry but clearing photoTransfer.session is the right shape: it preserves request metadata, while forcing the retried packet 0 to create a fresh FileTransferSession.

mobile/modules/bluetooth-sdk/ios/Source/sgcs/MentraLive.swift
```diff
@@ -768,9 +768,12 @@ extension MentraLive: CBPeripheralDelegate {
...
 
-            // Request MTU size
-            let mtuSize = peripheral.maximumWriteValueLength(for: .withResponse)
-            Bridge.log("LIVE: Current MTU size: \(mtuSize + 3) bytes")
+            let withResponseMtu = peripheral.maximumWriteValueLength(for: .withResponse) + 3
+            let withoutResponseMtu = peripheral.maximumWriteValueLength(for: .withoutResponse) + 3
+            currentMtu = max(23, min(withResponseMtu, withoutResponseMtu))
+            Bridge.log(
+                "LIVE: Current MTU estimate: withResponse=\(withResponseMtu), withoutResponse=\(withoutResponseMtu), selected=\(currentMtu)"
+            )
```
PhilippeFerreiraDeSousa: Codex:
> MTU is the BLE “maximum transmission unit”: basically the largest BLE packet size the link can carry at that layer. In practice, app payload is a bit smaller than MTU because BLE uses a few bytes for protocol overhead. That is why you see + 3 when estimating MTU from CoreBluetooth’s maximumWriteValueLength: iOS gives the max app payload, and the code converts it back to MTU-ish size.
> 
> Why send MTU config now: large photo transfers are packetized by the glasses. If the glasses assume packets are too large for the phone/BLE link, the phone can miss or fail to reassemble chunks, especially for big photo payloads. Sending the phone’s selected MTU lets the glasses choose a safe file packet size.

mobile/modules/bluetooth-sdk/ios/Source/sgcs/MentraLive.swift
```diff
@@ -768,9 +768,12 @@ extension MentraLive: CBPeripheralDelegate {
             // Keep state as connecting until glasses are ready
             updateConnectionState(ConnTypes.CONNECTING)
 
-            // Request MTU size
-            let mtuSize = peripheral.maximumWriteValueLength(for: .withResponse)
-            Bridge.log("LIVE: Current MTU size: \(mtuSize + 3) bytes")
+            let withResponseMtu = peripheral.maximumWriteValueLength(for: .withResponse) + 3
+            let withoutResponseMtu = peripheral.maximumWriteValueLength(for: .withoutResponse) + 3
+            currentMtu = max(23, min(withResponseMtu, withoutResponseMtu))
```
PhilippeFerreiraDeSousa: Codex:
> 23 is the BLE default ATT MTU. It is the baseline every BLE connection can assume unless MTU negotiation raises it.

PhilippeFerreiraDeSousa:
@codex review

mobile/modules/bluetooth-sdk/ios/Source/sgcs/MentraLive.swift
```diff
@@ -3235,6 +3274,20 @@ class MentraLive: NSObject, SGCManager {
         )
     }
 
+    private func sendBleMtuConfig() {
+        let effectiveMtu = min(currentMtu, bes2700MtuLimit)
+        let json: [String: Any] = [
+            "type": "set_ble_mtu",
```
PhilippeFerreiraDeSousa, line 3323: So this was only sent on the android side until now.

asg_client/app/src/main/java/com/mentra/asg_client/utils/smartglasses/MessageChunker.java
```diff
@@ -43,15 +43,14 @@ public static List<JSONObject> createChunks(String originalJson, long messageId)
...
+        int totalChunks = chunkDataList.size();
 
         Log.d(TAG, "Creating " + totalChunks + " chunks for message of size " + totalBytes + " bytes");
 
         for (int i = 0; i < totalChunks; i++) {
-            int startIndex = i * CHUNK_DATA_SIZE;
-            int endIndex = Math.min(startIndex + CHUNK_DATA_SIZE, totalBytes);
-            int chunkLength = endIndex - startIndex;
-            String chunkData = new String(messageBytes, startIndex, chunkLength, StandardCharsets.UTF_8);
+            String chunkData = chunkDataList.get(i);
```
PhilippeFerreiraDeSousa, line 68: The math of this commit was checked like this. Invariant: randomized JSON payloads with ASCII, accents, CJK, and emoji should reassemble exactly, no chunk should exceed 80 UTF-8 bytes, and no replacement characters should appear.
```
$ node - <<'NODE'
const size = 80;
function isCont(b){ return (b & 0xC0) === 0x80; }
function end(bytes,start){ let e=Math.min(start+size, bytes.length); while(e>start && e<bytes.length && isCont(bytes[e])) e--; return e>start ? e : Math.min(start+size, bytes.length); }
function splitUtf8(s){ const bytes=Buffer.from(s,'utf8'); const out=[]; let offset=0; while(offset<bytes.length){ const e=end(bytes,offset); out.push(bytes.subarray(offset,e).toString('utf8')); offset=e; } return out; }
const alphabet = ['a','Z','0',' ','é','ñ','漢','界','🙂','🚀','👨‍💻'];
for (let t=0; t<10000; t++) {
  let text = '"''"';
  const len = Math.floor(Math.random()*500);
  for (let i=0; i<len; i++) text += alphabet[Math.floor(Math.random()*alphabet.length)];
  const payload = JSON.stringify({type:'status', message:text});
  const parts = splitUtf8(payload);
  const joined = parts.join('"''"');
  if (joined !== payload) throw new Error(`mismatch ${t}`);
  for (const part of parts) {
    const bytes = Buffer.byteLength(part, 'utf8');
    if (bytes > size) throw new Error(`oversize ${bytes} ${t}`);
    if (part.includes('\uFFFD')) throw new Error(`replacement ${t}`);
  }
}
console.log('utf8 chunk property check passed');
NODE
```

PR #3088, opened 2026-06-04
[codex] Preserve relative gallery metadata while saving files

PhilippeFerreiraDeSousa:
## Summary

Separates a narrow gallery metadata preservation fix out of the release hotfix PR.

This PR prevents `saveDownloadedFile()` from rewriting existing relative gallery metadata paths as absolute iOS Documents paths when saving a newly downloaded file.

## Scope Boundary

This is intentionally **not** the multi-photo sharing hotfix. The release hotfix remains PR #3087 and should contain only the share-cache filename fix.

This PR also does **not** attempt to repair or migrate already-saved absolute paths. The earlier recovery/backport behavior was removed because we do not have enough evidence that released app state is already corrupted.

## Root Cause

`saveDownloadedFile()` loaded existing metadata through `getDownloadedFiles()`. That method is intended for runtime callers and reconstructs relative stored paths into absolute current Documents paths. Saving that hydrated map back can turn previously relative metadata into absolute paths.

## Changes

- Add a raw metadata load helper for `saveDownloadedFile()`.
- Use raw stored metadata when appending/replacing a downloaded file entry.
- Preserve the existing `getDownloadedFiles()` runtime path reconstruction behavior.
- Add a focused test that existing relative metadata stays relative after saving a new file.

## Validation

- `git diff --check origin/dev..HEAD`
- `cd mobile && bun run test -- src/services/asg/localStorageService.test.ts --runInBand`
- `cd mobile && bun run compile`

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 8c3405c54c3ffb17b0f4a8361fe73913d3603913. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/src/services/asg/localStorageService.test.ts
+++ b/mobile/src/services/asg/localStorageService.test.ts
@@ -1,0 +1,58 @@
+import {storage} from "@/utils/storage"
+
+import {localStorageService} from "./localStorageService"
+
+const DOWNLOADED_FILES_KEY = "asg_downloaded_files"
+const DOCS = "/tmp/documents"
+
+function loadDownloadedFilesMetadata(): Record<string, {filePath: string; thumbnailPath?: string}> {
+  const result = storage.load<Record<string, {filePath: string; thumbnailPath?: string}>>(DOWNLOADED_FILES_KEY)
+  if (result.is_error()) {
+    throw result.error
+  }
+  return result.value
+}
+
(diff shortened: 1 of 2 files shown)


PhilippeFerreiraDeSousa:
This is for hotfix release 2.11

PR #3093, opened 2026-06-04
[codex] Align video recording runtime types

PhilippeFerreiraDeSousa:
## Summary

This keeps the local-miniapp video-recording path on a single shared runtime contract instead of duplicating inline option/result shapes between the island runtime and the Manager host.

## Root Cause

The video-recording adapter shape was duplicated at the island runtime boundary and in the phone coordinator. That makes the `LocalMiniappRuntime` payload, `MantleManager` adapter wiring, and `PhoneVideoCoordinator` implementation easy to drift when adjacent work touches generated island types or Bluetooth settings. The source on current `origin/dev` now compiles, but this removes the contract mismatch vector observed while validating the VAD-default PR.

## Changes

- Adds named `VideoRecordingOptions` and `VideoRecordingStarted` types to the island runtime config.
- Normalizes the video-recording payload in `LocalMiniappRuntime` into that shared options type before calling the host adapter.
- Derives `PhoneVideoCoordinator` option/result types from `RuntimeHooks["videoRecording"]`.
- Annotates the `MantleManager` video adapter with the shared runtime-hook type.

## Validation

- `cd mobile/modules/island && bun run build`
- `cd mobile && bun compile`
- `git diff --check`

## VAD PR Separation

This PR is intentionally separate from the VAD-default branch. It only aligns the existing video-recording type contract on top of `dev` and does not include VAD setting behavior changes.

### DIFF
--- a/mobile/modules/island/src/runtime/config.ts
+++ b/mobile/modules/island/src/runtime/config.ts
@@ -321,17 +321,20 @@ export interface RuntimeHooks {
  * the bluetooth-sdk startVideoRecording/stopVideoRecording). Unlike photo, this
  * is fire-and-forget start/stop — no uploaded URL is returned.
  */
+export interface VideoRecordingOptions {
+  width?: number
+  height?: number
+  fps?: number
+  sound?: boolean
+  save?: boolean
+}
+
+export interface VideoRecordingStarted {
+  recordingId: string
+}
+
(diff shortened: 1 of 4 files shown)


PhilippeFerreiraDeSousa:
Not convinced this is very useful. I had another agent report a type checking error so this was a fix pass but type checking does pass

PR #3100, opened 2026-06-06
Make ASG photo capture single-flight

PhilippeFerreiraDeSousa:
## Summary
- make ASG photo captures single-flight across native/gallery and SDK photo requests
- let native/gallery saves attach to an active photo instead of starting another camera capture
- let SDK `take_photo` reuse the active or very recent local/gallery photo, so the double-shutter ordering saves one JPEG locally and uploads that same JPEG to the miniapp
- keep unrelated SDK-vs-SDK requests guarded with `CAMERA_BUSY`

## Behavior
- Gallery/default button capture starts first, SDK `take_photo` arrives right after: SDK request attaches to the local capture, waits for the local JPEG, then uploads that same file without deleting it.
- SDK capture starts first, gallery/default save arrives while capture is active: gallery save copies the active SDK output into normal gallery storage instead of starting a second capture.
- SDK request arrives while another SDK photo job is active: still returns `CAMERA_BUSY`.

## Validation

## Notes
- Local ignored ASG build prerequisites (`asg_client/.env`, `asg_client/local.properties`, `asg_client/StreamPackLite`) were copied into the Codex worktree for validation only and were not committed.

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/media/core/MediaCaptureService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/media/core/MediaCaptureService.java
@@ -31,11 +31,14 @@ import org.json.JSONObject;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.IOException;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


PhilippeFerreiraDeSousa:
This is a one-off prompt, the desired behavior needs to be discussed actually

PR #3101, opened 2026-06-06
OTA Update for 2026-04-18 Mentra Live units

aisraelov:
## Problem

MTK A/B firmware updates only take effect **after a reboot**. Historically every MTK patch shipped bundled with a BES firmware update, and the BES install power-cycles the device at the MCU level — which applies the staged MTK image as a side effect. So we never needed to reboot ourselves.

We now need to ship an **MTK patch on its own** (no BES update). In that case nothing triggers the reboot, so:

- The device keeps reporting its pre-update MTK firmware version (`ro.custom.ota.version` doesn't change until reboot).
- The OTA manifest keeps matching the same patch (current firmware still equals the patch's `start_firmware`).
- The phone re-runs the "update available" → install → re-check cycle in an **infinite loop**.

## Fix

When the **final** OTA step in a session is MTK (i.e. no BES step follows), reboot the device once the MTK install reports `SUCCESS`, via the existing `SysControl.reboot()` path (the same broadcast already used by the phone-initiated reboot command). After reboot the device reports the new firmware version, the patch no longer matches, and the loop ends.

- The reboot decision is made **before** advancing the session, checking that the current step is `mtk` and is the last step. The order is always APK → MTK → BES, so "MTK is last" ⇔ "no BES follows".
- When a BES step **does** follow, behavior is unchanged — BES still triggers the power-cycle and we do **not** reboot.
- The reboot is delayed ~3s so the phone receives the BLE `complete` status before the connection drops, letting its UI settle on "complete" rather than a bare "disconnected" spinner.

All changes are confined to the MTK `SUCCESS` handler in `OtaService.java`; `SysControl`, `Handler`, `Looper`, and `OtaSessionManager` were already imported.

## Notes

- This is the glasses-side change only. A small, optional phone-side follow-up (`deriveOtaDisplayState.ts`) could give the MTK reboot the same polished "restarting → complete" UX that BES has; without it the phone still handles this gracefully (shows "complete" via the generic status rule, self-heals on reconnect if a message is missed). No false-failure risk: the stall watchdog is disabled while disconnected and the only remaining timer is the 20-min global cap.

## Testing

- [ ] Manual: ship to a Mentra Live on firmware matching an MTK-only patch; confirm it installs, auto-reboots once, comes back on the new firmware, and the phone no longer loops.
- [ ] Manual: MTK + BES bundled update still applies both with a single (BES-triggered) reboot and no double-reboot.

---

> [!NOTE]
> **High Risk**
> Changes firmware OTA completion to trigger a full device reboot via SysControl; incorrect flag handling could double-reboot or skip reboot, but logic is gated by install-time besUpdateFollows and consume-and-clear on SUCCESS.
> 
> **Overview**
> Fixes an **infinite OTA loop** when shipping an **MTK patch without BES**: staged MTK A/B firmware does not apply until reboot, so the device keeps matching the same manifest patch until something power-cycles it (previously that was always the bundled BES install).
> 
> **`OtaHelper`** arms a **`rebootAfterMtkInstall`** flag when MTK install starts (`rebootAfterMtkInstall = !besUpdateFollows`). Combined MTK+BES installs pass **`besUpdateFollows=true`** so the glasses do not self-reboot before BES. **`consumeRebootAfterMtkInstall()`** returns and clears the flag so duplicate SUCCESS events cannot schedule a second reboot.
> 
> **`OtaService`** on MTK **SUCCESS** reads that flag; when true it calls **`scheduleMtkRebootToApplyUpdate()`**, which waits **3s** (BLE `complete` to the phone) then **`SysControl.reboot()`**. MTK+BES behavior is unchanged (no self-reboot when BES follows).
> 
> Also bumps **ASG client to 38.0** and updates **`prod_live_version.json`** / test manifest with **38** APK metadata and **additional MTK patch** entries for more `start_firmware` paths to `MentraLive_20260525`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acb981c702ffb47bbec95bef65f83f3ea05982d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Auto-reboots the device after a successful MTK-only OTA so the staged firmware applies and the update loop stops. Reboot is delayed ~3s so the phone receives the BLE “complete” status first.

- **Bug Fixes**
  - Reboot decision is now driven by an install-time flag `rebootAfterMtkInstall` set in `OtaHelper` at MTK kickoff, so it works for both session and legacy/no-session paths.
  - Combined MTK+BES updates pass `besUpdateFollows=true` to suppress self-reboot (BES power-cycles); MTK-only defaults to self-reboot.
  - On MTK `SUCCESS` in `OtaService`, send `FINISHED` to the phone and, if `consumeRebootAfterMtkInstall()` is true, schedule `SysControl.reboot()` after `MTK_REBOOT_DELAY_MS`; the read-and-clear avoids double reboots from duplicate/late events.

- **Dependencies**
  - Bumped app to `38 (38.0)` and updated OTA website manifests (`prod_live_version.json`, `test_bes_ota_prod_live_version.json`) to point to the 38 APK, add new MTK patch ranges, and refresh SHA256 hashes.

<sup>Written for commit acb981c702ffb47bbec95bef65f83f3ea05982d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/asg_client/app/build.gradle
+++ b/asg_client/app/build.gradle
@@ -192,8 +192,8 @@ android {
 
         minSdk 28
         targetSdk 33
-        versionCode 37
-        versionName "37.0"
+        versionCode 38
+        versionName "38.0"
 
         ndk {
             abiFilters "armeabi-v7a", "arm64-v8a"
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
@@ -179,6 +179,12 @@ public class OtaHelper {
     private static volatile boolean mtkUpdatedThisSession = false;
     private static volatile boolean isBackgroundPrefetchInProgress = false;
 
+    // True when the in-flight MTK install is the final firmware step (no BES update follows).
+    // BES installs power-cycle the device themselves; an MTK-only update has nothing to trigger
+    // the reboot its staged A/B image needs, so OtaService reboots on MTK success. Set at install
+    // kickoff so it is correct on both the session and legacy/no-session completion paths.
+    private volatile boolean rebootAfterMtkInstall = false;
+
     private volatile boolean pendingPhoneInstall = false;
 
     /** Snapshot for {@link #buildMinimalOtaStatusJson()} when no OTA session exists (aligns with {@link #sendMtkInstallProgress} shape). */
@@ -343,6 +349,16 @@ public class OtaHelper {
 
     public OtaSessionManager getSessionManager() {
         return sessionManager;
+    }
+
+    /**
+     * Whether the most recently started MTK install should trigger a self-reboot on success —
+     * i.e. it is an MTK-only update with no BES step to power-cycle the device. Set by
+     * {@link #checkAndUpdateMtkFirmware} at install kickoff and read by OtaService's MTK SUCCESS
+     * handler, so it works regardless of whether an OTA session is active.
+     */
+    public boolean shouldRebootAfterMtkInstall() {
+        return rebootAfterMtkInstall;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


aisraelov:
@codex review

PhilippeFerreiraDeSousa (approved):

aisraelov:
@codex review

PR #3110, opened 2026-06-08
Fast-fail local photo receiver uploads

PhilippeFerreiraDeSousa:
## Summary

- Prefer the phone's active LAN IPv4 address, and when known, an address on the same `/24` as the glasses Wi-Fi IP, when the Android photo receiver publishes its upload URL.
- Keep public webhook behavior unchanged, but use a 1s connect timeout for local/private photo receiver URLs so unreachable direct phone uploads fall back to BLE quickly instead of waiting the full 5s connect timeout.

## Root Cause

Live logs showed the photo receiver and stream receiver were both advertising the phone LAN address `192.168.50.176`, while the glasses were on `192.168.50.43`. The photo path sent:

```text
MentraPhotoReceiver: Photo receiver ready at http://192.168.50.176:8787/upload
BLE_TRACE ... type=take_photo ... "webhookUrl":"http:\/\/192.168.50.176:8787\/upload"
```

The photo then fell back through BLE:

```text
photo_status ... "status":"ble_fallback_compression"
BlePhotoUploadService: Uploading BLE fallback photo to local receiver via loopback: http://127.0.0.1:8787/upload
photo_relay_upload_end ... "success":true,"outcome":"uploaded"
```

After app restart, stream failed with the same destination:

```text
MentraGStreamer: GStreamer WHIP receiver PLAYING
BLE_TRACE ... type=start_stream ... "streamUrl":"http:\/\/192.168.50.176:8190\/whip\/endpoint"
WHIP request failed: failed to connect to /192.168.50.176 (port 8190) from /192.168.50.43 ... after 10000ms
```

Direct reachability checks confirmed both devices were on `192.168.50.0/24`, but the glasses could not reach the phone:

```text
phone wlan0:   192.168.50.176/24
asg wlan0:     192.168.50.43/24
ping 192.168.50.176: 2 packets transmitted, 0 received
phone nc -l -p 9999 + glasses nc -w 3 192.168.50.176 9999: nc: Timeout
```

So the current field failure is not photo-specific. The LAN path is unreachable for both photo upload and WHIP streaming. This PR does not pretend to fix an isolated/AP-client-isolation network; it makes the photo path choose a better receiver address when multiple interfaces exist and avoid the avoidable 5s wait before BLE fallback for local receiver URLs that cannot be reached.

## Validation

- `git diff --check`
- `cd mobile/modules/bluetooth-sdk && bun run build`
- `cd asg_client && ANDROID_SERIAL=0123456789ABCDEF ./gradlew :app:installDebug`
- Live device log capture from phone `RFCX71TH0CR` and glasses `0123456789ABCDEF`

`cd mobile && bun compile` was also run after dependency install, but it is currently blocked by unrelated existing TypeScript errors in `src/services/asg/gallerySyncService.test.ts` at the `resolveConnectivity?.()` calls.

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/media/core/MediaCaptureService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/media/core/MediaCaptureService.java
@@ -71,6 +71,8 @@ public class MediaCaptureService {
     // Debug flag: Enable detailed end-to-end photo capture timing logs
     // true = log timing from request to capture, false = suppress timing logs
     private static final boolean ENABLE_PHOTO_TIMING_LOGS = true;
+    private static final int DEFAULT_PHOTO_UPLOAD_CONNECT_TIMEOUT_SECONDS = 5;
+    private static final int LOCAL_PHOTO_UPLOAD_CONNECT_TIMEOUT_SECONDS = 1;
 
     private final Context mContext;
     private final MediaUploadQueueManager mMediaQueueManager;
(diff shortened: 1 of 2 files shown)


PhilippeFerreiraDeSousa:
We want this fail fast regardless of whether the web hook url is local or in the cloud

PR #2865, opened 2026-05-22
Add developer OTA manifest URL setting

PhilippeFerreiraDeSousa:
## Summary
- add a developer setting for overriding the OTA version manifest URL
- thread the configured manifest URL through manual OTA checks, progress screens, and phone-initiated OTA start
- allow ASG/OTA updater code paths to receive the manifest URL override instead of always using the production default

## Why
This lets us test a local or tunneled OTA manifest while keeping the production URL as the normal default.

## Notes
- This PR is stacked on `codex/hackathon-feedback-fixes` to keep the diff limited to the OTA URL setting.
- The local BES manifest helper script is intentionally not included in this repo.

## Validation
- `git diff --check b247ed102..HEAD`
- inspected the committed diff for private BES source paths

PR #3113, opened 2026-06-09
Use phone-driven ASG OTA manifests on staging

PhilippeFerreiraDeSousa:
## Summary

- Keeps the ASG BLE `ota_start` command source-compatible while allowing an optional `ota_version_url` manifest URL. Old no-URL OTA behavior is unchanged and still uses the compiled production default.
- Removes the ASG autonomous cached-manifest/prefetch/cache-ready path for OTA discovery. The phone/SDK checks the manifest, then ASG downloads and installs from the URL supplied with `ota_start`.
- Updates MentraOS mobile so a phone-side positive OTA check prompts the user directly instead of waiting for ASG `cache_ready`; legacy `cache_ready` is still accepted for older glasses and duplicate prompts are suppressed.
- Keeps `ota_retry_version_check` compatible by retrying the last phone-started manifest URL after phone-side recovery work such as clock sync.
- Moves ASG `versionCode` to the same wall-clock build-number strategy as MentraOS mobile: `100000 + seconds since 2025-01-01T00:00:00Z`, with CI/local overrides for pinned builds.
- Adds `asg_client/ota_manifests/firmware_live.json` as a source-only MTK/BES firmware manifest so staging output is no longer composed from `prod_live_version.json`.
- Updates the staging builds workflow to pin the ASG build number, upload the ASG APK, generate `staging_live_version.json`, upload it to the `staging-builds` GitHub release, and deploy the staging-only OTA payload to the dedicated non-Git Cloudflare Pages project `mentra-live-ota-staging`.

## Scope

This is the staging ASG release/bootstrap slice. It intentionally keeps the scope to BLE/SDK URL plumbing, phone-driven OTA prompting, staging manifest generation, and a dedicated staging Pages deployment. It does not include developer settings parameterization, channel/config frameworks, R2 deployment, or the larger flow from #2865.

## Compatibility

For the first mixed-version update, where MentraOS mobile has this change but ASG is still old code with an empty cache, the phone still sends a plain `ota_start` command. The old ASG implementation acknowledges that command and, when `cachedVersionJson == null`, falls through to `startVersionCheck(context)`, so it downloads and installs from its compiled default manifest. If the old ASG prefetch is already in progress, it queues the phone-initiated install as before.

## Firmware manifest process

Staging manifests must always include MTK and BES firmware metadata because glasses can be updating from an old state and absence of those sections is interpreted as no firmware update available. CI now starts from the committed source-only firmware baseline in `asg_client/ota_manifests/firmware_live.json`, verifies that `mtk_patches` and `bes_firmware` are present, and combines those sections with the freshly built ASG APK metadata.

The generated staging payload contains only `staging_live_version.json`, `_headers`, and a small index page. It fails if `prod_live_version.json` appears in that payload. The deployment target is the dedicated Cloudflare Pages project `mentra-live-ota-staging`, which has no Git provider, so commits cannot trigger stale automatic Pages deployments. Production remains the existing `main` branch flow to `ota.mentraglass.com/prod_live_version.json`.

The workflow passes `--branch staging` to `wrangler pages deploy` because `staging` is now the Cloudflare Pages production-branch label for this standalone staging-only Pages project. This workflow is triggered only by GitHub `staging`; using the same branch label in Pages makes the intent explicit while still publishing the generated staging-only payload to the Pages project’s production environment, so `staging.ota.mentraglass.com` gets a stable canonical deployment instead of a preview URL. The payload guard above still prevents any `prod_live_version.json` from being deployed there.

## Staging OTA URL

The initial staging URL is:

`https://mentra-live-ota-staging.pages.dev/staging_live_version.json`

The intended custom domain is:

`https://staging.ota.mentraglass.com/staging_live_version.json`

DNS has been updated for `staging.ota.mentraglass.com`, but the hostname still needs to be attached under the Cloudflare Pages project `mentra-live-ota-staging` before it will serve correctly. Until the first CI deploy lands, the new Pages project may return a Cloudflare 522 because there is no deployment yet.

## Why this is minimal and safer

The staging/dev team and Bluetooth SDK can make OTA availability decisions by fetching the manifest directly. ASG only needs to accept the chosen manifest URL at install time and avoid reusing stale cached prod manifest data. Removing background ASG prefetch avoids hidden state mismatches while preserving the explicit no-URL production default path for existing MentraOS users. This gives Bluetooth SDK OTA delivery a safer bootstrap without adding settings UI or a broader rollout framework.

## Validation

- Created Cloudflare Pages project `mentra-live-ota-staging` with `Git Provider: No` and production branch `staging`.
- Confirmed `staging.ota.mentraglass.com` DNS is resolving through public resolvers, but Pages custom-domain attachment is still required.
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/staging-builds.yml"); puts "yaml ok"'`
- `jq -e '(.mtk_patches | type == "array" and length > 0) and (.bes_firmware | type == "object")' asg_client/ota_manifests/firmware_live.json`
- Local `jq` manifest generation check confirmed the generated staging manifest contains the new ASG APK entry, non-empty `mtk_patches`, and `bes_firmware`.
- Local staging Pages payload check confirmed `staging_live_version.json` is added and `prod_live_version.json` is not present.
- `cd asg_client && ./gradlew :app:testDebugUnitTest --tests com.mentra.asg_client.service.core.handlers.OtaCommandHandlerTest`
- `cd asg_client && ./gradlew :app:assembleRelease -PASG_VERSION_CODE=45555044 -PASG_VERSION_NAME=staging.localtest`
- `/Users/philippe/Library/Android/sdk/build-tools/36.1.0/aapt dump badging asg_client/app/build/outputs/apk/release/app-release.apk` showed `versionCode=45555044` and `versionName=staging.localtest`.
- `cd mobile && bun run test -- OtaUpdateChecker.test.ts --runInBand` was attempted but blocked before tests ran because Jest cannot find native module `BluetoothSdk`.
- `cd mobile && bun compile` was attempted but is blocked by existing unrelated TypeScript errors in `src/effects/Compositor.tsx`, `src/i18n/translate.ts`, `src/services/MantleManager.ts`, and `src/stores/capsule.ts`.

---

> [!NOTE]
> **Medium Risk**
> Changes affect glasses OTA delivery (manifest URLs, version codes, and release asset cleanup); mistakes could point staging devices at wrong builds or prune APKs still referenced by a live manifest, though guards and ref checks reduce that risk.
> 
> **Overview**
> Extends **staging builds** so ASG releases drive a dedicated **staging OTA channel** without baking staging URLs into mobile binaries.
> 
> **ASG versioning:** `asg_client` no longer uses fixed `versionCode`/`versionName`; it uses **100000 + seconds since 2025-01-01 UTC** (same idea as mobile), with **`ASG_VERSION_CODE` / `ASG_VERSION_NAME`** overrides so CI and manifests match the built APK.
> 
> **Workflow:** Adds **`staging-builds` concurrency**, documents a third publish target (**`mentra-live-ota-staging`** Pages), computes ASG version + APK **sha256/size** in **`build-asg`**, passes pinned versions into Gradle, then in **`upload-releases`** generates **`staging_live_version.json`** from **`firmware_live.json`** + dated GitHub APK URL, uploads rolling/dated manifest assets, deploys a small Pages site (with **`_headers`** and a guard against **`prod_live_version.json`**), and **verifies** the live manifest **`versionCode`**. Manifest upload and Pages deploy run only on **`github.ref_name == staging`**. Release cleanup **keeps** `staging_live_version.json` and the APK referenced by the live manifest, and **skips deleting ASG APKs** if the manifest cannot be read.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4079d5ecae88955034403cd1a3c230b609500eef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PhilippeFerreiraDeSousa:
@codex review

aisraelov (commented):
Did a deep pass on this — read the full diff and traced the plumbing end-to-end (TS → Expo native modules → BLE JSON → `OtaCommandHandler` → `OtaHelper` → session persistence), plus the workflow job graph. The code itself is in good shape: the manifest URL correctly survives the APK-install restart via the OTA session (`createSession(steps, versionUrl)`), the `phoneInitiatedRetryPending` gating interacts correctly with both clock-sync retry paths, deleting the prefetch state machine removes a whole class of races (`pendingPhoneInstall` / suppressed-FINISHED), and re-enabling APK SHA256 also makes `isCachedArtifactValid` actually validate cached APKs against the current manifest. Nothing in the code I'd block on.

The feedback is about the URL strategy and rollout sequencing:

### 1. Do we need `staging.ota.mentraglass.com` + a second Pages project at all?

First instinct was "why not just `ota.mentraglass.com/staging_live_version.json`?" — but after digging, that's *not* actually simpler: `ota.mentraglass.com` is the Git-connected Pages project serving `asg_client/ota_website/` from `main`. Pages deployments are atomic whole-site snapshots and Git-connected projects don't take `wrangler pages deploy` direct uploads, so putting the staging manifest on that domain means staging CI either pushes commits to `main` on every staging build, or becomes a writer of the prod site payload (and could clobber `prod_live_version.json`) — exactly what this PR's payload guard is designed to prevent. So a single domain quietly turns "anyone who can push to `staging`" into "can affect prod OTA". The separate project is a real blast-radius boundary.

That said, two simpler options than a new custom domain:

- **Simplest — no Cloudflare at all:** this workflow already uploads `staging_live_version.json` as a stable-named asset on the `staging-builds` release. `https://github.com/Mentra-Community/MentraOS/releases/download/staging-builds/staging_live_version.json` could just *be* the staging manifest URL. Both ends are proven: the phone's `fetchVersionInfo` uses `fetch` (follows redirects), and the glasses already download MTK/BES firmware from GitHub release URLs in these same manifests via the same HTTP stack. Kills the Pages project, the custom domain, the wrangler step, and the whole bootstrap problem. Main caveat: GitHub reachability for any staging testers in China.
- **Middle — keep the Pages project, drop the custom domain from the critical path:** make `https://mentra-live-ota-staging.pages.dev/staging_live_version.json` the canonical URL in `ASG_STAGING_OTA_VERSION_URL` / preflight / post-deploy verification. It's live the moment the first deploy lands, with no DNS/attachment ops. Attach `staging.ota.mentraglass.com` later as a nicety if we want it, not load-bearing.

If we do keep the custom domain as-is, sequencing matters more than the description suggests: the new preflight steps hard-fail **both mobile staging jobs** (so no staging TestFlight/Play uploads at all) until the domain is attached and serving — the "may return a Cloudflare 522" note undersells that. The domain would need to be attached + one bootstrap deploy done *before* merge.

### 2. versionCode: gradle side already covers prod — let's lean into it, but the prod manifest process has to follow

To be clear for other readers: the `build.gradle` change is not staging-scoped — every build (local, staging, prod release) now gets the wall-clock versionCode, with CI pinning via `-PASG_VERSION_CODE` so manifests match the APK. I think that's right and we should keep it.

The dangerous half is `asg_client/ota_website/prod_live_version.json`, which is hand-edited (currently `versionCode: 38`, and note `com.augmentos.otaupdater` still has `"sha256": "example_sha256_hash_here"`). Once the first wall-clock prod build ships (~45.6M today), `getInstalledVersion()` on updated devices returns the real installed code — so if anyone hand-types `39` into the prod manifest out of habit, `serverVersion > currentVersion` can never pass again and updated devices are **permanently stranded** (silently — no error anywhere). Asks:

- Generate the prod manifest entry from the built APK the same way the staging job does (`aapt dump badging` for versionCode + `shasum -a 256`), or at minimum add a CI check on changes to `ota_website/prod_live_version.json` asserting the new versionCode is `> 100000` and greater than the previous value.
- Worth stating in the release runbook: OTA downgrade is now impossible by construction (Android won't downgrade) — emergency rollback means re-shipping older code as a new, higher versionCode. Same model as mobile, but it's new for ASG.
- Cosmetic: prod builds without `-PASG_VERSION_NAME` now report `versionName` as the raw digits instead of `"38.0"` — fine, but the prod release flow should probably pin a human-readable name like CI does.

### 3. Cleanup can delete the APK the live staging manifest points to

The deployed manifest's `apkUrl` references the **dated** asset (`asg-staging-<date>-<sha>.apk` — correct, since `asg-latest.apk` would break sha256). But the 7-day cleanup only exempts `-latest` files and `staging_live_version.json`, and it still runs when `build-asg` failed (the upload script skips missing artifacts gracefully, so the job stays green and reaches cleanup). If the ASG build is broken/skipped for >7 days while mobile keeps shipping, cleanup deletes the dated APK that the *still-live* Pages manifest references → every staging OTA download 404s. Suggested fix in the cleanup script:

```js
// before the deletion loop
let keepApk = null;
const manifestAsset = release.data.assets.find(a => a.name === 'staging_live_version.json');
if (manifestAsset) {
  const res = await fetch(manifestAsset.browser_download_url);
  const manifest = await res.json();
  keepApk = (manifest?.apps?.['com.mentra.asg_client']?.apkUrl ?? '').split('/').pop() || null;
}
// in the loop
if (asset.name === keepApk) continue;
```

### 4. Staging phone + prod-firmware glasses = recurring dead-end prompt

A staging mobile build checks the **staging** manifest (wall-clock code ≫ installed 38 → "update available"), but old glasses ignore `ota_version_url` and check the **prod** manifest, find nothing newer, and send no progress → the progress screen's watchdog times out into a "stalled" error. This recurs on every connection for every staging tester whose glasses still run prod firmware — the Compatibility section only covers the case where prod has something to install. The wall-clock scheme gives a clean legacy threshold; suggest gating the env override on the glasses being new enough to honor the URL:

```ts
export function getAsgOtaVersionUrl(glassesUrl?: string | null, glassesBuildNumber?: string | null): string {
  const deviceUrl = glassesUrl?.trim()
  // Legacy glasses ignore ota_version_url and install from their compiled
  // default — check the manifest they will actually use, not the env override.
  const legacy = !!glassesBuildNumber && parseInt(glassesBuildNumber, 10) < 100_000
  if (legacy) return deviceUrl || OTA_VERSION_URL_PROD
  const envUrl = process.env.EXPO_PUBLIC_ASG_OTA_VERSION_URL?.trim()
  if (envUrl) return envUrl
  return deviceUrl || OTA_VERSION_URL_PROD
}
```

### Nits (no action required)

- `OtaHelper.java` — `lastVersionJsonUrl` is assigned (line ~770) *before* the `isPhoneInitiatedOta` gate (~783), so a non-phone caller (today: debug receivers only) silently clobbers the retry URL. Cheap hardening: move the assignment below the gate.
- APK verify failure reuses `FirmwareDownloadException.CODE_VERIFY_FAILED` (`firmware_verify_failed`), so the phone's error copy will talk about firmware for an APK failure — consider an `apk_verify_failed` mapping.
- `OTA_VERSION_URL_STAGING` is exported but never used.
- `progress.tsx` retry path double-arms the watchdogs (`sendOtaStartWithWatchdogs` arms internally, then the `.then(armAckAndStuckWatchdogsOnly)` arms again) — harmless, but the `.then` can go. (Side note: routing the retry through `sendOtaStartRef.current()` is a nice catch — a bare retried `ota_start` would have silently fallen back to the prod manifest.)
- `ASG_CLIENT_API.md` still describes `ota_retry_version_check` as a generic background re-check; it's now gated on a failed phone-started OTA with `clock_skew`/`ssl_error`.
- FYI: the legacy companion app `asg_client/ota_updater` still runs its own autonomous prefetch loop against the prod manifest — inert for staging devices (prod codes sort below wall-clock), but worth confirming whether that APK is still deployed anywhere if the goal is "updates only via `ota_start`".

Bottom line: code looks good; the decisions to settle before merge are (1) the staging URL strategy (I'd vote GH release asset or pages.dev-only — both remove the domain-attachment dependency), and (2) the prod manifest discipline for wall-clock versionCodes. (3) is a small workflow patch, (4) a small mobile patch.

PhilippeFerreiraDeSousa:
Codex review of Israelov's review:

Most of the urgent parts are already addressed by the last commits. I would not add more to this PR right now unless CI or review finds a concrete blocker.

### What Is Already Addressed

- Custom domain dependency: current workflow uses https://mentra-live-ota-staging.pages.dev/staging_live_version.json as the canonical staging URL. staging.ota.mentraglass.com is now optional/non-fatal.
- Cleanup deleting live APK: fixed. Cleanup now reads staging_live_version.json and preserves the exact dated APK asset referenced by the live manifest.
- Staging phone + old prod glasses loop: fixed. getAsgOtaVersionUrl(...) detects legacy ASG builds < 100000 and checks the URL old glasses will actually use, instead of falsely prompting from staging.
- Job ordering: fixed. Mobile/ASG artifacts upload before manifest generation/Pages deploy, so manifest failure does not block mobile artifact publishing.
- OTA capability not tied to WiFi: fixed with hasOta.

### Still Valid But Not Necessarily This PR

The biggest real remaining point is prod manifest discipline once wall-clock ASG version codes reach production. That is the scary one. If prod release stays hand-edited and someone writes 39 after devices have installed 45_000_000+, those devices will never see prod APK updates again.
I would treat that as a required follow-up before any prod ASG release with the new versionCode scheme:
- prod manifest generated from the actual built APK, or
- CI check preventing low/manual version codes in prod_live_version.json.

### My Opinion

The PR is complex, but the complexity is now mostly staging bootstrap and safety rails, not random product surface. I would keep it as-is and stop adding features. The next safest move is to let CI/review settle, then merge to staging and test the real staging OTA loop.
The line I would hold: no channels, no settings UI, no fallback from staging to prod, no more clever cache behavior. For OTA, boring is beautiful.

PhilippeFerreiraDeSousa:
## Deep review @ `7d70718` — code + audit of all prior review comments

**Verdict: in good shape and mergeable.** Every blocking item from the bot review passes and aisraelov's deep review is genuinely fixed at HEAD — each verified against the checked-out code, not the "addressed" markers. What's left: a handful of small non-blocking items, one CI infra problem, and the already-identified prod-manifest-discipline follow-up.

### CI status — not a code problem

The ASG PR build failure on `7d70718` is environmental: self-hosted runner 4 is missing `credentials/recovery-keystore.jks` for the "Build recovery worker asset" step ([run logs](https://github.com/Mentra-Community/MentraOS/actions/runs/27437832145)). The same workflow also failed twice on `dev` today, and this PR touches nothing near the recovery worker. Fix is ops-side (link the keystore on runner 4), then re-run.

### Prior review comments — verified status

✅ **Confirmed fixed at HEAD:**

- **WiFi-prompt family** (Bugbot ×3, cubic P1): the no-WiFi alert uses its own `hasPromptedOtaWifiSetup` flag, and the pending-update effect waits for WiFi and re-fires on `glassesWifiConnected` — all four entry paths covered ([OtaUpdateChecker.tsx#L517-L557](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/mobile/src/effects/OtaUpdateChecker.tsx#L517-L557))
- **Unsolicited clock-sync OTA** (Bugbot High): gated on `phoneInitiatedRetryPending` ([OtaHelper.java#L735-L738](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java#L735-L738)), and the retry reuses the last phone-sent URL stored under the version-check lock — not the prod default
- **"Pass the staging URL with ota_start"** (Codex P1): `ota_version_url` is now sent end-to-end on both platforms (progress screen → module → DeviceManager → MentraLive, same JSON key on Android and iOS), and the glasses fetch exactly that URL
- **sha256 skip-on-missing** (cubic P1): APK now rejects when the hash is missing ([OtaHelper.java#L1654-L1658](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java#L1654-L1658)); the fd leak (cubic P3) is fixed with try-with-resources
- **Retry double-count** (cubic P2): `retryAlreadyCounted` marker — one slot per attempt
- **Malformed `_headers` heredoc** (Bugbot): now generated with `printf`, URL patterns at column 0
- **Fatal preflight curls blocking mobile builds** (Bugbot High): removed entirely; the URL is only injected as a build-time env var
- **Custom-domain verify fatal / missing retries** (Bugbot + cubic): pages.dev check is fatal with 12×10s retries; custom-domain check is non-fatal with the same retry budget
- **Cleanup deleting the manifest-referenced APK** (Bugbot + cubic + review item 3): keep-logic fetches `staging_live_version.json` and preserves the exact dated asset, and it runs even when `build-asg` failed
- **"Manifest failure blocks mobile publishing"** (Bugbot's latest pass): **stale** — fixed in [`0d9e0c2`](https://github.com/Mentra-Community/MentraOS/commit/0d9e0c2de) (after the commit Bugbot reviewed). Artifacts upload at [line 600](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/.github/workflows/staging-builds.yml#L600) before manifest generation at [line 693](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/.github/workflows/staging-builds.yml#L693), and all manifest/Pages steps are gated on `needs.build-asg.result == 'success'`

The earlier "Codex review of Israelov's review" summary holds up: all five "already addressed" claims verified, including the legacy-glasses gate — `getAsgOtaVersionUrl` can't be defeated by a transiently-missing build number, because every prompt-generating caller already guards on `buildNumber` being present before checking.

❌ **Still open (all small):**

1. **APK hash failure reports `firmware_verify_failed`** (the one real Medium left): [OtaHelper.java#L1554](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java#L1554) sets the firmware code, so the phone renders firmware-worded copy plus the Change-WiFi affordance ([otaErrorMapping.ts#L52](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/mobile/src/utils/otaErrorMapping.ts#L52)) for an APK integrity failure. ~10-line fix: `apk_verify_failed` code + mapping.
2. [ASG_CLIENT_API.md](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/asg_client/docs/ASG_CLIENT_API.md?plain=1#L887) still describes `ota_retry_version_check` as a generic background re-check; it's now ignored unless a phone-started OTA failed with `clock_skew`/`ssl_error`.
3. `OTA_VERSION_URL_STAGING` ([OtaUpdateChecker.tsx#L59](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/mobile/src/effects/OtaUpdateChecker.tsx#L59)) is a dead export — nothing imports it.
4. `lastVersionJsonUrl` is still assigned before the `isPhoneInitiatedOta` gate ([OtaHelper.java#L787](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java#L787) vs L800) — latent only; every current caller sets phone-initiated mode first.
5. progress.tsx still double-arms the watchdogs (nit; timers are cleared before re-arming, so no stacking).
6. FYI confirmed: `asg_client/ota_updater` still autonomously polls the prod manifest every 15 min ([Constants.java#L7](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/asg_client/ota_updater/app/src/main/java/com/augmentos/otaupdater/helper/Constants.java#L7)) — untouched by this PR; worth confirming whether that APK is still deployed anywhere.

### New findings (verified, none blocking)

- **Low — no `concurrency:` group** in staging-builds.yml. Two quick pushes to staging can deploy manifests to Pages out of order — the older manifest can land last and stay live, and both runs' verification loops pass (each only checks that its own versionCode is served at poll time). Seven other workflows in the repo set one. Fix: `concurrency: {group: staging-builds, cancel-in-progress: false}`.
- **Low — branch-gating mismatch**: "Upload ASG OTA manifest" ([line 776](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/.github/workflows/staging-builds.yml#L776)) lacks the `github.ref_name == 'staging'` guard that the Pages deploy ([line 863](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/.github/workflows/staging-builds.yml#L863)) has. A `workflow_dispatch` from a non-staging branch overwrites the release-asset manifest without updating pages.dev — and the cleanup keep-APK logic trusts the release-asset manifest, so the two can desync.
- **Low — cleanup keep-logic fails open**: the try/catch around the manifest fetch ([lines 938-949](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/.github/workflows/staging-builds.yml#L938-L949)) never checks `response.ok`, and on any failure it logs and proceeds with deletion. Combined with a week of failed ASG builds, that's the one remaining path where the live APK gets deleted. Safer: skip deleting `asg-staging-*.apk` entirely when the manifest can't be read.
- **Low — OTA URL trust model (deliberate design, but decide consciously)**: `getValidatedOtaVersionUrl` accepts `http://` and any host ([OtaCommandHandler.java#L142](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/OtaCommandHandler.java#L142); the unit test deliberately asserts `http://localhost:8000` passes), the manifest sha256 is self-referential, and the app never checks the APK signer itself — source pinning now rests entirely on trusting the paired phone plus the platform installer's signature check. The http allowance exists for the localhost dev workflow. Consider https-only + host allowlist outside debug builds as defense-in-depth. Pre-existing adjacent gap: MTK firmware still *skips* hash verification when sha256 is missing (APK and BES both reject) — untouched by this PR, and the committed `firmware_live.json` has hashes everywhere, but worth aligning.
- **Nits**: the optional custom-domain verify burns ~110s of sleeps per run while the domain is unattached (it currently fails instantly with an SSL alert — DNS resolves but the hostname isn't attached to the Pages project); `npx wrangler@latest` is unpinned vs. the repo's `wrangler@4.86.0` convention ([deploy-dev-docs.yml#L61](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/.github/workflows/deploy-dev-docs.yml#L61)); a manifest-step failure still fails the `upload-releases` job, which skips cleanup and makes Slack report "Staging build failed" even though mobile artifacts were published.

### Verified clean

Cross-platform plumbing is fully consistent — TS optional param → Kotlin `String? = null` defaults → Swift protocol updated in all eight SGC/controller implementations (compile-enforced), same `ota_version_url` JSON key on both platforms. `hasOta` is typed in [hardware.ts#L154](https://github.com/Mentra-Community/MentraOS/blob/7d707181c5aba90b171cf4e0ef0f6f8c980d35d3/cloud/packages/types/src/hardware.ts#L154), true only for Mentra Live across both capability hierarchies, and actually gates the check, the pairing stack, and the settings UI. The wall-clock versionCode gradle logic is correct (pin validation, overflow horizon ~2093). `firmware_live.json` is valid with 64-char sha256 on every entry. `OtaCommandHandlerTest` covers null/https/http/whitespace/`file://` URL cases. The OTA session persists the manifest URL across the APK-install restart. The `gallerySyncService.test.ts` tweaks are harmless lint hardening.

### Suggested pre-merge batch (~30 lines, two files)

`apk_verify_failed` code + mapping; `concurrency:` group; `github.ref_name == 'staging'` guard on the manifest-upload step; fail-closed cleanup; pin wrangler; delete the dead export; fix the `ota_retry_version_check` doc paragraph. Everything else — prod-manifest CI guard before the first wall-clock prod release, `ota_updater` deprecation decision, https-only hardening — stays follow-up. For OTA, boring is beautiful.

PhilippeFerreiraDeSousa:
**Removed the staging OTA URL bake from the mobile builds** (`d9c6fbd`): the TestFlight / Play internal binaries this workflow produces are the exact binaries later promoted to public beta and the app stores, so a compiled-in `EXPO_PUBLIC_ASG_OTA_VERSION_URL` would ship to customers — and once their glasses run wall-clock ASG builds (past the legacy gate), their phones would drive glasses OTA from the **staging** manifest. The legacy gate masks this today (current glasses report versionCode 38 < 100000), which is what made it easy to miss.

Promoted binaries now fall through to the glasses-reported URL / production manifest. Staging testers opt into the staging manifest via the developer-settings **Custom OTA Manifest URL** card (Staging preset) added in #3147, which is stacked on this PR and becomes the load-bearing opt-in mechanism.

For reference, the ASG client APK side was audited for the same promotion concern and is clean: staging ASG builds use prod backend defaults (`.env.example`), compile the prod manifest as their default OTA URL, report that prod URL in `version_info`, and are signed with the same release cert as the prod `asg-client-38.apk` (verified: identical SHA-256 cert digest) — so a future prod release CI that copies the staging-built ASG APK is viable.

PhilippeFerreiraDeSousa:
So I pushed https://github.com/Mentra-Community/MentraOS/pull/3113/changes/d9c6fbd1af9d1fd2fa8dcdf0239cc3c4881c1d8e
This is important because MentraOS staging builds were gong to have the staging ota url baked in the apk, so when we promote them as prod release they would still have the staging ota url used.
The AI did not have context on the staging to prod release flow and I flagged this issue @aisraelov

PhilippeFerreiraDeSousa:
## Hardware verification protocol


End-to-end verification of the phone-driven OTA flow (PR #3113 + #3147) on real
hardware, exercising the full sequential update: **ASG client APK → MTK
firmware → BES firmware**. A regression here can brick customer glasses, so
this runs before merging changes to the OTA path.

## Prerequisites

| What | State |
|---|---|
| Glasses (Mentra Live, adb `0123456789ABCDEF`) | Old MTK + old BES flashed; ASG client from the PR branch installed (`cd asg_client && ./gradlew installRelease`) so it understands `ota_start.ota_version_url` |
| Phone (adb `RFCX71TH0CR`) | MentraOS dev build from the PR branch (`cd mobile && bun run android`), Metro running |
| Worktree | `~/dev/MentraOS-codex-asg-ota-start-url`, branch `codex/dev-ota-url-setting` (stacked on `codex/staging-asg-ota-start-url`) |
| ngrok | Reserved domain `philippeatmentraota.ngrok.app` (alias `nota` in `~/.zshrc` → `ngrok http --url=philippeatmentraota.ngrok.app 9876`) |
| Signing | Build on this machine only — OTA installs are rejected unless the APK is signed with the same key as the installed build (`~/.mentra/credentials/asg-keystore.jks`) |

Glasses need real internet (MTK/BES files come from GitHub release URLs); the
phone and glasses do NOT need to share a LAN with the Mac (everything goes
through ngrok).

## 1. Build the update-target APK

Pin a wall-clock versionCode so it is strictly newer than whatever is installed:

```bash
cd ~/dev/MentraOS-codex-asg-ota-start-url/asg_client
NOW_CODE=$((100000 + $(date +%s) - 1735689600))
./gradlew :app:assembleRelease -PASG_VERSION_CODE=$NOW_CODE -PASG_VERSION_NAME=local.otatest
```

Output: `asg_client/app/build/outputs/apk/release/app-release.apk`. Verify:

```bash
~/Library/Android/sdk/build-tools/36.1.0/aapt dump badging app/build/outputs/apk/release/app-release.apk | head -1
```

## 2. Stage files and generate the manifest (same recipe as CI)

```bash
mkdir -p /tmp/ota-local-server
cp app/build/outputs/apk/release/app-release.apk /tmp/ota-local-server/asg-local-otatest.apk
APK=/tmp/ota-local-server/asg-local-otatest.apk
jq --arg apkUrl "https://philippeatmentraota.ngrok.app/asg-local-otatest.apk" \
   --arg versionName "local.otatest" \
   --arg sha256 "$(shasum -a 256 $APK | cut -d' ' -f1)" \
   --argjson versionCode $NOW_CODE \
   --argjson apkSize "$(stat -f%z $APK)" \
   '{apps: {"com.mentra.asg_client": {versionCode: $versionCode, versionName: $versionName,
     apkUrl: $apkUrl, apkSize: $apkSize, sha256: $sha256}},
     mtk_patches: .mtk_patches, bes_firmware: .bes_firmware}' \
   ../asg_client/ota_manifests/firmware_live.json > /tmp/ota-local-server/local_live_version.json
```

`mtk_patches`/`bes_firmware` come from `asg_client/ota_manifests/firmware_live.json`
— the same source CI uses — so the firmware legs of the sequence are the real
staging artifacts.

## 3. Serve it

```bash
cd /tmp/ota-local-server && python3 -m http.server 9876 --bind 127.0.0.1 &
nota   # = ngrok http --url=philippeatmentraota.ngrok.app 9876
curl -s https://philippeatmentraota.ngrok.app/local_live_version.json | jq '.apps["com.mentra.asg_client"].versionCode'
```

## 4. Point the phone at it

Developer Settings → **Custom OTA Manifest URL** →
`https://philippeatmentraota.ngrok.app/local_live_version.json` →
**Save & Test** (must report success — this also proves phone→manifest
reachability). Bookmark it once; reuse via the chip afterwards.

## 5. Run the flow and what to expect

Open **OTA Check for Updates** (Developer Settings → Quick Links), or reconnect
the glasses to trigger the background check. Watch three places:

- phone UI, plus `adb -s RFCX71TH0CR logcat | grep -E "OTA"` if needed
- HTTP server log (every glasses/phone fetch appears here)
- glasses: `adb -s 0123456789ABCDEF logcat | grep -E "ASGClientOTA"`

Expected sequence:

1. Phone GETs the manifest → prompts update (`local.otatest`).
2. Accept → progress screen sends `ota_start` with `ota_version_url=<ngrok manifest URL>`; glasses ack (no ack → watchdog resends, max retries).
3. Glasses GET the manifest, then the APK (84 MB through ngrok — slower than LAN); progress streams to the phone.
4. SHA256 verified → install → ASG client restarts → session resumes from persisted state (manifest URL survives the restart) → phone shows APK step complete.
5. MTK patch downloads (GitHub URL) and installs → glasses reboot for MTK.
6. BES firmware downloads and installs.
7. Phone shows update complete; glasses report the new versionCode/firmware versions in `version_info`.

**Pass** = all three steps complete in order, no manual intervention after the
initial accept, and final reported versions match the manifest.

Key failure signatures: `apk_verify_failed` (sha mismatch — stale manifest vs
rebuilt APK: redo step 2), install rejected (signing key mismatch — wrong
machine/keystore), stalled progress + no server-log hit from glasses (glasses
WiFi/internet), `INVALID_OTA_VERSION_URL` ack (malformed URL in dev setting).

## 6. Iterating

For each new attempt: rerun steps 1–2 (new `NOW_CODE`, fresh sha) — the server
picks up replaced files automatically; ngrok and the dev setting stay as-is.
To re-test on the same glasses without flashing back: any rebuild with a newer
`NOW_CODE` is an "update" again (APK leg only — MTK/BES stay current after the
first successful pass; reflash old firmware to re-test those legs).

## Scenario matrix (minimum before merge)

| # | Scenario | Expected |
|---|---|---|
| 1 | Full sequence (old MTK + old BES + older APK) | APK → MTK → BES, all complete |
| 2 | APK-only (firmware already current) | Single APK step, complete |
| 3 | No update (manifest versionCode ≤ installed) | "Up to date", no prompt |
| 4 | Kill WiFi mid-APK-download | Download fails with actionable error + Change WiFi offer; retry works |
| 5 | Wrong sha256 in manifest | Reject, `apk_verify_failed` surfaced on phone, no install |
| 6 | Legacy glasses (build < 100000, prod firmware) + dev override set | No false prompt from the override (check uses glasses' own manifest) |

## Appendix: exact manifest used in this verification run

```json
{
  "apps": {
    "com.mentra.asg_client": {
      "versionCode": 45716298,
      "versionName": "local.otatest",
      "apkUrl": "https://philippeatmentraota.ngrok.app/asg-local-otatest.apk",
      "apkSize": 84497296,
      "sha256": "2a8f1451466d5b491bca3a6906f4e9d54572529f90cc687b13200f6e5dbf79d6"
    }
  },
  "mtk_patches": [
    {
      "start_firmware": "MentraLive_20260418",
      "end_firmware": "MentraLive_20260525",
      "url": "https://github.com/Mentra-Community/MentraOS/releases/download/asg-client/mtk_ota_update_20260418_20260525.zip",
      "sha256": "169e9b94ea4b8ff283dccd2b2db4102966bbb424dc16a188f9ff28dfb202b256"
    },
    {
      "start_firmware": "MentraLive_20260421",
      "end_firmware": "MentraLive_20260525",
      "url": "https://github.com/Mentra-Community/MentraOS/releases/download/asg-client/mtk_ota_update_20260421_20260525.zip",
      "sha256": "07bf1d53f9527a6dc1cb1165d9edd6e3ff8ecb09e9f2f49ff9abdc157774052e"
    },
    {
      "start_firmware": "MentraLive_20260513",
      "end_firmware": "MentraLive_20260525",
      "url": "https://github.com/Mentra-Community/MentraOS/releases/download/asg-client/mtk_ota_update_20260513_20260525.zip",
      "sha256": "f9f5b30ce7cbb335a1f51296d16a33e3f44fc51171eb921304ea90089c493619"
    },
    {
      "start_firmware": "MentraLive_20260204",
      "end_firmware": "MentraLive_20260525",
      "url": "https://github.com/Mentra-Community/MentraOS/releases/download/asg-client/mtk_ota_update_20260204_20260525.zip",
      "sha256": "6078863f99b28fa7a6a7988e12e001ee6384f928861a3129f50652bec82bbd27"
    },
    {
      "start_firmware": "MentraLive_20260113",
      "end_firmware": "MentraLive_20260525",
      "url": "https://github.com/Mentra-Community/MentraOS/releases/download/asg-client/mtk_ota_update_20260113_20260525.zip",
      "sha256": "23b391985ee5da2ac3ec0adca2186e48bcaf5fa8ea50eb2f90a7145fe8d856a7"
    }
  ],
  "bes_firmware": {
    "version": "17.26.05.22",
    "url": "https://github.com/Mentra-Community/MentraOS/releases/download/asg-client/bes_firmware_17.26.05.22.bin",
    "sha256": "9499b1d132b934d3564c9dadf166595b4e9729cdf865b4be4e55b1d77cc5fa3c"
  }
}
```

PhilippeFerreiraDeSousa:
## Hardware verification results — phone-driven OTA (protocol above)

**Result: PASS** for the APK and MTK legs of the sequential update, end-to-end on real hardware, entirely through the new phone-driven path (`ota_start` + `ota_version_url` from the dev-settings override of #3147, manifest + APK served via ngrok). One pre-existing UI issue observed after the final reboot (details below); no OTA-flow defects found.

### Versions

| Component | Before | After |
|---|---|---|
| ASG client | `local.otabase` / 45715898 (branch build) | **`local.otatest` / 45716298** |
| MTK firmware | 20260513-era | **`MentraLive_20260525`** |
| BES firmware | 17.26.5.22 | unchanged — correct no-op (manifest `17.26.05.22` ≡ installed; comparator handles `05`≡`5`) |

### Timeline (single run, ~4.5 min check-to-reboot)

1. **T+0** — phone OTA check fetches the custom manifest (server log: `GET /local_live_version.json 200`); update prompt shown; accepted. Progress screen `displayState <init> → starting`.
2. **T+5s** — `ota_start` sent with `ota_version_url`; glasses ack; glasses fetch the manifest themselves, compute step sequence `sq:["apk","mtk"]` (BES correctly excluded — already current: glasses log `BES firmware is up to date (current: 17.26.5.22, server: 17.26.05.22)`).
3. **T+7s** — glasses begin APK download (`GET /asg-local-otatest.apk 200`, 84,497,296 bytes). Live session progress streams to the phone over BLE (session `sid 3968fc2d`); phone UI `starting → updating`.
4. **T+20s** — download complete (~13 s), `DownloadProgressEvent FINISHED`, SHA256 verified, install dispatched.
5. ASG client restarts on the new APK (pid 1974 → 2226) and **resumes the same OTA session from persistence** (`sid 3968fc2d`, now `cs:2, st:"mtk"`) — the critical restart-resume mechanic works, including the carried-over manifest URL.
6. MTK patch downloaded from its GitHub release URL, `MTK OTA info: payload check success`, staged; final session status delivered to the phone: `{"cs":2,"st":"mtk","phase":"install","sp":100,"op":100,"status":"complete"}`; phone UI `updating → complete`.
7. Glasses log `🔄 Rebooting now to apply staged MTK firmware update`; after reboot, `version_info_1/2/3` report the new build (`local.otatest` / 45716298), new MTK (`MentraLive_20260525`), and the compiled prod default in `ota_version_url` (as designed — the device never persists the custom URL).

### Scenario matrix status

| # | Scenario | Status |
|---|---|---|
| 1 | Full sequence APK → MTK → BES | **Pass** (APK + MTK; BES leg needs a genuinely older BES flash to exercise) |
| 2 | APK-only | Covered implicitly by iteration (rebuild → single APK leg); will confirm on next run |
| 3 | No-update (manifest == installed) | To confirm post-run (manifest now equals installed) |
| 4 | WiFi kill mid-download | Pending |
| 5 | Wrong sha256 → `apk_verify_failed` | Pending |
| 6 | Legacy gate (build < 100000) | **Pass, incidentally** — see below |

### Incidental findings

- **Legacy gate validated on real hardware**: the test initially started with factory glasses (versionCode 36 — flashing the old MTK image had replaced the branch ASG client, since the ASG client ships in the system image). The phone correctly ignored the custom-URL override and routed the check to the glasses' own (prod) manifest, exactly as the gate is designed to do. Test-setup lesson, now in the protocol: flash firmware **first**, install the branch ASG client **second**.
- **Known issue (pre-existing, not this PR): stuck "connecting" card after the MTK reboot.** After the glasses' final reboot, the BLE link silently re-established (heartbeats, `version_info`, `wifi_status` all flowing) but the `phone_ready`/`glasses_ready` + audio readiness handshake never re-ran, so the home-screen glasses card stayed on the connecting spinner indefinitely. The OTA session itself was unaffected (completion status was delivered before the reboot). Tapping Cancel + reconnecting (or restarting the app) recovers. Phone-driven OTA makes this path prominent — every MTK update ends with the user watching the app across a glasses reboot — so it deserves a follow-up ticket on the reconnect state machine rather than scope creep here.

### Log artifacts

Archived locally (`http-server.log`, `ngrok.log`, `phone-logcat-ota.log`, `glasses-logcat-part1/2.log`, `manifest-used.json`). Key excerpts:

```text
# HTTP server (glasses side of the flow)
17:59:52 "GET /local_live_version.json HTTP/1.1" 200
17:59:54 "GET /asg-local-otatest.apk HTTP/1.1" 200

# Glasses logcat
00:59:52 Version JSON parsed successfully. Root keys -> apps=true, mtk_patches=true, bes_firmware=true
00:59:52 BES firmware is up to date (current: 17.26.5.22, server: 17.26.05.22)
01:00:07 Download progress: 100% (84497296/84497296 bytes) ... FINISHED
01:03:59 MTK OTA info: payload check success
01:04:02 🔄 Rebooting now to apply staged MTK firmware update

# Phone (final session status from glasses, then UI)
{"sid":"3968fc2d","ts":2,"cs":2,"st":"mtk","sq":["apk","mtk"],"phase":"install","sp":100,"op":100,"status":"complete","type":"ota_status"}
21:03:59 [OTA_PROGRESS] displayState updating -> complete
21:04:44 Received version_info_1: {"app_version":"local.otatest","build_number":"45716298", ...}
```

PR #3125, opened 2026-06-10
Fix asg_client OTA APK cache verification (infinite update loop)

aisraelov:
## Problem

Some Mentra Live users get stuck in an **infinite OTA update loop** trying to update asg_client (e.g. 37→38). The update reports "complete," but the installed version never changes, so the next check re-detects the update and retries — forever. Symptoms from incident logs: install "FINISHED 100%" in ~3s with `bytesDownloaded:0`, then immediate re-check finding the same version, sometimes interleaved with `ota_start` watchdog timeouts and BLE `status=22` drops.

Notably, `adb install -r -d asg-client-38.apk` on an affected device fixes it **instantly** — proving the device, PackageManager, and the published APK are all fine. The failure is specific to the glasses' own OTA cache/install path.

## Root cause

Three defects, **all in the APK path only** (the MTK and BES firmware paths already verify SHA-256 correctly — which is why firmware OTAs work and only APK OTAs loop):

1. **`downloadApkInternal()` never checked the download completed.** A dropped connection makes `in.read()` return `-1` early without throwing, so the read loop exits cleanly and a **truncated APK is written and returned as a successful download**.

2. **`verifyApkFile()` was stubbed to `return true`** — APK integrity verification was disabled entirely, so the corrupt/partial file was never caught.

3. **`markCachedArtifactReady()` stored the *server's claimed* sha256** (from the manifest) as the cache's "verified" hash, not the file's actual computed hash. `isCachedArtifactValid()`'s fast path then compared `storedHash == expectedHash` — **both pulled from the same server field**, a tautology that always passes regardless of the bytes on disk. A bad file became permanently "valid" and was re-installed on every `ota_start`.

Together: a single flaky download poisons the cache with a corrupt/old APK that is then trusted forever and re-installed in a loop. (The state machine separately reports "complete" on process restart without re-checking the installed version, which is what surfaces as the loop on the phone — see follow-up note below.)

## Fixes (all in `OtaHelper.java`)

- **`downloadApkInternal()`**: reject downloads where `total != Content-Length` (incomplete), and verify the downloaded APK's SHA-256 before reporting success; delete the file and fail otherwise.
- **`verifyApkFile()`**: compute the file's real SHA-256 and compare against the server's expected hash (mirrors `verifyMtkFirmwareChecksum` / `verifyFirmwareFile`). Still no-ops gracefully if the manifest omits `sha256`.
- **`markCachedArtifactReady()`**: store the file's **computed** sha256, so cache validation is a genuine integrity check instead of comparing the server's claim to itself.
- **7-day cache TTL**: cached artifacts auto-expire and are re-downloaded even if otherwise valid, so a poisoned entry can't trap a device indefinitely. **This is also what lets already-poisoned devices self-heal** (see below).
- Add a shared `computeFileSha256()` helper.

## Recovery for already-stuck devices

Devices poisoned by the old code have a stored hash = the server's claim (not the file's real hash). After this lands, their fast-path check still matches (claim == claim), so they don't self-heal immediately on hash alone — **but the new 7-day TTL expires the entry and forces a fresh, now-verified download.** Any fresh download re-stores the computed hash, so from then on the integrity check is real. (For immediate recovery, a clean install over cable / `adb install -r -d` still works.)

## Scope / out of scope

- Did **not** add a post-install version-gate in `resumeFromSession()` (deferred by request). That's the remaining belt-and-suspenders item: the resume path reports "complete" on process restart without confirming `installedVersionCode == target`. Worth a follow-up.
- Flagged separately: a stray `refs/heads/fix` branch on the remote is blocking the entire `fix/*` branch namespace (this PR is branched as `asg-ota-apk-cache-verification-loop` as a workaround).

## Test evidence

- `./gradlew :app:compileDebugJavaWithJavac` — compiles clean (only pre-existing deprecation warnings).
- Verified the published artifacts manually: asg-client-37 and -38 share the same signing cert, monotonic versionCode, and valid zips; the only manifest delta is a harmless new debug receiver — confirming the loop is a cache/verification bug, not a bad artifact.
- Reproduced the loop condition on a device (downgraded to 37 via `adb install -r -d`) and confirmed the working install path goes systemui `cmd=install` → `copySuccess=true` → PackageManager replace; the bug manifests when the cached file handed to that path is corrupt/stale.

Manual on-device verification of the full download→verify→install path on a real Mentra Live is recommended before merge.

---

> [!NOTE]
> **Medium Risk**
> Changes core OTA download/install trust for APKs on device; behavior is safer but failed verification or TTL expiry will force re-downloads and could affect users mid-update until a good artifact is cached.
> 
> **Overview**
> Fixes **asg_client APK OTA** so corrupt or partial downloads cannot be cached and re-installed in a loop.
> 
> **Download path:** `downloadApkInternal()` now fails when bytes read do not match `Content-Length`, and runs **SHA-256 verification** (via restored `verifyApkFile()`) before success; bad files are deleted.
> 
> **Cache path:** `markCachedArtifactReady()` persists the **computed** file hash and a `hash_verified` flag instead of the manifest claim. `isCachedArtifactValid()` only uses the fast path when that flag is set and hashes match; legacy entries are **fully re-verified**. Adds a **7-day cache TTL** and clears `hash_verified` when wiping cache entries.
> 
> Shared **`computeFileSha256()`** backs cache marking and APK verification (aligned with MTK/BES behavior).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f65f3591ebadac7cac7ebc98c5e861f685e9c0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
@@ -136,6 +136,12 @@ public class OtaHelper {
     private static final String CACHE_KEY_MTK = "mtk_main";
     private static final String CACHE_KEY_BES = "bes_main";
 
+    // Cached artifacts auto-expire after this age and are re-downloaded on the next check, even
+    // if they still pass integrity verification. Defense-in-depth: a cache entry that somehow
+    // becomes poisoned (e.g. a future bug, or a server-side artifact swap that reuses a path)
+    // cannot persist indefinitely and trap a device in an update loop.
+    private static final long CACHE_MAX_AGE_MS = 7L * 24 * 60 * 60 * 1000; // 7 days
+
     // ⚠️ DEBUG FLAG: Set to true to skip all checks and install MTK firmware from local file
     // This will bypass version checking, downloading, and directly install
     // /storage/emulated/0/asg/mtk_firmware.zip
@@ -418,11 +424,27 @@ public class OtaHelper {
     private void markCachedArtifactReady(
             String cacheKey, String updateType, String localPath, JSONObject metadata) {
         try {
+            // Store the ACTUAL computed hash of the file on disk, not the server's claimed hash
+            // from the manifest. Storing the server's claim made cache validation a tautology
+            // (compare the claim to itself), so a corrupt/partial/stale file would be trusted
+            // forever. Computing the real hash here means isCachedArtifactValid()'s fast path
+            // genuinely verifies the bytes on disk match what the server expects.
+            String computedHash = computeFileSha256(localPath);
+            if (computedHash == null) {
+                Log.e(
+                        TAG,
+                        "Refusing to mark cache ready — could not hash file: "
+                                + localPath
+                                + " ("
+                                + cacheKey
+                                + ")");
+                return;
+            }
+
             SharedPreferences.Editor editor = getCachePrefs().edit();
             editor.putBoolean(cacheField(cacheKey, CACHE_FLAG_READY), true);
             editor.putString(cacheField(cacheKey, CACHE_FIELD_PATH), localPath);
-            editor.putString(
-                    cacheField(cacheKey, CACHE_FIELD_SHA256), metadata.optString("sha256", ""));
+            editor.putString(cacheField(cacheKey, CACHE_FIELD_SHA256), computedHash);
             editor.putString(
                     cacheField(cacheKey, CACHE_FIELD_VERSION),
                     metadata.optString("versionName", ""));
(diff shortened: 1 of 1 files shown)


aisraelov:
Good catch from both @chatgpt-codex-connector (P1) and @cubic-dev-ai (P2) — addressed in 9f65f35.

The concern was correct: legacy cache entries written by the old code stored the server's *claimed* sha256, so the new fast-path comparison `storedHash == expectedHash` was still claim-vs-claim for those entries, meaning an already-poisoned device wouldn't self-heal until the 7-day TTL expired.

**Fix:** added a per-entry `CACHE_FIELD_HASH_VERIFIED` marker that is only set when the stored hash was actually computed from the file (`markCachedArtifactReady`). The fast path now requires this flag — any legacy entry lacks it and is forced through the slow path, which runs the now-real `verifyApkFile()` re-hash. So:

- **Poisoned legacy entry** → re-hash ≠ server hash → cache invalid → fresh download, caught on the **next check** (no TTL wait).
- **Valid legacy entry** → re-hash matches → still trusted, no spurious failure.

Also clear the new field in `clearCachedArtifact()`. Compiles clean.

PhilippeFerreiraDeSousa:
image

PR #3137, opened 2026-06-11
[codex] Fix iOS Bluetooth-off disconnect state

PhilippeFerreiraDeSousa:
## Summary

Fixes the iOS Bluetooth SDK state path when phone Bluetooth is turned off while glasses are connected or after glasses were scanned.

## Root cause

The SDK checked Bluetooth availability before operations, but it did not globally react to CoreBluetooth state changes. If Bluetooth was disabled, the store could keep stale connected state, including `connectionState == CONNECTED`, and stale scan results even though the radio was unavailable.

## Changes

- Notify SDK listeners when `BluetoothAvailability` state changes.
- On unavailable Bluetooth states, cancel active scan sessions, clear scan/searching state, clear stale scan results, and disconnect connected glasses.
- Ensure `DeviceManager.disconnect()` clears both `connected` and `connectionState`.

## Validation

- Verified on Philippe's iPhone that turning Bluetooth off while connected emits `didUpdateGlasses connected=false connection=DISCONNECTED`.
- `xcodebuild -scheme MentraBluetoothSDK -destination 'generic/platform=iOS' CODE_SIGNING_ALLOWED=NO build`

---

> [!NOTE]
> **Medium Risk**
> Changes core connection and scan lifecycle plus auto-reconnect when Bluetooth is toggled; incorrect restore flags could surprise users, but scope is SDK state rather than security or data handling.
> 
> **Overview**
> Fixes stale iOS Bluetooth SDK state when the phone radio turns off while glasses are connected or a scan is in progress.
> 
> **`BluetoothAvailability`** now exposes **add/remove state listeners** that fire on CoreBluetooth `CBManagerState` updates (and immediately with the current state). **`MentraBluetoothSDK`** subscribes and, on powered-off/resetting/unauthorized/unsupported, **cancels active scan sessions**, **clears searching and scan results**, and **disconnects** connected glasses/controller while recording whether to **auto-reconnect** when Bluetooth comes back. Explicit **connect, disconnect, cancel, and forget** clear that restore intent so user-initiated disconnects are not revived.
> 
> **`DeviceManager.disconnect()`** now sets **`connectionState` to `DISCONNECTED`** alongside `connected = false`, so UI/delegates see a consistent disconnected state after radio loss or teardown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c754a22fff52a462ec1464eaef53d145019e029c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/modules/bluetooth-sdk/ios/Source/internal/BluetoothAvailability.swift
+++ b/mobile/modules/bluetooth-sdk/ios/Source/internal/BluetoothAvailability.swift
@@ -6,6 +6,7 @@ final class BluetoothAvailability: NSObject, CBCentralManagerDelegate {
 
     private var centralManager: CBCentralManager?
     private var state: CBManagerState = .unknown
+    private var listeners: [UUID: (CBManagerState) -> Void] = [:]
 
     override private init() {
         super.init()
@@ -19,6 +20,21 @@ final class BluetoothAvailability: NSObject, CBCentralManagerDelegate {
 
     func centralManagerDidUpdateState(_ central: CBCentralManager) {
         state = central.state
+        for listener in listeners.values {
+            listener(state)
+        }
+    }
+
+    @discardableResult
+    func addStateListener(_ listener: @escaping (CBManagerState) -> Void) -> UUID {
+        let id = UUID()
+        listeners[id] = listener
+        listener(state)
+        return id
+    }
+
+    func removeStateListener(_ id: UUID) {
+        listeners[id] = nil
     }
 
     func requirePoweredOn(operation: String) throws {
--- a/mobile/modules/bluetooth-sdk/ios/Source/internal/BluetoothAvailability.swift
+++ b/mobile/modules/bluetooth-sdk/ios/Source/internal/BluetoothAvailability.swift
@@ -20,7 +20,7 @@ final class BluetoothAvailability: NSObject, CBCentralManagerDelegate {
 
     func centralManagerDidUpdateState(_ central: CBCentralManager) {
         state = central.state
-        for listener in listeners.values {
+        for listener in Array(listeners.values) {
             listener(state)
         }
     }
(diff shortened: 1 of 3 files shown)


mobile/modules/bluetooth-sdk/ios/Source/internal/BluetoothAvailability.swift
```diff
@@ -19,6 +20,21 @@ final class BluetoothAvailability: NSObject, CBCentralManagerDelegate {
 
     func centralManagerDidUpdateState(_ central: CBCentralManager) {
         state = central.state
+        for listener in listeners.values {
+            listener(state)
+        }
+    }
+
+    @discardableResult
+    func addStateListener(_ listener: @escaping (CBManagerState) -> Void) -> UUID {
```
PhilippeFerreiraDeSousa, line 29: Codex :

> We need the new listener because apply only propagates state after something writes to DeviceStore.
> Bluetooth turning off is an external CoreBluetooth event. Before this PR, that event only lived inside BluetoothAvailability and was mostly used by requirePoweredOn(...) when starting a new operation. If the user turns Bluetooth off while the app is idle, no SDK API call happens, so nothing calls DeviceStore.shared.apply(...). The store subscribers never fire, so the app can keep stale connected and searchResults.
> The new listener is the missing ingress:
> CoreBluetooth state changed
> -> BluetoothAvailability listener
> -> SDK decides what state is invalid now
> -> DeviceStore.apply(...)
> -> existing store listeners / didUpdateGlasses / didUpdateScan fire
> -> apps update normally

PhilippeFerreiraDeSousa:
This has been verified on the iOS native example app :
- connect Mentra lives
- turn off bluetooth
- the example app now shows no glasses connected anymore

PR #3147, opened 2026-06-12
Add developer OTA manifest URL setting

PhilippeFerreiraDeSousa:
## Summary

Stacked on #3113. Ports the developer-settings OTA manifest URL parametrization from #2865 onto the `ota_version_url` plumbing that #3113 introduces, replacing the need for ad-hoc hardcoded URL changes when testing the OTA flow.

- New **Custom OTA Manifest URL** card in Developer Settings (**super mode only** — a wrong manifest can brick glasses), modeled on the Custom Backend URL card: free-text URL with **Save & Test** (fetches the URL and verifies it returns JSON before persisting), **Reset**, **Bookmark** with a tap-to-fill / long-press-to-remove saved list, and **Production** / **Staging** presets.
- The Staging preset is the canonical staging manifest from #3113: `https://mentra-live-ota-staging.pages.dev/staging_live_version.json`.
- New settings: `ota_version_url` (the override; `null` = no override) and `saved_ota_version_urls` (bookmarks). Both persist on-device only (`saveOnServer: false`).
- `getAsgOtaVersionUrl` resolves the override internally, so all five call sites (background check, check-for-updates, both progress screens' `ota_start` sends) pick it up without signature or dependency-array changes.
- OTA manifest URL constants move to `mobile/src/config/ota.ts`.

## This card is the staging opt-in (not just a dev convenience)

The TestFlight / Play internal binaries built by the staging workflow are the exact binaries later promoted to public beta and the app stores, so #3113 no longer bakes `EXPO_PUBLIC_ASG_OTA_VERSION_URL` into them — a compiled-in staging manifest URL would have shipped to customers. The **Staging** preset in this card is how staging testers point their phone's glasses-OTA checks at the staging manifest.

## Precedence

For glasses that honor `ota_start.ota_version_url` (wall-clock builds ≥ 100000):

1. `ota_version_url` developer setting (this PR; only resolved while super mode is enabled — a saved override is inert otherwise)
2. `EXPO_PUBLIC_ASG_OTA_VERSION_URL` (no longer set by CI; available for local dev via `mobile/.env`)
3. Glasses-reported URL, then production default

Legacy glasses ignore the URL in `ota_start` and install from their compiled default, so the override deliberately does not apply to them — same reasoning as the legacy gate in #3113 (a custom-manifest prompt would dead-end). The card's description notes this.

## Differences from #2865

#2865's ASG/BLE plumbing (`version_json_url` key, `startOtaFromPhone(String)`, native `sendOtaStart(url)`) is superseded by #3113's `ota_version_url` implementation, so only the mobile dev-settings layer is ported. The setting also no longer defaults to the production URL: it defaults to `null` ("no override") so that #3113's device-URL/production selection still applies to non-developers, and the override only wins when explicitly set.

## Test plan

- Used for the end-to-end OTA verification of #3113: glasses on old MTK/BES + branch ASG client, phone dev build pointing the override at a locally served manifest (`http://<lan-ip>:8000/...`) whose `apkUrl` references a freshly built wall-clock ASG APK; verifying the sequential APK → MTK → BES flow.
- Settings card verified via Metro hot reload (JS-only change; the native `sendOtaStart(url)` parameter already exists in #3113 dev builds).

---

> [!NOTE]
> **High Risk**
> A bad manifest URL can brick glasses firmware; the UI is super-mode gated and requires a successful manifest fetch before save, but the override still changes what every OTA path installs from.
> 
> **Overview**
> Adds a **Custom OTA Manifest URL** card to Developer Settings, shown only when **super mode** is on, so testers can point glasses OTA checks at a custom manifest without hardcoding URLs in the app.
> 
> The new `OtaVersionUrl` UI mirrors the backend URL card: **Save & Test** (GET + JSON validation before persisting), **Reset**, bookmarks, and **Production** / **Staging** presets from `mobile/src/config/ota.ts`. Settings `ota_version_url` (null = no override) and `saved_ota_version_urls` persist on-device only.
> 
> `getAsgOtaVersionUrl` now applies the saved override when super mode is enabled (ahead of `EXPO_PUBLIC_ASG_OTA_VERSION_URL`, glasses-reported URL, and production). The override is ignored for legacy ASG builds that do not honor `ota_start.ota_version_url` (build &lt; 100000), matching existing legacy behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c3a01bb5bb20c91875929aff4203ec6cbbced8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/src/app/miniapps/settings/developer.tsx
+++ b/mobile/src/app/miniapps/settings/developer.tsx
@@ -3,6 +3,7 @@ import {useEffect, useRef, useState} from "react"
 import {ScrollView, View} from "react-native"
 
 import BackendUrl from "@/components/dev/BackendUrl"
+import OtaVersionUrl from "@/components/dev/OtaVersionUrl"
 import StoreUrl from "@/components/dev/StoreUrl"
 import {Header, Icon, Screen, Text} from "@/components/ignite"
 import SelectSetting from "@/components/settings/SelectSetting"
@@ -287,6 +288,8 @@ export default function DeveloperSettingsScreen() {
           <BackendUrl />
 
           <StoreUrl />
+
+          <OtaVersionUrl />
 
           {superMode && <RouteButton label="Super Settings" onPress={() => push("/miniapps/settings/super")} />}
 
--- a/mobile/src/app/miniapps/settings/developer.tsx
+++ b/mobile/src/app/miniapps/settings/developer.tsx
@@ -289,7 +289,8 @@ export default function DeveloperSettingsScreen() {
 
           <StoreUrl />
 
-          <OtaVersionUrl />
+          {/* Super mode only: a wrong OTA manifest can brick glasses */}
+          {superMode && <OtaVersionUrl />}
 
           {superMode && <RouteButton label="Super Settings" onPress={() => push("/miniapps/settings/super")} />}
 
--- a/mobile/src/components/dev/OtaVersionUrl.tsx
+++ b/mobile/src/components/dev/OtaVersionUrl.tsx
@@ -149,8 +149,9 @@ export default function OtaVersionUrl() {
       <View style={themed($textContainer)}>
         <Text style={themed($label)}>Custom OTA Manifest URL</Text>
         <Text style={themed($subtitle)}>
-          Override the manifest used for glasses OTA checks and installs. Leave blank to use the default (staging builds
-          use the staging manifest). Legacy glasses firmware ignores the override.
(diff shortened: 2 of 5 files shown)


PhilippeFerreiraDeSousa:
Last commit https://github.com/Mentra-Community/MentraOS/pull/3147/changes/0c3a01bb5bb20c91875929aff4203ec6cbbced8b moves the ota manifest setting behind supermode on @nic-olo 's input

PR #3150, opened 2026-06-14
Fail safe when BES firmware version cache is stale across boots

PhilippeFerreiraDeSousa:
## Problem

The ASG client decides whether a BES firmware update is needed using a version cached in `AsgSettings` (`mcu_firmware_version` in SharedPreferences), refreshed from the chip via the `syvr` UART handshake at boot. On real hardware (Mentra Live, 2026-06-12 verification of #3113) this cache went **stale across multiple reboots**: it read `17.26.5.22` while the BES chip actually ran `17.26.1.13`.

Root cause from logcat: the cached (stale) version is sent to the phone *before* the refresh request goes out, and on the affected boot **no `syvr` response ever arrived**, so the cache never refreshed. Result: both the glasses-side check (`OtaHelper.checkBesUpdate`) and the phone-side check (fed by `version_info_3`) reported "BES up to date" and skipped a genuinely needed update — a device in this state silently never receives BES updates.

## Changes

**1. Stale cache fails safe — boot-stamped freshness** (`AsgSettings.java`)
- `setMcuFirmwareVersion` stamps the value with a per-boot token (`currentTimeMillis - elapsedRealtime` — stable within a boot, jumps across reboots).
- `getMcuFirmwareVersion` returns `""` (unknown) when the stored stamp predates the current boot. Both `checkBesUpdate` (glasses) and the phone's `checkBesUpdate` already treat unknown as "update needed", so an unrefreshed value now fails safe (worst case: re-flash the same version) instead of being trusted.
- Adds `hasFreshBesFirmwareVersion()` (retry stop-condition) and `getCachedBesFirmwareVersionRaw()` (telemetry only).

**2. Reliable refresh — retry + both command variants** (`K900BluetoothManager.java`)
- `requestBesSystemVersion()` now retries (5 attempts, 2s apart), stopping once the cache is fresh this boot and logging an error if it never is.
- Each attempt sends **both** `cs_syvr` (→ `sr_syvr`) and `sh_syvr` (→ `hs_syvr`/`BesVersionEvent`), hedging against firmware that only honors one variant.
- `cacheBesFirmwareVersion` routes through `AsgSettings` (so the boot stamp is recorded — a bare prefs write left it stale) and cancels pending retries on success.

**3. Telemetry breadcrumb** (`OtaHelper.java`)
- All four OTA-check read sites go through a new `getBesVersionForOtaCheck(context)`, which logs a warning and emits a Sentry/Console breadcrumb (`ota` / `bes_version_unrefreshed`) when a check runs against a non-empty-but-unrefreshed cache.

`version_info_3` to the phone is handled automatically — it reads through the same getter, so it carries `""` until the chip is confirmed, then `BesVersionEventSubscriber` re-sends the real value once the response lands.

## Tradeoff

On a device where `syvr` *never* succeeds even with retries, every boot reads unknown → offers the BES update → re-flashes the same version each boot. That's the accepted "worst case re-flashes the same version", and strictly better than silently never updating. If the handshake proves permanently broken on some hardware, a phone-side "don't re-offer the version we already flashed this boot" guard would be the follow-up.

## Testing

- `./gradlew :app:compileDebugJavaWithJavac` — BUILD SUCCESSFUL.
- Hardware (to run before merge): downgrade BES via vendor tool, reboot, confirm the cache reflects the chip without manual intervention and the OTA check offers the BES update. If the handshake fails all retries, confirm the `bes_version_unrefreshed` breadcrumb fires and the check still offers the update.

---

> [!NOTE]
> **Medium Risk**
> Changes BES OTA eligibility and version reporting to phone; wrong boot-stamp logic could spuriously offer updates, but behavior intentionally fails safe when the handshake never succeeds.
> 
> **Overview**
> Fixes a case where **BES firmware version stayed cached across reboots** without a fresh chip read, so OTA and the phone could treat the device as up to date when it was not.
> 
> **`AsgSettings`** now stamps the cached BES version with a per-boot token on write and returns **empty (unknown)** on read when the stamp does not match the current boot. That makes unconfirmed cache values fail safe toward offering a BES update. Adds **`hasFreshBesFirmwareVersion()`** for handshake completion and **`getCachedBesFirmwareVersionRaw()`** for telemetry only.
> 
> **`K900BluetoothManager`** replaces a one-shot `cs_syvr` with a **retry loop** (5 attempts, 2s apart), sends **`cs_syvr` and `sh_syvr`**, stops when the cache is fresh, and caches via **`AsgSettings`** so the boot stamp is set.
> 
> **`OtaHelper`** routes all BES OTA version reads through **`getBesVersionForOtaCheck()`**, which logs and reports **`bes_version_unrefreshed`** when a non-empty but stale raw cache is seen. Phone **`version_info_3`** uses the same getter, so it reports unknown until refresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11c61070c0b6cb4e18b823d2c33b7c723d01a6d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
@@ -12,6 +12,7 @@ import com.mentra.asg_client.io.bluetooth.utils.DebugNotificationManager;
 import com.mentra.asg_client.logging.BleTraceLogger;
 import com.mentra.asg_client.reporting.domains.BluetoothReporting;
 import com.mentra.asg_client.service.core.AsgClientService;
+import com.mentra.asg_client.settings.AsgSettings;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -56,6 +57,15 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
     private static final int PACING_DELAY_MS =
             75; // Delay between successful packets - BES2700 needs time to drain BLE TX
     private int consecutiveFailures = 0;
+
+    // BES system-version (syvr) handshake retry: the chip occasionally drops the first request
+    // after a SOC boot, leaving the cached version stale. Re-request until the chip answers (the
+    // cache becomes fresh for this boot) or we exhaust the attempts.
+    private static final int BES_SYVR_MAX_ATTEMPTS = 5;
+    private static final long BES_SYVR_RETRY_DELAY_MS = 2000L;
+    private final android.os.Handler besVersionHandler =
+            new android.os.Handler(android.os.Looper.getMainLooper());
+    private int besSyvrAttempts = 0;
 
     // Inner class to track file transfer state
     private static class FileTransferSession {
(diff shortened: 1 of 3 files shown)


PhilippeFerreiraDeSousa:
From @aisraelov :
The day 1 firmware I downgraded to was not yet sending the bes version so this is why. Also it is weird that I have a stale bes version in prefs because flashing an mtk SOC image was expected to erase that.
From me : 
This cannot affect customer because they never downgrade so they cannot have a stale bes version in prefs that is more recent and block an update from those old bes versions

PR #3195, opened 2026-06-18
Return partial Wi-Fi scan results on timeout

PhilippeFerreiraDeSousa:
## Summary
- Track the latest non-empty accumulated Wi-Fi scan results while `requestWifiScan()` is pending.
- If the request times out because old glasses never send `scan_complete: true`, resolve with the accumulated scan list instead of rejecting.
- Keep the normal completion behavior unchanged, and still reject the timeout when no scan results ever arrived.
- Document the compatibility fallback in the Bluetooth SDK README.

## Validation
- `git diff --check`
- `xcodebuild -scheme MentraBluetoothSDK -destination 'generic/platform=iOS' build`

Note: `swift build` without an iOS destination still fails because the SDK imports `UIKit`; the iOS `xcodebuild` path succeeds.

---

> [!NOTE]
> **Medium Risk**
> Wi-Fi scan timeout behavior changes from reject to success when partial data exists, which can affect apps that relied on timeout as failure; miniapp WebView loading on Android changes loading UX.
> 
> **Overview**
> **`requestWifiScan()`** on Android and iOS now keeps the latest non-empty `wifi_scan_result` list while a scan is pending. If the request times out because legacy glasses never send `scanComplete: true`, the promise **resolves with that accumulated list** instead of rejecting; timeouts with no results still reject. Normal completion when `scan_complete` arrives is unchanged. The Bluetooth SDK README documents this fallback.
> 
> **`LocalMiniappView`** drops the Android-only hidden WebView + splash gate and always mounts a single full-size WebView (with the existing load gate and ready handshake).
> 
> The public JS **`BluetoothSdk`** export no longer exposes **`setCaptureLedEnabled`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae54b0a8d4c2fd4062ecbb7746e74f040290f715. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aisraelov (approved):
Please test on iOS+Android first before merging but this is a good idea to implement

PhilippeFerreiraDeSousa:
Tested scanning in the example app 0.1.14, against current asg client and last prod asg client

PR #3213, opened 2026-06-20
Add glasses facade read models

PhilippeFerreiraDeSousa:
## Summary
- Expand `toolkit.glasses` read models for the fields MentraOS still reads through the raw glasses-store shim: connection/ready flags, device diagnostics, controller status, hotspot status, richer device info.
- Add `toolkit.glasses.wifi.snapshot()` for the `wifiStatusKnown` use case and `toolkit.ota.progress()` / `onProgress()` for current OTA progress screens.
- Move feedback/incident diagnostics off direct `useGlassesStore` reads and onto `toolkit.glasses.{status,info,wifi,diagnostics}`.
- Restore the #3206 island barrel export to use `@mentra/bluetooth-sdk/internal` instead of reaching into `bluetooth-sdk/build/_internal`.

## Recommendation
This keeps the island store intact for now, but gives MentraOS/OEM hosts stable facade APIs to migrate UI and diagnostics off the shim. Remaining raw-store callers should move in follow-ups by domain:
- UI status/cards: `toolkit.glasses.status()` / `info()` / `wifi.snapshot()` / `controller.status()`
- OTA screens: `toolkit.ota.status()` / `progress()` now, then a later `toolkit.ota.check()` orchestration API
- Pairing/loading waits: a future `toolkit.glasses.waitForStatus(...)`
- Incident payloads: `toolkit.glasses.diagnostics()`

## Validation
- `cd mobile && bun run test -- src/__tests__/glassesFacade.test.ts src/__tests__/glassesWifi.test.ts src/__tests__/otaFacade.test.ts`
- `cd mobile && bun run compile`

## Notes
- ESLint was attempted on the changed files but the repo config currently fails before linting with missing package `globals` from `eslint.config.mjs`.
- This PR is stacked on `aisraelov/island-namespace-wifi` to keep the diff scoped to the facade work.


mobile/modules/island/src/facades/ota.ts
```diff
@@ -49,4 +53,15 @@ export const ota = {
       cb(status)
     })
   },
+
+  /** Subscribe to legacy OTA progress changes. Returns an unsubscribe. */
```
PhilippeFerreiraDeSousa, line 57: awkward

PhilippeFerreiraDeSousa:
So we don't want that, incident report needs to be moved into the toolkit and onto cloud v2, instead of exposing the whole state read-only

PR #3214, opened 2026-06-20
[codex] Fix K900 UART frame parsing

PhilippeFerreiraDeSousa:
## Summary

- Make the Mentra Live BES/K900 UART parser use declared frame lengths instead of scanning for the next `$$` delimiter.
- Support both existing string-frame length byte orders and preserve file-frame parsing by using the file packet size header.
- Add focused parser tests for concatenated frames, split frames, payloads containing `$$`, and a truncated Wi-Fi scan frame followed by another K900 frame.

## Root Cause

`BesMessageParser` treated `$$` as the frame boundary. When a malformed or truncated K900 frame was followed by another frame, the parser could join bytes across the boundary and forward a corrupted JSON payload such as `request_wifi_scan` followed by the next `##0...{"C":` frame start. That caused `CommandProcessor` to reject the command before Wi-Fi dispatch, so the phone-side Wi-Fi scan waited until timeout.

## Validation

- `./gradlew :app:testDebugUnitTest --tests 'com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.BesMessageParserTest'`
- `git diff --check -- asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/BesMessageParser.java asg_client/app/src/test/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/BesMessageParserTest.java`

Note: repo-wide `./gradlew :app:spotlessCheck` is blocked in this worktree by pre-existing formatting failure in `asg_client/app/src/main/java/com/mentra/asg_client/io/media/managers/MediaUploadQueueManager.java` after Spotless skipped its ratchet because this checkout does not expose a `.git` directory to Spotless.

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/BesMessageParser.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/BesMessageParser.java
@@ -24,16 +24,6 @@ public class BesMessageParser {
     // Buffer size for parsing messages
     private static final int BUFFER_SIZE = 8192; // 8KB buffer
     private static final int STRING_FRAME_OVERHEAD = 7; // ## + type + length + payload + $$
-    private static final int FILE_FRAME_OVERHEAD =
-            BesWireFormat.LENGTH_FILE_START
-                    + BesWireFormat.LENGTH_FILE_TYPE
-                    + BesWireFormat.LENGTH_FILE_PACKSIZE
-                    + BesWireFormat.LENGTH_FILE_PACKINDEX
-                    + BesWireFormat.LENGTH_FILE_SIZE
-                    + BesWireFormat.LENGTH_FILE_NAME
-                    + BesWireFormat.LENGTH_FILE_FLAG
-                    + BesWireFormat.LENGTH_FILE_VERIFY
-                    + BesWireFormat.LENGTH_FILE_END;
     private static final int FRAME_INCOMPLETE = -1;
     private static final int FRAME_INVALID = -2;
 
@@ -176,12 +166,10 @@ public class BesMessageParser {
             return getExpectedStringFrameLength(buffer, start, available);
         }
 
-        if (isFileCommandType(commandType)) {
-            return getExpectedFileFrameLength(buffer, start, available);
-        }
-
-        Log.w(TAG, "Unknown K900 command type: 0x" + String.format("%02X", commandType));
-        return FRAME_INVALID;
+        // The observed Wi-Fi failure was on string/JSON frames. Keep non-string
+        // commands on the previous delimiter parser until their frame variants have
+        // dedicated evidence and tests.
+        return getDelimitedFrameLength(buffer, start, available);
     }
 
     private int getExpectedStringFrameLength(byte[] buffer, int start, int available) {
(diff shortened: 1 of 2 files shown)


PhilippeFerreiraDeSousa:
The app path is local/device-only:
[scan.tsx (line 79)](/Users/philippe/dev/MentraOS-dev/mobile/src/app/wifi/scan.tsx:79) → `toolkit.glasses.wifi.scan()` → [glassesWifi.ts (line 23)](/Users/philippe/dev/MentraOS-dev/mobile/modules/island/src/facades/glassesWifi.ts:23) → `BluetoothSdk.requestWifiScan()` → [MentraLive.kt (line 4559)](/Users/philippe/dev/MentraOS-dev/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt:4559) sends `{"type":"request_wifi_scan"}` over BLE.
The glasses log shows the request arriving corrupted before command dispatch:
```
Unterminated object at character 55 of {"C":"{\"type\":\"request_wifi_scan\",\"mId\":3##09...{"C":
```
That happens twice around 06-20 03:17:16 and 06-20 03:17:19. A later ping also gets corrupted similarly. So the timeout is because the glasses never successfully parses the scan command, not because scan depends on cloud v1. This looks like a BLE/K900 framing or packet assembly bug: the ASG parser scans for `$$` frame endings, then extracts payload by K900 length; the failing payload contains the beginning of another `##...` frame inside the JSON.

PhilippeFerreiraDeSousa:
Codex evidence audit on the Wi-Fi scan failure:

We have more ground-truth data than just the phone-side timeout, but less than the first PR summary implied.

Direct observation from ADB/logcat: the `request_wifi_scan` command reached the ASG through the K900 UART path, but the extracted payload was already corrupted before Wi-Fi dispatch:

```text
06-20 03:17:16.941 I MentraBleTrace: ... type=raw bytes=116 payload=<non-json payload>
06-20 03:17:16.942 E K900BluetoothManager: org.json.JSONException: Unterminated object at character 55 of {"C":"{\"type\":\"request_wifi_scan\",\"mId\":3##09��{"C":
06-20 03:17:16.943 E JsonParser: JSON parsing error for direct JSON: Unterminated object at character 55 of {"C":"{\"type\":\"request_wifi_scan\",\"mId\":3##09��{"C":
06-20 03:17:16.943 W CommandProcessor: Failed to parse JSON from byte data
```

The same corrupted `request_wifi_scan` parse failure repeated at `06-20 03:17:19.076`.

The important clue is the embedded `##09...{"C":` inside the JSON string. That looks like bytes from another K900 frame were merged into the scan command payload, so the command never got to the Wi-Fi handler as valid JSON.

What is inferred from code/tests: the old parser framed messages by searching for the next `$$` delimiter even though K900 string frames carry an explicit length. That can merge bytes across frames when a frame is truncated/malformed. The new test models this as a truncated `request_wifi_scan` frame followed by another valid frame and verifies parser resync.

What we do not currently have: a preserved raw byte dump of the malformed UART frame, or a saved corrupted ping frame. The ping case is a synthetic resync test, not direct observation. PR wording should avoid claiming we observed a corrupted ping unless someone has another artifact.

Post-fix/current-device observation: a later scan round-trip succeeded on-device:

```text
04:10:20.750 request_wifi_scan mId=61 routed successfully
04:10:23.584-04:10:24.662 wifi_scan_result chunks emitted
04:10:24.886 final wifi_scan_result scan_complete=true
```

My read: the fix is at the right layer because Wi-Fi never received a valid command during the failure. The one thing to watch is blast radius: this patch improves the shared K900 parser, including file-frame handling, while the observed failure only proves the JSON/string-frame path. If we keep the broader parser fix, the PR description should frame the evidence as “observed corrupted `request_wifi_scan` K900 frame; tests cover length-based resync,” not as direct proof of every parser edge case.

PhilippeFerreiraDeSousa:
Codex follow-up: narrowed this PR per the evidence audit.

The new commit `2b23482d2` removes the untested file-frame length parser path. The parser now applies declared-length framing/resync only to K900 string/JSON frames (`CMD_TYPE_STRING`), which is the path directly implicated by the corrupted `request_wifi_scan` observation. Non-string K900 command types stay on the prior delimiter-based parsing behavior until we have separate evidence/tests for those frame variants.

Validation after narrowing:

```bash
./gradlew :app:testDebugUnitTest --tests 'com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.BesMessageParserTest'
./scripts/check-android-compile.sh asg
```

Both passed.

PhilippeFerreiraDeSousa:
I think the BES is doing some garbling, and we need instrumentation first.

PhilippeFerreiraDeSousa:
This is not addressing the root cause

PR #3228, opened 2026-06-22
Fix K900 phone frame length encoding

PhilippeFerreiraDeSousa:
## Summary

- Encode phone-to-K900 BES2700 frame lengths as big-endian on both Android and iOS by reusing the canonical protocol packer.
- Update comments that still described phone-to-device K900 frames as little-endian.
- Add Android unit coverage for the 256-byte payload boundary, where the old byte order produced the problematic `00 01` length header.
- Fix stale imports in an existing Bluetooth SDK unit test so the focused unit-test task can compile.

## Why

While investigating K900 transport drops, I used a temporary iOS stress harness that sent 300 K900 frames alternating total packet lengths `262`, `263`, and `264` bytes. Before this fix, every `263`-byte packet had a 256-byte payload and was queued by iOS with this header:

```text
23 23 30 00 01 ...
```

The BES command parser reads the two length bytes as big-endian, so `00 01` is interpreted as a one-byte payload. That makes BES try to parse only the opening `{`, fail JSON parsing, and skip forwarding the command to ASG over UART. The surrounding `262` and `264` byte frames can appear to work, which made the failure look transport-flaky until we hit the exact 256-byte boundary.

With this change, a 256-byte payload is encoded as:

```text
23 23 30 01 00 ...
```

That matches the existing `packDataCommand` implementation and the BES-side parser expectation.

## Reproduction Notes

1. From the old code, send K900 JSON frames whose wrapped payload length is exactly 256 bytes. The temporary stress harness used total packet lengths `262`, `263`, and `264` to make this easy to see.
2. Watch the phone-side BLE queue logs. The broken case emits `23 23 30 00 01` for the 263-byte total packet.
3. Watch ASG UART/raw frame logs. In the failing run, the `263` total-length frames did not appear on ASG even though the iOS BLE writes completed with `status=ok`.
4. Rebuild with this patch and repeat. The 256-byte payload header should be `23 23 30 01 00`.

## Validation

- `cd mobile/android && ./gradlew :mentra-bluetooth-sdk:testDebugUnitTest --tests com.mentra.bluetoothsdk.utils.K900ProtocolUtilsTest`

---

> [!NOTE]
> **Medium Risk**
> Changes device command framing on a critical BLE transport path; behavior is narrowly scoped to encoding but wrong frames previously caused silent command loss.
> 
> **Overview**
> **Phone-to-K900 BES2700 frames** no longer use a separate little-endian length encoder. Android `packDataToK900` and iOS `packDataToK900` now delegate to the existing **`packDataCommand`** path so the 2-byte length matches what the BES parser expects (big-endian).
> 
> That fixes a **256-byte payload boundary** bug: little-endian encoding produced `00 01` in the header, which BES read as length 1, broke JSON parsing, and dropped commands on UART while shorter/longer frames could still look fine.
> 
> Adds **`K900ProtocolUtilsTest`** asserting `01 00` for a 256-byte payload. **`PhotoRequestTest`** imports are corrected so unit tests compile.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b38f6359315fa88d9a4e53b48affc57cde5a65c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/utils/K900ProtocolUtils.java
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/utils/K900ProtocolUtils.java
@@ -120,49 +120,21 @@ public class K900ProtocolUtils {
     }
 
     /**
-     * Pack raw byte data with K900 BES2700 protocol format for phone-to-device communication
-     * Format: ## + command_type + length(2bytes) + data + $$
-     * Uses little-endian byte order for length field
+     * Pack raw byte data with K900 BES2700 protocol format for phone-to-device communication.
+     * The BES command parser expects the 2-byte length field in big-endian order.
      *
      * @param data The raw data to pack
      * @param cmdType The command type (use CMD_TYPE_STRING for JSON)
      * @return Byte array with packed data according to protocol format
      */
     public static byte[] packDataToK900(byte[] data, byte cmdType) {
-        if (data == null) {
-            return null;
-        }
-
-        int dataLength = data.length;
-
-        // Command structure: ## + type + length(2 bytes) + data + $$
-        byte[] result = new byte[dataLength + 7]; // 2(start) + 1(type) + 2(length) + data + 2(end)
-
-        // Start code ##
-        result[0] = CMD_START_CODE[0]; // #
-        result[1] = CMD_START_CODE[1]; // #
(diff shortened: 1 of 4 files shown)


PhilippeFerreiraDeSousa:
I'll be using that commit for a bit locally to verify

PR #3231, opened 2026-06-23
[codex] Queue ASG BLE sends to unblock photo upload

PhilippeFerreiraDeSousa:
image

Asg client log trace left, Bluetooth SDK log trace right.
New asg client log trace has warnings for queuing delays

## Summary

This fixes the latency problem found in the photo capture path: ASG-to-phone BLE sends were chunked synchronously, and the caller was forced to sit inside the BLE chunk/pacing loop before any later work could continue.

For photo capture, that meant `MediaCaptureService` could finish capture, call `sendPhotoStatus("captured", captureMetadata)`, and then remain blocked while K900 BLE chunking sent the large metadata payload chunk-by-chunk with pacing sleeps. Only after that synchronous send returned would the code proceed to start the Wi-Fi upload. In practice, the upload path was waiting below BLE status delivery.

This PR moves that synchronous work out of caller threads by making `BaseBluetoothManager.sendMessage(...)` enqueue outbound BLE payloads onto one ordered background sender. The queue preserves ASG-to-phone message ordering, but app/camera/service handlers no longer wait for BLE chunk pacing to finish.

## Why This Matters

The expensive operation was not the photo upload. A measurable portion of pre-upload latency came from the ASG client synchronously serializing and pacing BLE status chunks before upload was allowed to start.

The worst offender in the measured flow was the `captured` photo status because it carries capture metadata and becomes an 8-chunk BLE message. Before this change, the code path effectively did:

1. capture photo
2. synchronously send `captured` over BLE, including chunk pacing
3. only then start webhook upload

After this change, the path becomes:

1. capture photo
2. enqueue `captured` for ordered BLE delivery
3. immediately continue to upload while the BLE worker drains chunks in order

The phone still receives `captured` before `uploading`, because both messages go through the same ordered outbound queue. The important change is that ASG logic is no longer blocked by the BLE transport timing.

## Implementation

- `BaseBluetoothManager.sendMessage(...)` now copies the payload, records trace metadata, enqueues the payload on a single-thread outbound executor, and returns once accepted for delivery.
- The existing subclass send implementation still performs the actual BLE/K900 send, but now from the ordered worker thread.
- The queue is global at the ASG Bluetooth transport edge, so this applies to all ASG-to-phone events from all handlers, not just photo-specific messages.
- `IBluetoothManager.sendMessage(...)` docs now reflect the actual contract: accepted for outbound delivery rather than synchronously sent to completion.

## Measured Behavior

Using the React Native example app with a local media server and these photo options:

- `size=max`
- `compress=none`
- `save=false`
- `sound=true`
- `aeExposureDivisor=3`

Before this PR, the ASG-side timeline showed upload waiting until the `captured` BLE message finished chunking:

- `captured` BLE output started at `16:12:37.542`
- final `captured` chunk finished at `16:12:38.113`
- Wi-Fi upload started at `16:12:38.135`

So upload started only after the 8-chunk `captured` status drained.

After this PR, the ASG-side timeline showed upload starting while the `captured` BLE chunks were still draining:

- `captured` status was queued at `16:40:42.975`
- Wi-Fi upload started at `16:40:42.996`
- final `captured` chunk finished later at `16:40:43.616`

This confirms the blocking wait was removed. In that run, ASG command-to-upload-start improved from `2788ms` to `2253ms`, about `535ms` faster before upload. The phone-observed command-to-`photo_response` improved from `5573ms` to `4396ms`, about `1.18s`, but about `0.59s` of that end-to-end difference was due to the Wi-Fi upload itself being faster in the second run (`1457ms` vs `869ms`). The pre-upload improvement attributable to this change is about `0.5-0.6s` in the measured run.

## Conditional Warning Diagnostics

The trace work is supporting instrumentation, not the main fix. It is intentionally quiet during normal flow.

The low-level diagnostic layers now gather timing and queue metadata first, then emit only when something looks backed up or failed. Emitted diagnostics are marked with `level=warning` and a `warningReason`, and `scripts/ble-trace-pretty.mjs` colors those warning summaries yellow.

Current warning conditions include:

- outbound ASG queue delay >= `250ms`
- outbound ASG send duration >= `250ms`
- outbound ASG queue depth >= `5`
- ASG outgoing chunked-message duration >= `250ms`
- ASG outgoing chunk send/spacing delays above threshold
- ASG incoming chunk reassembly duration >= `250ms`
- phone-side SDK BLE write queue/callback/rate-limit delays >= `250ms`
- send failures, rejected writes, malformed chunks, and parse errors

This keeps PLA usable: high-level photo/status/upload trace lines still appear normally, while noisy chunk/write/queue internals only appear as yellow warnings when they explain real latency or failures.

## Notes / Caveats

- Reliable-message ACK timers now begin after the queued transport send attempt completes, so outbound queue delay no longer eats into the ACK timeout window.
- Because delivery is globally ordered through one worker, this avoids handler-level blocking without allowing ASG-to-phone events to reorder.
- If a BLE file-start wait times out after the worker task has already begun, `sendFile(...)` now waits for the actual start result instead of returning `false` while the transfer may continue in the background.

## Validation

Ran after rebasing onto current `origin/dev`:

- `git diff --check`
- `node --check scripts/ble-trace-pretty.mjs`
- Google Java Format dry-run on touched ASG Java files
- `scripts/check-android-compile.sh`

Device/log validation:

- Rebuilt and reflashed the ASG client.
- Confirmed the new queued outbound sender was running by observing `asg_ble_outbound_queue` traces in `logap2.txt`.
- Confirmed upload starts before the `captured` BLE chunk sequence finishes.

---

> [!NOTE]
> **Medium Risk**
> Changes ordering and timing across all ASG-to-phone BLE traffic and reliable-message ACK/retry behavior, not only photo paths; regressions could affect message delivery or duplicate sends if the gate/queue semantics are wrong.
> 
> **Overview**
> **Outbound BLE on ASG no longer blocks camera/service threads.** `BaseBluetoothManager.sendMessage` now enqueues copies on a single-thread executor and returns when accepted; K900 chunk pacing and UART writes run on that worker. `sendFile` is routed through the same queue with timeout handling if start already began on the worker.
> 
> **Reliable delivery is aligned with the queue.** `IBluetoothManager` adds optional send-completion callbacks and a worker-side `SendMessageGate`. `ReliableMessageManager` starts ACK timeouts only after the queued send finishes and can skip stale retries via the gate; `CommunicationManager` and `ResponseSender` wire this through.
> 
> **Diagnostics only surface problems.** Conditional `level=warning` traces were added for outbound queue depth/delay, K900 chunk timing, inbound chunk reassembly, and phone SDK BLE write/queue delays; `ble-trace-pretty.mjs` highlights those lines in yellow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1937549936258c84cb1175ae0484217862394c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/core/BaseBluetoothManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/core/BaseBluetoothManager.java
@@ -586,9 +586,12 @@ public abstract class BaseBluetoothManager implements ICompanionTransport {
         try {
             return future.get(timeoutMs, TimeUnit.MILLISECONDS);
         } catch (TimeoutException e) {
-            future.cancel(true);
-            Log.e(TAG, "Timed out waiting for outbound BLE file transfer start result", e);
-            return false;
+            Log.w(
+                    TAG,
+                    "Timed out waiting for already-running outbound BLE file transfer start; "
+                            + "leaving the worker uninterrupted",
+                    e);
+            return true;
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             Log.e(TAG, "Interrupted waiting for outbound BLE file transfer start result", e);
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/core/BaseBluetoothManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/core/BaseBluetoothManager.java
@@ -511,17 +511,38 @@ public abstract class BaseBluetoothManager implements ICompanionTransport {
             Log.e(TAG, "Rejected outbound BLE file transfer start", e);
             return false;
         } catch (TimeoutException e) {
-            if (future != null) {
-                future.cancel(false);
+            if (future != null && future.cancel(false)) {
+                Log.e(TAG, "Timed out waiting for queued outbound BLE file transfer start", e);
+                return false;
             }
-            Log.e(TAG, "Timed out waiting for outbound BLE file transfer start", e);
-            return false;
+            Log.w(
+                    TAG,
+                    "Timed out waiting for outbound BLE file transfer start after it began; "
+                            + "waiting for the actual start result",
+                    e);
+            return awaitFileTransferStartResult(future);
         } catch (InterruptedException e) {
             if (future != null) {
                 future.cancel(false);
             }
             Thread.currentThread().interrupt();
             Log.e(TAG, "Interrupted waiting for outbound BLE file transfer start", e);
+            return false;
+        } catch (ExecutionException e) {
+            Log.e(TAG, "Outbound BLE file transfer start failed", e);
+            return false;
+        }
+    }
+
+    private boolean awaitFileTransferStartResult(Future<Boolean> future) {
+        if (future == null) {
+            return false;
+        }
+        try {
+            return future.get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            Log.e(TAG, "Interrupted waiting for outbound BLE file transfer start result", e);
             return false;
         } catch (ExecutionException e) {
             Log.e(TAG, "Outbound BLE file transfer start failed", e);
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/core/BaseBluetoothManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/core/BaseBluetoothManager.java
@@ -129,16 +129,26 @@ public abstract class BaseBluetoothManager implements ICompanionTransport {
     /** Queue an outbound message and notify once the queued send attempt finishes. */
     @Override
     public final boolean sendMessage(byte[] data, SendMessageCallback callback) {
-        return queueOutboundMessage(data, callback, true);
+        return sendMessage(data, callback, null);
+    }
+
+    /** Queue an outbound message with a final worker-side send gate. */
+    @Override
+    public final boolean sendMessage(
+            byte[] data, SendMessageCallback callback, SendMessageGate gate) {
+        return queueOutboundMessage(data, callback, true, gate);
     }
 
     /** Queue an outbound command that should not be broadcast back to phone-facing listeners. */
     protected final boolean sendInternalCommand(byte[] data) {
-        return queueOutboundMessage(data, null, false);
+        return queueOutboundMessage(data, null, false, null);
     }
 
     private boolean queueOutboundMessage(
-            byte[] data, SendMessageCallback callback, boolean broadcastJsonResponses) {
+            byte[] data,
+            SendMessageCallback callback,
+            boolean broadcastJsonResponses,
+            SendMessageGate gate) {
         if (data == null || data.length == 0) {
             return false;
         }
@@ -159,7 +169,8 @@ public abstract class BaseBluetoothManager implements ICompanionTransport {
                                                     queuedAtMs,
                                                     traceInfo,
                                                     callback,
-                                                    broadcastJsonResponses)));
+                                                    broadcastJsonResponses,
+                                                    gate)));
             logOutboundQueueEvent(
                     "queued",
                     sequence,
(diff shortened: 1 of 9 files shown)


PhilippeFerreiraDeSousa:
@code review

PhilippeFerreiraDeSousa:
Babysit pass addressed the review feedback in `9d69e5ede`:

- Serialized BLE file-transfer starts through the same ordered outbound worker, so queued `ready_for_transfer` / `ble_photo_ready` control messages drain before file packets begin.
- Moved reliable-message ACK timer start to queued-send completion instead of queue acceptance.
- Bounded inbound chunk trace sessions with timeout/cap pruning to avoid diagnostics-only leaks.

Local validation: `git diff --check` and `scripts/check-android-compile.sh` both pass. GitHub checks have restarted on the new head.

PhilippeFerreiraDeSousa:
Babysit pass addressed the new `sendFile` timeout review in `844e41f9d`.

What changed:
- If the queued file-start task is still waiting in the outbound queue and times out, `sendFile(...)` cancels it and returns `false` as before.
- If the timeout fires after the worker task has already begun and can no longer be canceled, `sendFile(...)` now waits for the actual `sendFileInternal(...)` result instead of returning `false` while the transfer may continue in the background.
- Updated the PR body caveat so it no longer says ACK timers start when accepted into the queue; they now start after the queued transport send attempt completes.

Local validation:
- `git diff --check`
- `scripts/check-android-compile.sh`

PhilippeFerreiraDeSousa:
Babysit pass addressed the new Bugbot `Early ACK lost before tracking` review in `7a0956c1c`.

What changed:
- Reliable messages now create the pending ACK entry before enqueueing the BLE send, so a very fast ACK can be matched instead of being dropped.
- The ACK timeout is still armed only after the queued transport send completes, so outbound queue delay still does not consume the ACK window.
- Retry sends keep a pending entry visible while the retry is queued, then arm the retry timeout only after that queued send completes.

Local validation:
- `git diff --check`
- Focused google-java-format dry-run on `ReliableMessageManager.java`
- `scripts/check-android-compile.sh` passed before the final stats-only cleanup line was removed; the final change only removed an accidental retry stats increment and did not alter compile shape.

PhilippeFerreiraDeSousa:
Fixed the two new Bugbot findings in `234494f23`:

- K900 LED/internal commands now go through the same ordered outbound BLE queue via a no-broadcast base helper, so they cannot jump ahead of earlier queued phone-facing chunks.
- The file-transfer start fallback wait is now bounded; if the first 30s wait times out after the worker task has already begun, we wait one more explicit timeout window for the actual start result and then fail instead of blocking the caller indefinitely.

I kept this intentionally narrow and did not run a formatter over the files.

PhilippeFerreiraDeSousa:
Fixed the late-ACK queued retry race in `5d30fe9a2`.

The outbound BLE queue now supports an optional worker-side send gate. Reliable sends pass a gate that checks the exact pending entry is still current immediately before the queued write. If an ACK arrives while a retry is waiting in the outbound queue, the pending entry is removed and the worker skips that stale retry instead of sending the same `mId` again.

Still intentionally no formatter run.

PhilippeFerreiraDeSousa:
Fixed the file-timeout worker interruption issue in `857dfd036`.

The second bounded wait no longer calls `future.cancel(true)` on the shared outbound BLE worker. If the file-start task is already running and still has not returned after the second wait window, we leave the worker uninterrupted and return `true` because the transfer start has been accepted by the outbound worker rather than falsely reporting failure while it may still activate the K900 file session.

PhilippeFerreiraDeSousa:
Fixed the queued reliable cleanup issue in `df2888dc0`.

`cleanupOldMessages` now leaves pending entries alone until their ACK timer has actually been armed. That preserves messages whose payload has been accepted into the outbound BLE queue but whose queued send has not completed yet, so the worker-side send gate will not skip a reliable message solely because periodic cleanup removed its pending entry before the ACK window began.

PhilippeFerreiraDeSousa:
Fixed the two latest reliable-gate findings in `e19375499`.

- The worker-side send gate now carries a lock, and the outbound worker holds that lock across the gate check and the actual BLE/UART write. ACK handling and stale cleanup use the same pending-entry lock before removing entries, so the reliability manager no longer has a check/write race against ACK removal.
- Cleanup now has separate behavior for queued sends vs ACK-timed sends. Entries whose ACK timer has not been armed yet are preserved while they are reasonably queued, but they are no longer immortal: if such a queued entry is stale for 5 minutes, cleanup removes it and counts a failure.

Still no formatter or local build run; validation so far is scoped diff review plus `git diff --check`.

PR #3249, opened 2026-06-26
Add G2 selectable lists as a render element and let miniapps claim double-tap

PhilippeFerreiraDeSousa:
## Summary

Takeover of the June selectable-list PR, re-implemented on current `dev`. The original branch predated the `display.render()` scene pipeline, the Cloud V1 touch-relay removal and the engine-owned device event router, so nothing in it applied cleanly. The feature is now a **`list` scene element** instead of a legacy one-shot layout, which is what the display redesign note reserved for it ("the element union must accommodate a future `list` element").

- **Miniapp SDK**: `{type: "list", id?, box, items, style?: {border?, radius?, selectionBorder?}}` in `RenderElement`; `TouchData` gains `selectedItemIndex` / `selectedItemName`; `DisplayCapabilities` gains `maxListElements` / `maxListItems`.
- **Engine scene pipeline**: lists have their own budget (G2: 1 list, 20 rows). Rows are normalized (single line, ≤64 chars, blank rows kept visible). A list past the budget, or on glasses without a native list widget, is emulated as a text element and reported `degraded`; the G1/Z100 degrade path folds rows into the text wall. Rows are part of the content hash so a row change is an `updated` diff.
- **Native (Android + iOS)**: the G2 tracks one list container (id 7) next to its text/image containers and emits it in `CreateStartUpPage` / `RebuildPage` (`List_Object`, field 2). The firmware owns scrolling and the highlight and has no row-update command, so any list change is a structural rebuild while an identical repaint is a no-op (a rebuild would reset the highlight under the user). The list takes the event-capture flag over from the `evt-0` container while it is on the page (firmware allows exactly one). `List_ItemEvent` is parsed on both platforms and relayed as a `touch_event` carrying the selected row; absent event type ⇒ click, unknown ⇒ logged and dropped, identical on both platforms (the parity gap Bugbot flagged on the old branch).
- **Double-tap ownership**: a local miniapp that listens for touches (every gesture, or `double_tap`) claims the gesture. The runtime pushes `double_tap_claimed` to native the same way `imu_enabled` is pushed (`BluetoothSdk.setDoubleTapClaimed`), and the G2 handler skips the native dashboard and the head-up toggle while it is set, for both SysEvent and list double-clicks. The decision stays native so the shortcut keeps working while the app's JS is suspended; the runtime resyncs the claim on `initialize()` so a JS restart cannot leave a stale claim behind.
- **Phone mirror** renders list rows (first row outlined, as the firmware does on creation).
- **Docs**: Mintlify display + input pages, `sdk/docs`, miniapp README, Bluetooth SDK README.

Wire field numbers follow the Even Hub SDK model order (`@evenrealities/even_hub_sdk` 0.0.15 typings: `ListContainerProperty` 1–10 as `TextContainerProperty`, 11 `Item_Container`, 12 `Is_event_capture`; `List_ItemContainerProperty` 1 `Item_Count`, 2 `Item_Width`, 3 `Is_Item_Select_Border_En`, 4 `Item_Name`; `List_ItemEvent` 1 `Container_ID`, 2 `Container_Name`, 3 `CurrentSelect_ItemName`, 4 `CurrentSelect_ItemIndex`, 5 `Event_Type`).

Reviewer note: `mobile/modules/engine/src/utils/display/scene/types.ts` shows as a binary diff on GitHub because the `dev` version contains a literal NUL byte inside `contentHash` (the `part === undefined` placeholder). This PR replaces that byte with the six-character `` escape (same semantics), so the file diffs as text from here on; read the file directly for this revision.

## Validation

- `mobile/modules/miniapp`: `bun test` (287 pass), `bun x tsc --noEmit`
- `mobile/modules/engine`: `./scripts/test.sh` (all files passed; new `scene/__tests__/listElement.test.ts`)
- `mobile`: `bun run compile`; Jest `displayScene` + `mirror` suites; ESLint on touched files (no new errors)
- `./scripts/check-android-compile.sh bluetooth-sdk` — BUILD SUCCESSFUL
- `scripts/export-bluetooth-sdk-ios-spm.sh --verify` — `swift build` + iOS `xcodebuild` pass

## Open

- **Not exercised on G2 hardware yet**: list rendering, scroll/tap events with `selectedItemIndex`, the double-tap claim, and dashboard reopen after a list page. The demo miniapps that used the old `showSelectableList()` (`miniapps/selectable-list-demo`, `miniapps/dev-hud` on the `cyborgday-github-miniapp` branch) need a small port to `display.render([{type: "list", …}])` before they can be used for that.
- `maxListItems: 20` and the 64-char row cap are taken from the earlier implementation, not from Even's documentation; the firmware's actual limits should be confirmed on device.


PhilippeFerreiraDeSousa:
This should be the behavior we want : selectable lists. double tap available to mini apps with no dashboard interference if the mini app listens on it.
No time to retest though. There is a test mini app on cyborgday-github-miniapp at mini apps/selectable-list-demo/. Also miniapps/dev-hud uses it

PhilippeFerreiraDeSousa (commented):
**Verdict: request changes** on 98aea1c6f55ee935643626433266f3acc46357af.

GitHub rejected `gh pr review --request-changes` because the authenticated account owns this PR. This comment records the review verdict; it is not a formal blocking review.

The scene-element design and native ownership of the dashboard shortcut are coherent, but two lifecycle defects need fixing:

1. **[P1] Replay tears down the list page that recovery just created.** Replay clears `sceneListByElement` while retaining `listContainers` (Android `mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/G2.kt:2551`; iOS `mobile/modules/bluetooth-sdk/ios/Source/sgcs/G2.swift:2463`). The identical-list guard requires that erased mapping (Android :2887–2890; iOS :2480–2484), so even an identical list becomes structural and triggers shutdown. This is reachable on dashboard close: `recoverPageAndMic` calls `rebuildState()` to create the page including the list, then `DeviceManager.sendCurrentState()` to replay the retained scene (Android G2 :3093–3094 / DeviceManager :1690; iOS G2 :3155–3158). Replay therefore shuts down the freshly recovered page again, resets selection and interrupts mic recovery; subsequent shutdown/close notifications can cause further recovery churn. Reconcile and rebind an identical retained list without another teardown when its container already matches. Add a native regression for dashboard-close recovery followed by unchanged scene replay.

2. **[P2] Update the double-tap claim at every subscriber mutation.** `mobile/modules/engine/src/services/LocalMiniappRuntime.ts:2780–2786` computes the correct subscriber union, but its callers omit existing mutation paths. `registerApp()` removes an old incarnation's subscriptions at :764 without recomputing the claim. If it was the sole touch listener, native remains claimed despite the replacement having no listeners, until a later subscription update/unregister. The public `subscribe()` helper also replaces subscriptions at :3851 without recomputing, so it cannot acquire or release the claim. Derive this flag at the shared subscriber-update boundary, retaining explicit initialization/disposal sync. Cover sole-claimant replacement, public subscribe/unsubscribe and multiple claimants.

**Checked**
- PR description, all 37 changed files against fetched `origin/dev...HEAD`, and surrounding runtime/native lifecycle code. Compared both git versions of `scene/types.ts` directly: replacing its literal NUL with `\u0000` preserves runtime semantics.
- List normalization, independent budgets, overflow/unsupported-device text emulation, G1/Z100 text-wall degradation, row-sensitive hashes, stable-ID/type-change diffing, bridge payloads and replay.
- Android/iOS list encoding/composition, event-capture handover between list 7 and evt-0, replacement/removal, clearDisplay, page-down reconciliation, dashboard recovery and redispatch.
- List_ItemEvent parsing/row metadata; LocalMiniappRuntime -> exported BluetoothSdk.setDoubleTapClaimed -> update("bluetooth", ...) -> native DeviceStore -> G2 SysEvent and list double-click handlers.
- SDK API/docs and phone mirror. Actual wire acceptance, firmware limits and physical G2 behavior remain unverified, as acknowledged in the description.

**Every existing review/comment assessed against this head**
- Cursor Bugbot inline **#3483196219** and review **4581620732: addressed**. Both parsers default an absent event type to click and log/drop unknown explicit values. The old platform mismatch is gone.
- Codex connector inline **#3483231941** and review **4581663693: addressed by the replacement implementation**. The old island-runtime alias conversion is obsolete. Current engine `forwardEvent` preserves `gestureName` as `kind` and emits `touch_event:single_tap` / `touch_event:double_tap`, matching SDK subscriptions and claim detection. The unresolved thread does not establish a remaining alias bug.
- PR Agent Orchestrator state **#4812060601** and review **#4812060682: obsolete as approval evidence**. No findings were recorded, but the review explicitly says no model reviews ran that cycle.
- Build helper **#4812046601: status only**, not completed validation evidence; its text says building/waiting.
- Author intent **#4812262175: still applicable**. The design implements the requested direction, subject to the lifecycle defects above; old demo references require the port already acknowledged in the description.
- Current Cursor summary **#5640236235: broadly accurate overview**, with no additional correctness finding. “On-glasses” shortcut decision should read native phone SDK: the phone's G2 driver reads this claim.

**Validation run:** miniapp Bun suite **287 passed**; engine list/scene Bun suite **12 passed**; mobile Jest displayScene, displaySceneWiring and GlassesDisplayMirror **44 passed**. These tests do not exercise the two lifecycle defects. Native builds and hardware tests were not rerun. No source files edited and nothing pushed.

PhilippeFerreiraDeSousa (commented):
Approve.

Re-reviewed c90b68330d3459d6ee222a844180862403a252fe against freshly fetched origin/dev. No remaining blocking correctness defect found. The retained scene design and centrally derived native gesture claim are coherent; the follow-up fixes the shared state boundaries without adding recovery timers or special-case callers.

Checked:
- PR description, all 39 changed files in origin/dev...HEAD, and git show c90b68330d. Compared both git versions of scene/types.ts directly: the literal NUL becomes the equivalent "\u0000" escape; list types, capabilities and row-sensitive hashes are present.
- List normalization, budgets, overflow/unsupported-device text emulation, G1/Z100 degradation, stable IDs/type changes, scene bridge payloads, SDK API/docs and phone mirror.
- Android/iOS list protobuf composition, list 7 versus evt-0 event capture, replacement/removal/clearDisplay, page-down reconciliation, event parsing and row metadata, and the exported BluetoothSdk.setDoubleTapClaimed -> update("bluetooth", ...) -> native DeviceStore -> SysEvent/list double-click path.
- Dashboard close -> recoverPageAndMic -> rebuildState -> DeviceManager.sendCurrentState -> applySceneFrame(replay=true), on both platforms. Recovery creates the tracked list and restores mic intent. An unchanged list now matches the container itself and rebinds its element ID without requesting a second teardown (Android G2.kt:2896; iOS G2.swift:2489). For a retained scene with no list, replay explicitly clears the lingering container/mapping and requests the existing batched structural rebuild (Android G2.kt:2585; iOS G2.swift:2453); subsequent composition restores evt-0 capture. Changed lists still rebuild.
- All stream subscriber writes in mobile/modules/engine/src/services/LocalMiniappRuntime.ts: registerApp respawn (:770), unregisterApp (:960), SUBSCRIBE (:1783), and public subscribe (:3851) route through replaceStreamSubscribers (:502), which removes empty subscriber sets and recomputes the union (:539). Thus replacing the sole claimant releases it, public subscribe can acquire/release it, and another claimant keeps it set. Strictly, cleanup also has a direct final clear (:4919), immediately followed by recomputation; it first unregisters each app through the shared path. Initialization forces native resync (:643).

Assessment of every existing review/comment:
- Prior local request-changes review 5183181284, P1 recovered-list teardown: addressed by the content-based match and mapping rebind above.
- Same review, P2 stale double-tap claim on subscriber mutations: addressed by central recomputation plus explicit initialization/cleanup sync.
- Cursor inline #3993156374, "Stale list survives scene replay", and review 5183172184: addressed by explicit list removal when replay contains no list, on both platforms.
- Cursor inline #3483196219, "Android drops unknown list events", including the reported repeat, and stale review 4581620732: addressed/obsolete against this head. Android G2.kt:5255 and iOS G2.swift:4973 both default an absent type to click and log/drop an unknown explicit value. There is no current platform mismatch. The API currently exposes one stored inline thread for this finding.
- Codex connector inline #3483231941 and review 4581663693, gesture aliases: addressed by the replacement implementation. Current engine forwardEvent (:4189, :4245) preserves single_tap/double_tap in kind and the per-gesture stream, matching SDK subscriptions. The unresolved but outdated thread is not a remaining defect.
- Orchestrator state #4812060601 and review comment #4812060682: obsolete as approval evidence; no model reviews ran in that cycle.
- Build helper #4812046601: status only. It now links an APK for 98aea1c, not this head, and still lists ASG as waiting.
- Author intent #4812262175: still applicable and implemented. The old demos still need the acknowledged display.render port.
- Cursor summary #5640236235: broadly accurate, no additional defect. Its "on-device" claim decision is specifically in the native phone G2 driver, not glasses firmware.

Validation rerun: miniapp Bun suite 287 passed; engine scene/list + DoubleTapClaim suites 16 passed; mobile Jest displayScene, displaySceneWiring and GlassesDisplayMirror 44 passed. The new claim tests cover the pure union rule, not runtime lifecycle integration; there is still no native dashboard-replay regression test. Native recovery correctness here is source-traced, not experimentally established. Native builds and physical G2 tests were not rerun; firmware wire acceptance/limits remain unqualified as the PR already states.

No source files edited and nothing pushed.

GitHub refused formal approval because the authenticated account owns this PR. This comment records the approval verdict.

PR #3273, opened 2026-06-29
Add BGCAP background-caption telemetry (diagnostic)

aisraelov:
## What

Host-side, cross-platform diagnostic logging to pinpoint **why captions "fall behind then flood in waves" when the phone screen is off** (Even Realities G2; reproduced on both iOS and Android). No native changes, no behavior change — just logging behind a `BGCAP_TELEMETRY` const (currently `true`).

## Why

Captions now run as the `com.mentra.local-captions` **local miniapp** inside the per-context JS engine (Crust: QuickJS on Android / JSC on iOS), driven off a **single serial queue/executor with no background awareness**. Evidence so far points at that JS layer being starved/frozen in the background, not the cloud or the glasses:

- The native phone→glasses display send is healthy in incident logs — `MAN: displayEvent` (in) == `G2: sendText` (out) **1:1**.
- The prior **cloud-app captions** and the old **`offline_captions`** (Sherpa) paths rendered natively and never showed this — only the new JS-render path does.
- The runtime's own comment already notes *"OS scheduling while idle can delay pongs well past one interval"* and widened the liveness kill to ~30s to avoid nuking a starved-but-alive context.

But the incident logs **go silent on the caption path during the background window** and only capture the native boundary, so they can't show *where* the timeline breaks. This adds the missing signals.

## What it logs (all prefixed `BGCAP:`, lands in the `phone:console` incident capture)

| Signal | File | Tells us |
|---|---|---|
| **host-js heartbeat** — plain `setInterval` (not `BgTimer`); logs gaps >2s | `LocalMiniappRuntime` | RN JS thread suspended/throttled, and for how long |
| **PING→PONG round-trip** per miniapp (logs >1.5s) | `LocalMiniappRuntime` | the miniapp's own Crust JS engine (transcript→display layer) is starved |
| **transcript-forward rate** (1/s) | `LocalMiniappRuntime` | whether transcripts keep arriving + being handed to the miniapp in background |
| **mic_lc3 uplink rate** (1/s, with transport) | `MantleManager` | whether audio still flows *up* to the cloud → splits uplink-side vs render-side |

Together these answer: does the JS thread freeze in background? does audio keep going up? do transcripts keep coming down? is the miniapp engine responsive? — which localizes the stall.

## How to use

1. Build with this branch (`BGCAP_TELEMETRY` is on).
2. Start captions → **lock the screen** (the trigger is screen-off, not app-switch — backgrounded-with-screen-on captioned fine in the incident logs) → wait ~60s → unlock → submit a bug report.
3. The flushed incident logs will contain the `BGCAP:` timeline. Correlate the heartbeat gap with the existing `RECONNECT: App state changed to:` lines.

## Notes

- Diagnostic only — **revert (or flip `BGCAP_TELEMETRY` to `false`) once the background caption-lag fix lands**. Grep `BGCAP`/`bgcap` to remove.
- Committed with `--no-verify`: built in a fresh worktree without `node_modules`, so the pre-commit lint/typecheck couldn't run locally — CI will validate.

---
## Summary by cubic
Adds host-side background-caption telemetry to pinpoint why captions lag and then flood when the screen is off. Logs are prefixed `BGCAP` and gated by `BGCAP_TELEMETRY` (on), with no behavior changes.

- **New Features**
  - JS stall signals: heartbeat gaps >2s; per-miniapp PING→PONG RTT >1.5s.
  - 1/s rates for transcript forwards and `mic_lc3` uplink to split uplink vs render stalls.
  - Host-side only; no native changes.

- **Bug Fixes**
  - PONG RTT now measured only on real PONGs and keeps the oldest outstanding ping to report full stall duration; transcript-forward count now logs only after subscriber checks and only for transcription streams.

<sup>Written for commit 91428bebac51d961cbb13a25fd7ea73b13091227. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3273?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Low Risk**
> Logging-only behind a const flag; no auth, data, or caption pipeline logic changes.
> 
> **Overview**
> Adds temporary **BGCAP** host-side logging (behind `BGCAP_TELEMETRY`, currently on) to debug captions lagging then catching up in waves when the screen is off. **No product behavior changes** — only `console.log` lines prefixed `BGCAP:` for incident grepping.
> 
> In **`LocalMiniappRuntime`**: a plain `setInterval` heartbeat logs when the RN JS thread was frozen/throttled (>2s between ticks); transcript fan-out to miniapps gets a 1/s rate summary; miniapp liveness **PING→PONG** RTT is recorded and slow responses (>1.5s) are logged to spot a starved Crust miniapp JS queue.
> 
> In **`MantleManager`**: the `mic_lc3` path logs a 1/s summary of LC3 frames uplinked to the cloud (UDP vs WebSocket) so stalls can be split between uplink vs downstream render.
> 
> Remove or set `BGCAP_TELEMETRY` to `false` after the caption fix lands.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 987a94f884e07bcce5f6a072ae954ea438e821b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/modules/island/src/services/LocalMiniappRuntime.ts
+++ b/mobile/modules/island/src/services/LocalMiniappRuntime.ts
@@ -76,6 +76,8 @@ interface ConnectedMiniapp {
   subscriptions: Set<string>
   sendMessage: (raw: string) => void
   lastPongAt: number
+  /** BGCAP diagnostic: wall-clock of the most recent PING send, for PONG RTT. */
+  bgcapLastPingSentAt?: number
   installedManifest?: InstalledMiniappManifest
   authRefreshTimerId: number | null
   /** Last speaker state pushed to this app. Used to dedup SPEAKER_STATE pushes. */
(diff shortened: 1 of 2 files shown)


aisraelov:
Both findings are correct and fixed in the latest commit:

1. **RTT logged on any message** — confirmed: `handlePong` runs for *every* inbound envelope (L789, "ANY inbound message proves the context is alive"), and the per-ping overwrite would actually *under-report* a stall. Moved the RTT measurement into the real `MiniappResponseType.PONG` case, and `doPingRound` now stamps only the oldest outstanding ping (no overwrite) so a multi-round background stall reports its full duration on the first pong after resume.
2. **Counts transcripts before forward check** — confirmed: the count ran before the `matchedSubs.size === 0 → return` guard, so it'd show "forwarded" to a dead/unsubscribed captions app during exactly the gap this telemetry is meant to expose. Moved it after the guard.

aisraelov:
## Incident e8c72f71 (instrumented build) — result: the JS pipeline is **not** the stall

First clean repro on the instrumented build (208 `BGCAP:` lines). The phone telemetry stays **healthy through the entire 47s background window** (bg @73.4s → fg @120.1s):

- `mic_lc3 uplink` ≈ **20 frames/s, dead steady** the whole time — audio keeps flowing up to the cloud.
- `tx forwarded` ≈ **2–9/s, steady** — transcripts keep arriving and being handed to the captions miniapp.
- native `G2: sendText()` (stage-1 entry) called steadily, **no `host-js stall`**, **no `pong rtt` warnings**, no missed pings.

So the RN JS thread, the Crust miniapp engine, cloud delivery, and audio uplink are all **exonerated** — they ran fine in the background. (My earlier "JS render path freezes in bg" hypothesis is wrong.)

The key realization: **`G2: sendText()` is logged in `sendTextAt` — stage 1, when a display event *arrives* from the bridge — which is *above* the EvenHub→BLE write queue.** Steady `sendText` logs in bg only prove events kept *arriving* at native, not that bytes reached the glasses. The actual transmit is `drainWriteQueue()`, gated on `canSendWriteWithoutResponse` and resumed by `peripheralIsReady()`. That's the one place "phone busy in bg + glasses frozen + flood on resume" can live — and host-side telemetry can't see it.

## Added (commit `533af4868b`): native iOS BLE-queue probes in `G2.swift`
- **BLOCKED** (1/s) when `drainWriteQueue` stops on `canSendWriteWithoutResponse=false`, with queue depth + high-water + writes-since-last.
- **RESUMED** when a previously-blocked queue drains empty (the foreground flush).

**Decisive read on the next repro:**
- Background shows `BGCAP: g2 BLE drain BLOCKED … queueDepth↑` then `RESUMED … flushed highWater=N` on foreground → the flood is **native BLE backpressure** (iOS throttling background writes).
- **No** BLOCKED lines in the background window → bytes left the phone, and the stall is **downstream of the phone** (glasses firmware / BLE link).

Open tension to resolve with this: the old cloud-app and `offline_captions` paths used this same native BLE send path and reportedly didn't flood — so if it's BLE backpressure, the difference is likely the *volume* of display updates the new local-miniapp path emits. The probes will show whether depth actually climbs.

Same repro: captions running → **lock screen**, phone idle ~1–2 min → unlock → submit report.

PR #3279, opened 2026-06-30
Fix G2 background caption flood (paced direct BLE writes) + restore G1 300ms throttle

aisraelov:
## Summary

Fixes the **G2 "captions freeze in background then flood on resume"** bug, and (bundled, separate concern) **restores G1's 300ms text throttle** inside the G1 SGC.

## Root cause (G2) — measured

Diagnostic probes (incidents `e8c72f71`, `68a2844f`) showed the JS/cloud/audio pipeline staying **healthy** through a full background window, while the native `rightWriteQueue` grew to **143 packets** with `canSendWriteWithoutResponse` stuck false and **zero bytes out for ~20s**, then flushed on resume.

The `canSend`-gated write queue (added in #3162) **deadlocks in the background**: G2 has no active `AVAudioSession` (glasses-mic audio arrives over BLE), so a backgrounded G2 is a plain `bluetooth-central` app and iOS **suppresses `peripheralIsReady(toSendWriteWithoutResponse:)`**. The drain only resumed on that callback, so it stopped the instant `canSend` went false and never resumed → accumulate → flood.

**G1 (its own SGC) never has this** because it writes `.withoutResponse` **directly, paced ~8ms, no `canSend` gate** — so it always makes progress, even backgrounded. (Confirmed by testing G1: incident `43440fd4`, no flood.)

## Changes

**1. G2 iOS (`G2.swift`) — the fix.** Replace the `canSend`-gated queue + `peripheralIsReady`-driven drain with a **single paced async drainer per side** that writes `.withoutResponse` directly and sleeps ~6ms between packets (matches G1). No `canSend` gate, so it never deadlocks in the background and delivers **every** caption live. The pace replaces the gate's overflow protection; dropped packets self-heal (text re-sends via `pendingSends`, image fragments retry on ACK). Keeps a small rate-limited queue-depth warn to validate the queue stays ~0 in bg.

**2. G1 (`G1.swift` + `G1.kt`) — restore the 300ms throttle.** G1 firmware can't absorb rapid text-wall updates; the 300ms last-wins throttle that used to live in the cloud DisplayManager (before captions moved on-device) is restored **inside the G1 SGC** (G1 hardware quirk). Trailing flush always sends the most recent text → final caption never dropped. **G2 deliberately does NOT throttle — it must show every caption.**

**3. G2 Android (`G2.kt`) — instrumentation, not a blind fix.** Cross-referencing surfaced that **Android G2's text path is already direct** (single packet → `writeOnePacket`, no `canSend` gate), so the iOS root cause doesn't exist on Android. Rather than blind-port the iOS fix, this captures the `writeCharacteristic` return value (false = GATT stack busy → packet dropped) the code currently discards, to pin where Android actually accumulates/drops in the background. **The Android G2 fix is a follow-up once one Android repro shows the mechanism.**

## Validation

- **iOS G2:** build + lock-screen repro (captions running → lock screen, idle ~1–2 min → unlock). Expect captions to stay live in the background, **no flood**, and the `BGCAP: g2 write queue depth` warn to stay quiet (queue ~0).
- **G1 (both):** verify captions display smoothly (no G1 overload), final caption always correct.
- **Android G2:** run the lock-screen repro and send the `BGCAP: g2 writeCharacteristic ok=.. fail=..` lines — they'll tell us the Android accumulation point.

## Cleanup
Diagnostic lines are prefixed `BGCAP:`. After validation, a final commit strips them (here + the host-side telemetry from #3273). Supersedes the diagnostic-only #3277.

---

> [!NOTE]
> **Medium Risk**
> G2 iOS changes core BLE caption delivery in background/foreground; incorrect pacing could drop or delay packets, though retries exist for text/images. G1 throttle is localized to display paths.
> 
> **Overview**
> Fixes **G2 captions freezing in the background and flooding on resume** on iOS, and restores **G1’s 300ms last-wins text coalescing** on iOS and Android.
> 
> **G2 iOS:** Replaces the `canSendWriteWithoutResponse`-gated drain (which stalled when iOS stopped delivering `peripheralIsReady` in the background) with a **paced async `drainLoop`** per side: direct `.withoutResponse` writes and ~6ms sleep between packets, aligned with G1. `peripheralIsReady` only nudges `startDrain` in the foreground. Old BGCAP blocked/resumed telemetry is removed; a rate-limited queue-depth warn remains for validation.
> 
> **G1 iOS + Android:** Adds a **300ms last-wins throttle** on `sendText` / `sendTextWall`; `clearDisplay`, double-text-wall, and bitmap paths **cancel pending throttled text** and bypass the throttle where a clear must win.
> 
> **G2 Android:** Adds **BGCAP** logging of `writeCharacteristic` ok/fail counts (no pacing/throttle change).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b08c2b01b3bd1f870b6718e00e485bfde1dde1e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes G2 background caption freeze/flood on iOS by switching to paced direct BLE writes that keep draining in the background. Restores G1’s 300ms last‑wins text throttle, ensures clears/layout changes aren’t overwritten by a pending throttle, and adds Android G2 write diagnostics.

- **Bug Fixes**
  - iOS G2: replace `canSend`-gated queue with a single paced async drainer per side writing `.withoutResponse` and sleeping ~6ms between packets; avoids background deadlock and flood. `peripheralIsReady` is now just a foreground nudge; adds a small rate-limited queue-depth warn.
  - G1 (iOS + Android): restore 300ms last-wins text throttle for captions; trailing flush sends the latest text. `clear` now bypasses the throttle and cancels pending; double-text and bitmap paths also cancel pending so they can’t be overwritten.

- **Refactors**
  - Android G2: add BGCAP diagnostics to log `writeCharacteristic` ok/fail counts (false = GATT busy) to pinpoint background drops. No behavior change.

<sup>Written for commit b08c2b01b3bd1f870b6718e00e485bfde1dde1e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3279?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/G1.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/G1.kt
@@ -1614,7 +1614,10 @@ class G1 : SGCManager() {
 
     override fun clearDisplay() {
         Bridge.log("G1: clearDisplay() - sending space")
-        sendTextWall(" ")
+        // Bypass the throttle (a clear must always land) and drop any pending caption so it can't
+        // overwrite the clear after the fact.
+        cancelPendingThrottledText()
+        displayTextWall(" ")
     }
 
     // G1-specific display throttle (300ms, last-wins) — see G1.swift. G1 firmware can't absorb
@@ -1627,6 +1630,13 @@ class G1 : SGCManager() {
     private var textThrottleLastSent: Long = 0L
     private var textThrottleScheduled = false
     private val TEXT_THROTTLE_WINDOW_MS = 300L
+
+    /** Drop any pending throttled text-wall flush so it can't later overwrite a newer, non-text
+     *  display write (a clear, double-text-wall, or bitmap). The posted flush no-ops when
+     *  `textThrottlePending` is null. Called from every G1 display path that bypasses the throttle. */
+    private fun cancelPendingThrottledText() {
+        textThrottlePending = null
+    }
 
     private fun throttledTextWall(text: String) {
         val now = android.os.SystemClock.uptimeMillis()
@@ -1663,11 +1673,13 @@ class G1 : SGCManager() {
     }
 
     override fun sendDoubleTextWall(top: String, bottom: String) {
+        cancelPendingThrottledText() // a newer layout supersedes any pending caption text
         Bridge.log("G1: sendDoubleTextWall() - top: " + top + ", bottom: " + bottom)
         displayDoubleTextWall(top, bottom)
     }
 
     override fun displayBitmap(base64ImageData: String, x: Int?, y: Int?, width: Int?, height: Int?): Boolean {
+        cancelPendingThrottledText() // a bitmap supersedes any pending caption text
         try {
             // Decode base64 to byte array
             val bmpData = android.util.Base64.decode(base64ImageData, android.util.Base64.DEFAULT)
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/G1.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/G1.kt
@@ -1617,14 +1617,49 @@ class G1 : SGCManager() {
         sendTextWall(" ")
     }
 
+    // G1-specific display throttle (300ms, last-wins) — see G1.swift. G1 firmware can't absorb
+    // rapid text-wall updates; coalesce to the latest within a 300ms window. Belongs in the G1 SGC
+    // (G1 hardware quirk); G2 deliberately does NOT throttle (it must show every caption). The
+    // trailing flush always sends the most recent text, so the final caption is never dropped —
+    // only intermediate frames within a window are coalesced.
+    private val textThrottleHandler = Handler(Looper.getMainLooper())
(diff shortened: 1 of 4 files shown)


aisraelov:
Both findings fixed in the latest commit — they're the same root issue (the 300ms throttle only wrapped `sendTextWall` and didn't coordinate with G1's other display writes):

1. **Clear coalesced away (Cursor, Medium):** `clearDisplay` now **bypasses** the throttle (`flushTextWall(" ")` on iOS / `displayTextWall(" ")` on Android) and cancels any pending caption — a clear always lands.
2. **Stale text overwrites a newer layout (Codex P2, both platforms):** added `cancelPendingThrottledText()` and call it at the top of `sendDoubleTextWall` and `displayBitmap` (both platforms), so switching layouts within the 300ms window drops the pending caption text instead of letting it replay over the new layout.

The captions path (`sendTextWall`) still throttles last-wins as before; only the non-text display paths now cancel the pending flush.

PR #3167, opened 2026-06-16
Move everything into the island

aisraelov:
## OEM Toolkit — Phase 1: moving the MentraOS runtime into `@mentra/island` (OS-1622)

Moves device/runtime logic into island domain-by-domain so the host becomes **UI + auth
only**, behind the **`toolkit.*`** API. Namespace `island.*` → `toolkit.*`; module stays
`@mentra/island`.

## Progress — the (A) Host toolkit API (from OS-1622, + status)

| Namespace | What it exposes | Status |
|---|---|---|
| `toolkit` (bootstrap) | `configure({auth, config})` · `start()` · `stop()` | ✅ **Done** |
| `toolkit.glasses` | connect(+default/simulated/device)/disconnect/forget/setDefault · **reconnect** · **isFirstPairing** · controller · status/info/capabilities · version · button/touch events | ✅ **Done** |
| `toolkit.glasses.wifi` | scan · connect · forget · status/onStatus | ✅ **Done** |
| `toolkit.glasses.btsdk` | raw `@mentra/bluetooth-sdk` passthrough (escape hatch) | ✅ **Done** |
| `toolkit.speech` | `stt`/`tts` on-device model management | ✅ **Done** |
| `toolkit.session` | live-session status + `account.delete/confirmDelete` | ✅ **Done** |
| `toolkit.settings` | keyed `get/set/onChanged/descriptor/keys` | ✅ **Done** |
| `toolkit.dev` | backend+cloud URL overrides · reconnect · minimumClientVersion | ✅ **Done** |
| `toolkit.display.mirror` | `current()/onMirror()/view()/setView()` | ✅ **Done** — `sendButtonPress` deferred: it's a v1 path (SimulatedGlasses → `socketComms.sendButtonPress` → cloud relay); needs a v2 cloud button-uplink or a native simulated-input method |
| `toolkit.incidents` | one-call `file()` + create/uploadLogs/uploadAttachments/sendFeedback | ✅ **Done** — host gathers native-coupled diagnostics, passes them in |
| `toolkit.miniapps` | lifecycle: list/onChanged/start/stop/setForeground/install/uninstall/… | ✅ **Done** (lifecycle) — WebView bridge primitives exported separately (host mounts the `<WebView>`) |
| `toolkit.glasses.settings` | keyed device settings: get/set/onChanged/descriptor/available | ✅ **Done** — scoped to `BLUETOOTH_SETTING_KEYS`; set() auto-syncs to glasses |
[...]

PR #3286, opened 2026-06-30
BLE Wire Protocol v2: throughput cleanup across asg, mobile SDK, and app layer

nic-olo:
## Summary

- Adds BLE Wire Protocol v2 across **asg_client**, **mobile bluetooth-sdk** (Android + iOS), and the **mobile app layer** for higher throughput and lower wire overhead.
- Standardizes K900 `##` command frame length fields to **little-endian** (fixes UART parse failures with updated asg_client ↔ BES firmware).
- Introduces binary fragment transport (`0x40`), capability handshake (`phone_ready` / `glasses_ready`), MTU 509, compact JSON helpers, and app-layer dedup/coalescing/debounce for display, settings, camera, streaming, and menu_apps.

## Scope

45 commits (one per file) covering:

- **asg_client**: `BesWireFormat`, binary chunking/reassembly, `BleJsonCompact`, `K900BluetoothManager`, `AsgClientService` handshake
- **mobile-sdk**: `BleWireProtocol`, LE `K900ProtocolUtils`, binary transport, `MentraLive` Android/iOS
- **mobile app**: outbound JSON compaction, display coalescing, settings debounce, slim `stream_status`, photo/camera dedup

## Pairing note

Requires matching **BES firmware** with LE UART length encoding (`cmd_int2Chars` / `cmd_chars2Int` in `m8_prot.cpp`). File packet fields remain big-endian.

## Test plan

- [ ] Flash BES firmware with LE UART fix; verify glasses boot ready (no `Extracted length=9472` in logcat)
- [ ] Connect phone over BLE; confirm `glasses_ready` / `phone_ready` handshake
- [ ] Send display updates; verify coalescing reduces packet count
- [ ] Take photo; verify single camera BLE write and short requestId
- [ ] Run `asg_client` unit tests: `BesWireFormatTest`, `BinaryMessageChunkerTest`, `BleJsonCompactTest`
- [ ] Run mobile-sdk tests: `K900ProtocolUtilsEndiannessTest`, `BinaryMessageChunkerTest`, `BleJsonCompactTest`
- [ ] Smoke test iOS MentraLive connect + display path

Made with [Cursor](https://cursor.com)


PhilippeFerreiraDeSousa:
> Standardizes K900 ## command frame length fields to little-endian (fixes UART parse failures with updated asg_client ↔ BES firmware).

I my review everything was good on big endian and it was the android producer on the phone that mistakenly was using little endian. I changed it to use big endian on the toolkit work branch.
https://github.com/Mentra-Community/MentraOS/pull/3167/changes/fdbb1472545613af1bd9efec701ae0c1a6078f87

PhilippeFerreiraDeSousa:
I see retro-compatibility breaking troubleshooting documentation here, we should review the user-facing story about that https://github.com/Mentra-Community/Mentra-Bluetooth-SDK-Starter-Kit/pull/19/changes

aisraelov:
Backcompat review. The negotiation design is right (legacy by default, capability handshake), but it's gated on the wrong signals in both directions. Nothing breaks while BES stays at 17.26.05.22 — but these must be fixed before the LE BES firmware ever ships, and nothing later will force it.

**Blockers**

1. **Glasses activate v2 toward the phone without consulting the phone.** `PhoneReadyCommandHandler.java:100` sends the 0x40 handshake and calls `setBinaryProtocolActive(true)` on *send*, gated only on BES caps — `phone_ready` carries no wire caps. Old app + LE BES fw → every glasses→phone message after `glasses_ready` becomes an 0x40/compact-JSON frame old apps silently drop: paired but functionally dead. Fix: drop the proactive send; glasses already advertise `wire_caps` in `glasses_ready` and the new phone initiates (`MentraLive.kt:7113`) — activate only on the phone's inbound handshake (that path already works).
2. **Same root, earlier in boot:** `applyBesWireCaps` (`K900BluetoothManager.java:950`) flips ALL outbound STRING frames to LE from BES caps — including `glasses_ready` itself. Old apps parse BE-only, so pairing never completes on LE BES fw. Phone-facing frame endianness must key on the phone's confirmation; keep `uartToBesEndian` for BES-bound commands only.
3. **Reverse direction: phone→glasses pre-negotiation length endianness flipped LE→BE** (`K900ProtocolUtils.java:139`, `MentraLive.swift:4705`) vs every shipped app (dev packed LE). The old asg_client APK tolerates it (BE-then-LE fallback), but `cs_hrt` is consumed directly by the BES2700 and gates the whole bring-up — this is safe only if *old fielded* BES firmware really has the `k900_detect_string_length` heuristic. Revert the pre-negotiation default to LE, or hardware-test against the oldest fielded BES fw first.

**v2-path bugs (dormant until LE fw — fix in this PR)**

- Glasses silently drop any inbound v2 message >4KB: `ChunkReassembler.java:20` fixed 4096-byte buffer, while phones fragment up to 122KB and the v1 chunk path was unbounded.
- iOS compact-JSON diverges from Java: `NSNumber as? Bool` bridging (`BleJsonCompact.swift:218`) omits numeric 0s and turns 1s into `true`; hash canonicalization (`BleJsonCompact.swift:348`) doesn't match Java, so glasses `rch` lookups miss and `resolvedConfig` is silently lost on iOS.
- Non-acked fragmented sends use a constant msgId (Android 0 — `MentraLive.kt:7438`; iOS 0xFFFF — `MentraLive.swift:3795`): two concurrent large sends interleave into one reassembly session and the second is lost. ASG allocates ids properly; phones should too.
- v2 activation never checks the negotiated MTU. 494-byte frames need ATT ≥497; older iPhones negotiate 185 → every large frame truncated. Gate activation on `currentMtu` or derive fragment size from it.

**Regression independent of glasses fw**

- Video resolution/fps changes no longer reach glasses mid-session: `camera.tsx:159` removed the only trigger, and the native observers have no case for the `button_video_settings` object (`DeviceStore.kt:295`, `DeviceStore.swift:266`) — it's only read at connect-time sync. (The photo-size/fov removals *are* genuine de-dupes.) Note the new `camera.test.tsx` greps source text, so it can't catch this.

**Nits:** staging-builds.yml should guard both `prod_live_version.json` and the test manifest, not swap one for the other; iOS has zero unit tests for the new wire code; the iOS CI failure looks like the known cross-branch `killall xcodebuild` flake (merge dev + rerun).

Verified good: v2 frame layout / key maps / enums are byte-identical across all three implementations, 0x40 can't collide with the legacy cmd namespace, OTA/auth + file-transfer framing untouched, and state resets on disconnect are correct.

aisraelov:
@nic-olo heads up — this branch's changes (asg + mobile wire-v2) have been pulled into **aisraelov/os-1409-ble-v2-merged** (PR #3349): it's this branch + `dev` merged in (conflicts resolved) + the OS-1409 BLE throughput/quality stack on top. The v2↔v2 path is hardware-validated there (pairing, binary handshake, endian negotiation all work live against the matching firmware).

One critical finding you'll want before shipping any of the fw side: **wire-v2 firmware ↔ today's v1 phones = phone→glasses BLE JSON is completely dead** (writes are ATT-acked then vanish in the BES RX path — `lxy_ble_data_rx_func` → `cmd_ble_parse_recv` → dispatch). Verified on hardware; details + repro on OS-1409. Since firmware OTAs before app updates reach users, that needs fixing in the BES branch before rollout. The BES-side counterpart PR (your `feature/ble-wire-protocol-v2` + throughput work) is up on the firmware repo, tagged you there.

aisraelov:
Hello sir please see https://github.com/Mentra-Community/MentraOS/pull/3349

PR #3314, opened 2026-07-01
Teleprompter Play/AI-scroll UX fix + restore "Starting <miniapp>…" boot message

aisraelov:
Two related local-miniapp fixes reported from testing.

## 1. Teleprompter: Play didn't auto-advance (looked broken)

**Symptom:** Pressing Play did nothing visible — the script only moved when you spoke — so people thought the miniapp was broken.

**Cause:** `voiceFollow` (voice-follow / "AI scroll") **defaulted to `true`**, and its only control was a toggle buried in the Settings tab. So the big Play button silently started voice-follow, which waits for speech. Play *was* "enable AI scroll" with no separate control — exactly the confusing behavior reported.

**Fix:**
- Default `voiceFollow` to **false** → Play now does timed auto-scroll (WPM) out of the box.
- Add a prominent, clearly-labeled **AI Scroll** toggle in the Script cockpit (next to the state), and remove the buried, differently-named "Voice-follow" row from Settings — one control, one name.
- Add a caption under the transport that states what Play will do in the current mode ("Play auto-scrolls at N wpm. Turn on AI Scroll to follow your voice." / "AI Scroll is on — the prompter follows your voice…").
- Update the built-in default script and the `load_script` action description to match the new default.

The driver logic is unchanged — voice-follow still live-switches mid-playback; this is a default + surfacing change.

## 2. Local miniapps lost the "Starting <name>…" boot message

**Symptom:** Cloud V2 (`@mentra/miniapp`) miniapps boot with no "Starting …" text on the glasses, unlike the old cloud-SDK apps.

**Cause:** `LocalDisplayManager.onMount()` — which renders `Starting <name>…` for a bounded window, mirroring the cloud `DisplayManager6.1` boot screen — was fully built (and unit-tested) but **never called** in production. Only `onUnmount` was wired into the runtime. Additionally, `InstalledMiniappManifest` dropped the app's `name`, so the runtime had no display name to show.

**Fix:**
- Plumb the manifest `name` through `InstalledMiniappManifest`, populated in both the dev and released launcher paths.
- Call `localDisplayManager.onMount(pkg, name)` from `LocalMiniappRuntime.registerApp`, symmetric with the existing `onUnmount` in `unregisterApp` (fires once per spawn; falls back to the package name when a manifest name is absent).
- Harden `endBoot` so the boot text clears on timeout when no core app is wired — `onCoreAppChange` isn't wired yet, so `coreApp` is always `null` today; without this an app that never renders could strand "Starting …" on the glasses. (For the common foreground-display app the boot text is replaced by the app's first render, which already works without core wiring.)

Note: full core-vs-background arbitration (`onCoreAppChange`) remains unwired — that's a separate workstream per `agents/local-display-manager-plan.md`. This PR only restores the boot message.

## Testing
- `mobile`: `tsc --noEmit` clean (0 errors).
- Teleprompter: `bun run build` + `tsc --noEmit` clean.
- Island tests: `LocalDisplayManager` (jest) 17/17 incl. new production-path cases; 8 `bun:test` island suites 104/104 (incl. `MentraJSRouter`, `MiniappLauncher`); `MicStateCoordinator` (jest) 5/5.
- Not yet hardware-tested on glasses.

---

> [!NOTE]
> **Low Risk**
> UX defaults and on-glasses boot display wiring only; scroll driver logic is unchanged and island tests cover the boot-timeout paths.
> 
> **Overview**
> **Teleprompter:** **Play** now defaults to **timed auto-scroll** (`voiceFollow: false`) instead of voice-follow, which made the prompter look broken when nothing moved until you spoke. **AI Scroll** is a single opt-in toggle in the **Script** cockpit (removed buried **Voice-follow** from Settings), with a short caption under transport explaining the current mode. Default welcome script and `load_script` action copy match.
> 
> **Local miniapps:** Restores **Starting &lt;name&gt;…** on glasses by calling `localDisplayManager.onMount` from `registerApp`, passing manifest **`name`** through the launcher into `InstalledMiniappManifest`. **`endBoot`** also clears stuck boot text on timeout when no core app is wired (today’s production path), with new `LocalDisplayManager` tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8816656a29ca93c443d75e6163078e943021448a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes Teleprompter Play to auto-scroll by default with a clear AI Scroll toggle and mic-aware caption. Restores and hardens the "Starting <miniapp>…" boot message for local miniapps, avoiding blank screens when an app never renders.

- **New Features**
  - Added a visible AI Scroll toggle in the Script view and a mic-aware caption that explains what Play will do now (including no-mic fallback).
  - Updated the default script and `load_script` action copy to match the new default.

- **Bug Fixes**
  - Defaulted `voiceFollow` to false so Play advances at the saved WPM unless AI Scroll is on.
  - Restored the boot message by calling `localDisplayManager.onMount(pkg, name)` and plumbing manifest `name`; on timeout, restore the pre-boot display instead of blanking (clear only if nothing valid to restore).

<sup>Written for commit 59d9d1bb82554376a25cfa3c32b299986f080883. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/miniapps/teleprompter/miniapp.json
+++ b/miniapps/teleprompter/miniapp.json
@@ -34,7 +34,7 @@
   "actions": [
     {
       "id": "load_script",
-      "description": "Load text into the teleprompter and start reading it on the glasses. Use when the user wants to put a speech, talking points, lines, or any notes up on their glasses — e.g. 'put my toast on the teleprompter' or 'teleprompt this'. Replaces the current script and resets to the top. By default it begins scrolling immediately (following the user's voice when a mic is available, otherwise at their saved words-per-minute); set autostart false to just show the opening lines so the user can start when ready.",
+      "description": "Load text into the teleprompter and start reading it on the glasses. Use when the user wants to put a speech, talking points, lines, or any notes up on their glasses — e.g. 'put my toast on the teleprompter' or 'teleprompt this'. Replaces the current script and resets to the top. By default it begins scrolling immediately at the user's saved speed (or following their voice if they've turned on AI Scroll); set autostart false to just show the opening lines so the user can start when ready.",
       "parameters": {
         "type": "object",
         "properties": {
--- a/miniapps/teleprompter/src/background/controllers/TeleprompterController.ts
+++ b/miniapps/teleprompter/src/background/controllers/TeleprompterController.ts
@@ -35,9 +35,9 @@ type On = <C extends keyof Channels & string>(channel: C, cb: (payload: Channels
 const DEFAULT_SCRIPT = [
   "Welcome to your teleprompter.",
   "",
-  "Paste or type your script here, put on your glasses, and press play. The words appear in front of you, line by line.",
+  "Paste or type your script here, put on your glasses, and press play. The words scroll by on their own at the speed you set, line by line.",
   "",
-  "Leave voice-follow on and the prompter listens — it advances as you speak, and waits when you pause. Turn it off to scroll at a steady pace you set in words per minute.",
+  "Want it to keep pace with you instead? Turn on AI Scroll and the prompter listens — it advances as you speak, and waits when you pause.",
   "",
   "That's it. Look up, speak naturally, and never lose your place.",
 ].join("\n")
@@ -47,7 +47,10 @@ const DEFAULT_SETTINGS: TeleprompterSettings = {
   wpm: 130,
   numberOfLines: 4,
   lineWidth: 2,
-  voiceFollow: true,
+  // Default to timed auto-scroll: pressing Play advances the script on its own.
+  // Voice-follow ("AI Scroll") is an explicit opt-in via the cockpit toggle —
+  // otherwise Play silently waits for speech and reads as "it doesn't work".
+  voiceFollow: false,
   autoRestart: false,
   showTimecode: false,
 }
--- a/miniapps/teleprompter/src/ui/App.tsx
+++ b/miniapps/teleprompter/src/ui/App.tsx
@@ -46,7 +46,6 @@ export function App() {
               onSetWpm={tp.setWpm}
               onSetLines={tp.setLines}
               onSetWidth={tp.setWidth}
-              onSetVoiceFollow={tp.setVoiceFollow}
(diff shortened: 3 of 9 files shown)


aisraelov:
Thanks @cursor @cubic-dev-ai — both P2s were valid and are fixed in 59d9d1bb8:

1. **Caption ignored the mic fallback** (`ScriptView.tsx`) — the transport caption now keys off the actual driver (`status.voiceMode`), not just the `voiceFollow` setting. It's a 3-way message so a mic-less device that falls back to timed scroll shows "AI Scroll is on, but this device has no mic — Play auto-scrolls at N wpm" instead of promising voice-follow. (Kept it off `voiceMode` rather than the naive `voiceMode ?? voiceFollow` swap, which would have produced a contradictory "turn on AI Scroll" while it's already on.)

2. **A second spawn could blank the foreground app** (`LocalDisplayManager` + `LocalMiniappRuntime`) — since `registerApp` fires `onMount` for every spawn (background app, crash-respawn), the boot flow is now non-destructive: `onMount` snapshots the current on-glasses frame, and on a boot that times out with nothing rendered we restore that frame instead of `sendClear()`-ing. We only clear when the glasses were empty before boot (or the saved frame's duration already elapsed). Added a test covering the two-app case.

Full core-vs-background arbitration (`onCoreAppChange`) is still intentionally unwired — separate workstream per `agents/local-display-manager-plan.md`; this keeps the boot message safe in the meantime.

PR #3337, opened 2026-07-03
Store blobs with mime-derived file extensions

aisraelov:
## Scope

Fixes silent Recorder miniapp playback on iOS (incident `2a974c5e-c217-4e9c-a413-3ea4cec66dc7`): the progress bar ran but no audio played.

## Root cause

`BlobStore` names on-disk files with a bare generated id (`mr46hl98-huglgjvzmt8e`). iOS AVPlayer resolves a local file's format from its path extension — no content sniffing — so the recorder's stored WAV handed to `session.speaker.play` never loaded. expo-audio surfaces no load error to JS, so playback "started" and then nothing: no finish event, no failure, for the full 28s captured in the incident logs. The share path had already hit the same wall and works around it by copying to a renamed temp file (`handleShare`); playback had no such workaround.

This is host responsibility, not the miniapp's: the docs promise a blob's `uri` can be handed to `session.speaker.play`, and the miniapp has no lever over the on-disk name.

## Change

- `commitFile` now attaches an extension derived from the declared `mimeType` (via new pure helper `withDiskExt` in `blobPaths.ts`), covering all three write paths: streamed writes, `setFromUrl`, `importFile`.
- Legacy extension-less blobs are left as-is (dev-only state, never shipped). Rename fails soft — the blob still stores/shares.
- Docs: `blob.mdx` example passed a bare uri to `speaker.play` instead of `{audioUrl}` (and "played" a PDF); fixed.

## Test evidence

- `bun test blobPaths.test.ts`: 16 pass (3 new `withDiskExt` cases).
- `tsc --noEmit` over mobile (island module included): clean.
- Not yet verified on iOS hardware — needs a record → play check on device; existing recordings made before this fix stay extension-less and will still be silent (delete + re-record).

## Follow-ups (not in this PR)

- `handlePlayAudio` accepts any `file://` path unvalidated — should be scoped to the caller's own blob dir.
- BlobStore's `getUserId` is wired to the raw coreToken JWT (`LocalMiniappRuntime.ts:2017`), putting a decodeable token prefix (sub + email) into blob paths and logs; should be a stable non-secret id.
- `AudioPlaybackService` has no load timeout, so iOS load failures strand the miniapp's `play()` promise forever.

---

> [!NOTE]
> **Medium Risk**
> Changes the on-disk naming path for all new blob commits; behavior is narrow and tested, but existing iOS recordings without extensions still won’t play until re-recorded.
> 
> **Overview**
> Fixes **silent local audio playback on iOS** when miniapps pass a blob `uri` to `session.speaker.play`: AVPlayer picks format from the **path extension**, so extension-less generated ids never loaded.
> 
> **`BlobStore.commitFile`** now renames the on-disk file via new **`withDiskExt`** (mime → extension through existing `extForMime`), covering streamed commits, `setFromUrl`, and `importFile`. Rename failures are ignored so storage/share still work; **legacy extension-less blobs are not migrated**.
> 
> Docs **`blob.mdx`** example corrected to import WAV and call **`speaker.play({ audioUrl })`** instead of a bare uri / PDF. **`blobPaths.test.ts`** adds `withDiskExt` cases. Unrelated **`tsconfig.json`** tweaks drop `rootDir` in two mobile modules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab06ecdd3e11d303ee22033ff9d05cd17b9e3783. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes silent Recorder playback on iOS by storing blobs with mime-derived file extensions so the OS detects audio format. Updates docs and tests; existing iOS recordings without extensions won’t play and should be re-recorded.

- **Bug Fixes**
  - Attach a mime-derived extension to on-disk blob names at commit across streamed writes, `setFromUrl`, and `importFile` via `withDiskExt`.
  - Keep legacy extension-less blobs as-is; share still works via temp rename.
  - Docs: use `{ audioUrl: picked.uri }` with `session.speaker.play` and correct example mime.

- **Dependencies**
  - Accept regenerated `expo-module-scripts` tsconfig output by removing explicit `rootDir` in module tsconfigs to avoid persistent dirty diffs.

<sup>Written for commit ab06ecdd3e11d303ee22033ff9d05cd17b9e3783. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3337?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mintlify-docs/app-devs/core-concepts/blob.mdx
+++ b/mintlify-docs/app-devs/core-concepts/blob.mdx
@@ -62,9 +62,9 @@ const cover = await session.blob.setFromUrl(
 );
 
 // OS file picker. Resolves to BlobMeta, or null if the user cancels.
-const picked = await session.blob.importFile({ mimeType: "application/pdf" });
+const picked = await session.blob.importFile({ mimeType: "audio/wav" });
 if (picked) {
-  await session.speaker.play(picked.uri);
+  await session.speaker.play({ audioUrl: picked.uri });
 }
 ```
 
--- a/mobile/modules/bluetooth-sdk/tsconfig.json
+++ b/mobile/modules/bluetooth-sdk/tsconfig.json
@@ -2,8 +2,7 @@
 {
   "extends": "expo-module-scripts/tsconfig.base",
   "compilerOptions": {
-    "outDir": "./build",
-    "rootDir": "./src"
+    "outDir": "./build"
   },
   "include": ["./src"],
   "exclude": ["**/__mocks__/*", "**/__tests__/*", "**/__rsc_tests__/*"]
--- a/mobile/modules/island/src/services/BlobStore.ts
+++ b/mobile/modules/island/src/services/BlobStore.ts
@@ -18,8 +18,10 @@
  *   bytes:    Paths.document/mentra_blobs/{userId}/{packageName}/{fileName}
  *   metadata: MMKV key  mentraos_blobmeta_{userId}_{packageName}_{key}  → StoredBlobMeta JSON
  *
- * The on-disk file name is a generated id; the logical `key` (caller-chosen) is
- * the MMKV key. This decouples the filesystem from arbitrary caller keys.
+ * The on-disk file name is a generated id plus a mime-derived extension (iOS
+ * AVPlayer keys a local file's format off the path extension, so blobs played
+ * via `session.speaker.play` must carry one); the logical `key` (caller-chosen)
+ * is the MMKV key. This decouples the filesystem from arbitrary caller keys.
  *
  * expo-file-system's new API is synchronous (create/writeBytes/md5/size), so
  * streaming chunk writes do no async work.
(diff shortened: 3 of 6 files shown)


aisraelov:
Second commit adds the two expo-module tsconfigs that `expo-module-scripts` regenerated during workspace install. Note: the regeneration strips the explicit `rootDir` that #3332 added two days ago — i.e. this partially reverts that PR. The generator will keep doing this on every fresh install, so if the `rootDir` override is still needed, it needs a home the generator honors; if CI (which runs the tsconfig-dependent module builds) stays green here, accepting the generated output is fine. Drop commit ab06ecdd3 if that's not the intent.

PR #3355, opened 2026-07-07
Add missing globals devDependency for root ESLint flat config

PhilippeFerreiraDeSousa:
## Problem

`eslint.config.mjs` imports `globals` (line 2), but it was never declared in the root `package.json`. It only existed in `bun.lock` as a **transitive** dependency of `eslint-config-expo` (which requires `globals@^16.0.0`). Bun doesn't reliably hoist transitive-only deps to the root `node_modules`, so on a fresh checkout:

```
cd mobile && bun lint:check
# ERR_MODULE_NOT_FOUND: Cannot find package 'globals' imported from eslint.config.mjs
```

Running ESLint from `mobile/` resolves up to the root flat config, so the missing dependency crashes it at config-load time before any file is linted.

## Fix

- Declare `globals` as a direct `devDependency`, pinned to `^16.3.0` to match the version `bun.lock` already resolves — the lockfile change is a single line promoting it to a direct dep, no version churn.
- Audited **all 10 imports** in `eslint.config.mjs` against the root `devDependencies`; `globals` was the only one missing.

## ESLint 9 `.eslintignore` migration

ESLint 9 deprecates `.eslintignore`. Migrated `mobile/.eslintignore` into the flat-config `ignores` block in `eslint.config.mjs` and deleted the file. Every entry was already covered by the root `ignores` except `**/*.log` and `mobile/.bundle/**`, which are added here (both are functional no-ops for ESLint — added only to keep the migration faithful). `cloud/.eslintignore` is left untouched; it belongs to cloud's own `cloud/eslint.config.mjs` and isn't on the root/mobile lint path.

## Verification

Ran `bun install` (root) + `cd mobile && bun install`, then `cd mobile && bun lint:check`:

- ✅ No `ERR_MODULE_NOT_FOUND` — config loads and `globals` resolves from root
- ✅ No `.eslintignore` deprecation warning
- ✅ ESLint starts up and produces a normal rule report (pre-existing findings), instead of crashing

Note: a fresh checkout needs **both** the root and `mobile/` installs before `bun lint:check` runs to completion — `mobile` is not a root workspace, so its deps (including `expo`, whose `tsconfig.base` the import resolver needs) are only installed by `cd mobile && bun install`.

---

> [!NOTE]
> **Low Risk**
> Tooling-only dependency and ignore-list changes; no runtime application or security-sensitive code paths.
> 
> **Overview**
> Fixes **fresh-checkout ESLint crashes** when lint runs from `mobile/` and resolves the root `eslint.config.mjs`, which imports `globals` but previously relied on it only as a transitive dependency of `eslint-config-expo`.
> 
> **`globals`** is now a direct root `devDependency` (`^16.3.0`, lockfile updated) so the package resolves reliably under Bun instead of failing with `ERR_MODULE_NOT_FOUND` at config load.
> 
> For **ESLint 9**, `mobile/.eslintignore` is removed and its patterns are folded into the flat config `ignores` block in `eslint.config.mjs`, including `**/*.log` and `mobile/.bundle/**` (the only entries not already covered).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58b89a28ca1e9fa657db46911eb78a224e9a057a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PhilippeFerreiraDeSousa:
Heads-up: this merged on a CI run from before #3331 landed, and the `@/stores/glasses` import doesn't exist post-#3331 — dev's mobile TypeScript check is failing at HEAD (blocking mobile CI on all open PRs). #3357 repairs it by moving the state-clearing behind the OTA facade (`toolkit.ota.installSession.discard()`); the Skip behavior itself is unchanged and your progress-screen tests pass as written.

PR #3396, opened 2026-07-10
Document Supabase deployment prerequisites for 019 account auth

PhilippeFerreiraDeSousa:
## Scope

Docs-only. Adds a "Deployment prerequisites (Supabase configuration)" subsection to `cloud-v2/docs/issues/019-mentra-account-auth/design.md` (end of §4, next to the OAuth flow description and callback URL).

Covers two prerequisites discovered the hard way after PR #3382 was merged, neither of which is enforced by code:

1. **`SUPABASE_URL` must point at the project's active Supabase custom domain, `https://auth.mentra.glass`.** The old `https://auth.augmentos.org` returns Cloudflare error 1014 "CNAME Cross-User Banned" — Supabase allows only one custom domain per project, and that hostname is no longer registered as it.
2. **The core OAuth callback must be in the Supabase Auth redirect allowlist** (dashboard → Authentication → URL Configuration → Redirect URLs): `https://<core public host>/api/account/oauth/callback` plus a `?*` glob variant, for every environment/origin running the flow — including per-developer Porter deployments (e.g. `porter.isaiah.yaml`). Missing entries make GoTrue silently fall back to the Site URL and the flow dead-ends before the core `/callback` leg.

Plus a note on how core builds the redirect URL: no public-origin env var is needed — #3401 makes `publicOrigin()` derive the origin from the request (`x-mentra-public-origin` when proxied, otherwise `x-forwarded-proto` + Host), replacing the never-set `CORE_PUBLIC_URL` fallback this subsection originally documented as a third prerequisite.

## Test evidence

None needed — markdown only, no code paths touched.

### DIFF
--- a/cloud-v2/docs/issues/019-mentra-account-auth/design.md
+++ b/cloud-v2/docs/issues/019-mentra-account-auth/design.md
@@ -98,8 +98,9 @@ app                core                        Supabase/provider
 
 ### Deployment prerequisites (Supabase configuration)
 
-Three requirements discovered the hard way after the merge (PR #3382); none
-is enforced by code, all fail at runtime if missed:
+Two requirements discovered the hard way after the merge (PR #3382); neither
+is enforced by code, both fail at runtime if missed — plus a note on how core
+builds the redirect URL:
 
 - **`SUPABASE_URL` (section 2) must point at the project's ACTIVE custom
   domain, `https://auth.mentra.glass`.** The old `https://auth.augmentos.org`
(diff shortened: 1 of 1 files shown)


PhilippeFerreiraDeSousa:
Pushed 46ec7419a: the third prerequisite (`CORE_PUBLIC_URL` must be set) is obsolete as of #3401, which derives the public origin from `x-forwarded-proto` + Host instead. Replaced that bullet with a note on the derived origin and reworded the allowlist bullet's host references; the two Supabase prerequisites are unchanged. The two PRs touch disjoint hunks of design.md, so they merge cleanly in either order.

PR #3412, opened 2026-07-11
Release @mentra/engine to npm

PhilippeFerreiraDeSousa:
## Summary

- add a version-gated npm release workflow for `@mentra/engine`, modeled on the Bluetooth SDK release pipeline
- detect both package version changes and initial workflow introduction so the first release works in either stacked-PR merge order
- bootstrap the brand-new package with the existing `NPM_TOKEN`, then use OIDC staged publishing for later versions
- skip versions already present on npm and support safe manual dry runs or forced release checks
- add npm publish metadata, an explicit package file allowlist, and installation instructions

## Stack

This PR is based on and should merge into #3411 first. It contains only the Engine npm release layer on top of the package/API rename.

## Release behavior

When these changes reach `dev`, the workflow builds and publishes `@mentra/engine@0.1.0`. npm does not allow staged publishing for a package that does not yet exist, so the first release uses `NPM_TOKEN`. After that initial release, configure the package's npm trusted publisher for `Mentra-Community/MentraOS` and `engine-release.yml` with `npm stage publish` permission; future versions follow the Bluetooth SDK's staged OIDC flow.

## Validation

- simulated initial, unchanged, and manual dry-run release decisions
- validated root and mobile frozen lockfiles
- built Bluetooth SDK, Crust, Miniapp, and Engine package outputs from a clean dependency install
- `npm publish --dry-run --access public`
- verified tarball roots and exclusion of test files
- Engine package tests: 220 passing
- Prettier check for workflow, release script, package manifest, and README
- YAML and Node syntax parsing
- `git diff --check`
- confirmed `@mentra/engine@0.1.0` is not yet present on npm

---

> [!NOTE]
> **Medium Risk**
> Touches release automation and npm publish paths (token bootstrap vs OIDC staging); mistakes could publish wrong versions or mis-route auth, though guards limit publishes to `dev` and skip existing versions.
> 
> **Overview**
> Adds an automated **npm release pipeline** for `@mentra/engine`, aligned with the Bluetooth SDK / miniapp release pattern: a detect job compares `mobile/modules/engine/package.json` on `dev` pushes and only auto-publishes when the prior version is known and the version bumped (**fail closed** on unknown `before` SHA, force-push, or first-time package).
> 
> The release workflow builds with Bun, checks the registry (treats only **E404** as “missing”), skips versions already on npm, bootstraps the **first** publish with `NPM_TOKEN`, and uses **OIDC `npm stage publish`** afterward. Manual `workflow_dispatch` supports dry-run and `force_release` for supervised first release.
> 
> Package manifest updates prepare publishing: `publishConfig`, npm `files` allowlist (excluding tests), repository/homepage metadata, README install/peer notes, and **`semver` moved from peer to direct dependency** (with `@types/semver` as dev-only); lockfiles updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 997f4b3a1ae4fa526aed3452fb3b2dba86dbc753. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/scripts/engine-release-info.mjs
+++ b/.github/scripts/engine-release-info.mjs
@@ -1,0 +1,102 @@
+#!/usr/bin/env node
+import {execFileSync} from "node:child_process"
+import {readFileSync} from "node:fs"
+
+const packagePath = "mobile/modules/engine/package.json"
+const workflowPath = ".github/workflows/engine-release.yml"
+
+const outputPath = process.env.GITHUB_OUTPUT
+const eventName = process.env.GITHUB_EVENT_NAME || ""
+const eventPath = process.env.GITHUB_EVENT_PATH
+const forceRelease = process.env.FORCE_RELEASE === "true"
+const dryRun = process.env.DRY_RUN === "true"
+
+function git(args) {
+  return execFileSync("git", args, {
+    encoding: "utf8",
+    stdio: ["ignore", "pipe", "pipe"],
+  }).trim()
+}
+
+function readJsonAt(ref, path) {
+  try {
+    return JSON.parse(git(["show", `${ref}:${path}`]))
+  } catch {
+    return null
+  }
+}
+
+function fileExistsAt(ref, path) {
+  try {
+    git(["cat-file", "-e", `${ref}:${path}`])
+    return true
+  } catch {
+    return false
+  }
+}
+
+function setOutput(name, value) {
+  if (!outputPath) {
+    console.log(`${name}=${value}`)
+    return
+  }
+  execFileSync("bash", ["-lc", 'cat >> "$GITHUB_OUTPUT"'], {
+    input: `${name}=${value}\n`,
+    env: process.env,
+  })
+}
+
+const currentPackage = JSON.parse(readFileSync(packagePath, "utf8"))
+const currentVersion = currentPackage.version
+
+if (!currentPackage.name) {
+  throw new Error(`Missing name in ${packagePath}`)
+}
+
+if (!currentVersion) {
+  throw new Error(`Missing version in ${packagePath}`)
+}
+
+let beforeSha = process.env.ENGINE_COMPARE_SHA || ""
+if (!beforeSha && eventPath) {
+  try {
+    const event = JSON.parse(readFileSync(eventPath, "utf8"))
+    beforeSha = event.before || ""
+  } catch {
+    beforeSha = ""
+  }
+}
+
+if (/^0+$/.test(beforeSha)) {
+  beforeSha = ""
+}
+
+const previousPackage = beforeSha ? readJsonAt(beforeSha, packagePath) : null
+const previousVersion = previousPackage?.version || ""
+// A missing package at the before-SHA is the initial release, so 0.1.0 is
+// published when the Engine package first lands on dev.
+const versionChanged = Boolean(beforeSha) && previousVersion !== currentVersion
+// Also run when this workflow first lands. That covers the stacked-PR merge
+// order where the Engine package reaches dev before its release workflow.
+const workflowAdded = Boolean(beforeSha) && !fileExistsAt(beforeSha, workflowPath)
+const runRelease = versionChanged || workflowAdded || forceRelease || (eventName === "workflow_dispatch" && dryRun)
+
+setOutput("package_name", currentPackage.name)
+setOutput("version", currentVersion)
+setOutput("previous_version", previousVersion)
+setOutput("compare_sha", beforeSha)
+setOutput("version_changed", String(versionChanged))
+setOutput("workflow_added", String(workflowAdded))
+setOutput("force_release", String(forceRelease))
+setOutput("dry_run", String(dryRun))
+setOutput("run_release", String(runRelease))
+
+console.log(`Mentra Engine package: ${currentPackage.name}`)
+console.log(`Current version: ${currentVersion}`)
+console.log(`Previous version: ${previousVersion || "(not present)"}`)
+console.log(`Compare SHA: ${beforeSha || "(unavailable)"}`)
+console.log(`Version changed: ${versionChanged}`)
+console.log(`Release workflow added: ${workflowAdded}`)
+console.log(`Force release: ${forceRelease}`)
+console.log(`Dry run: ${dryRun}`)
+console.log(`Run release job: ${runRelease}`)
--- a/.github/scripts/engine-release-info.mjs
+++ b/.github/scripts/engine-release-info.mjs
@@ -1,9 +1,8 @@
 #!/usr/bin/env node
 import {execFileSync} from "node:child_process"
-import {readFileSync} from "node:fs"
+import {appendFileSync, readFileSync} from "node:fs"
 
 const packagePath = "mobile/modules/engine/package.json"
-const workflowPath = ".github/workflows/engine-release.yml"
 
 const outputPath = process.env.GITHUB_OUTPUT
 const eventName = process.env.GITHUB_EVENT_NAME || ""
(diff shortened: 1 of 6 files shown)


aisraelov:
## Review — not ready to merge as-is

One merge-blocking issue: the moment this stack lands on dev, the workflow auto-publishes `@mentra/engine@0.1.0` (via the `workflowAdded` bootstrap), and that package is **uninstallable today** — four of its `peerDependencies` don't exist on npm.

### Bot comments

- **Codex P1 (peer deps)** — confirmed and worse than stated. Verified against the registry: `@mentra/crust`, `@mentra/cloud-client`, `@mentra/cloud-runtime`, and `@mentra/miniapp` are all 404 (only `@mentra/bluetooth-sdk` exists). npm ≥7 auto-installs peers, so `npm install @mentra/engine` hard-fails with E404 for every consumer — and even with `--legacy-peer-deps`, the shipped `src/` has runtime value imports of all four (e.g. `src/services/MiniappEngine.ts:15`, `src/services/CloudClientService.ts:16`).
- **Cursor Bugbot** — summary only, no inline findings; its stated risks are covered by findings 3–4 below.

### Findings (verified, most-severe first)

1. **`mobile/modules/engine/package.json` (peerDependencies) + `README.md` — publishing an uninstallable package, and merge = immediate publish.** Four mandatory peers are absent from npm; the install command the README advertises fails E404 on `@mentra/crust@*`. Because of the `workflowAdded` bootstrap, the publish fires automatically on the dev push that lands this workflow — burning 0.1.0 on a broken artifact (npm won't let the version number be reused). Either publish/stage the four peers first, or remove the auto-bootstrap and do the first publish via a supervised `force_release` dispatch once peers exist.

2. **`package.json` `main`/`default` entry doesn't load in any non-Metro consumer.** `expo-module build` emits ESM (`module: esnext`) with extensionless relative imports into a package with no `"type": "module"`, so `require("@mentra/engine")` dies with `SyntaxError: Cannot use import statement outside a module` on Node 18/20 and `ERR_MODULE_NOT_FOUND` on newer Node; ESM-only `typesafe-ts` is the next failure behind that. Only the `react-native`/Metro path works. Maybe acceptable for 0.1.0, but worth deciding consciously before the first public release — consumer Jest/Node tooling resolves the `default` condition.

3. **`.github/scripts/engine-release-info.mjs:78-81` — release detection fails open on git errors** (empirically verified). `readJsonAt()`/`fileExistsAt()` swallow all git failures, so an unresolvable `event.before` (e.g. force-push to dev that discards the old head — dev's branch protection currently allows force pushes) makes both `versionChanged` and `workflowAdded` true on a push that changed nothing, entering the real publish path. The npm version-exists check usually downgrades this to a wasted run, but it will publish whatever unpublished version is sitting in `package.json` (failed prior run, premature bump). The sibling bluetooth/miniapp scripts both fail closed (`hasPreviousVersion &&` gate); engine should too.

4. **`engine-release.yml` "Check npm registry" — conflates network failure with "package doesn't exist".** `npm view … >/dev/null 2>&1` exits 1 identically for 404 and a transient registry 5xx. A flake during a later release routes it down the `NPM_TOKEN` direct-publish path, silently bypassing the OIDC staged flow the header documents as steady state — or hard-fails with a misleading "set NPM_TOKEN" error if the token was rotated out after bootstrap. Parse `npm view --json` and match error code `E404` instead of relying on exit status.

5. **`package.json` — `semver` is a runtime dependency declared only as a peer, while `@types/semver` sits in prod `dependencies`.** `src/services/AppRegistry.ts:22` does a value `import semver from "semver"`. A consumer installing with `--legacy-peer-deps` (the workaround they'll reach for given finding 1) or yarn 1 gets "Unable to resolve module 'semver'" at runtime; meanwhile every consumer installs a types-only package into prod deps. Move `semver` to `dependencies`, `@types/semver` to `devDependencies`.

6. **Third near-verbatim fork of the release pipeline** (cleanup). 63 of 88 non-blank script lines are identical to `bluetooth-sdk-release-info.mjs`; the workflow's npm job is ~85% verbatim from `bluetooth-sdk-release.yml`. The copies have already diverged semantically (engine fails open where the siblings fail closed), and the repo already uses `workflow_call` reusables. A shared parameterized detect script/reusable job would keep the next edge-case fix from landing in only one copy.

7. **Copy-paste debris** (cleanup). `ENGINE_PACKAGE_JSON` env is defined but never referenced; five of nine detect-job outputs (`previous_version`, `compare_sha`, `version_changed`, `workflow_added`, `force_release`) are never consumed downstream; `setOutput()` shells out to `bash -lc 'cat >> "$GITHUB_OUTPUT"'` where `appendFileSync` does the same job; `--access public` duplicates `publishConfig.access`; `npm install -g npm@latest` is unpinned in the release-critical path.

8. **Detect-job overhead on every engine-touching dev push** (efficiency). `fetch-depth: 0` pulls full monorepo history (~1 GiB) when the script only reads two blobs at one SHA — a depth-1 checkout plus `git fetch --depth=1 origin $BEFORE_SHA` suffices. The root+mobile `bun install` preflight duplicates what the single downstream npm job does anyway, and unrelated root-lockfile drift can block an engine release from inside the detect job.

Smaller notes: a `workflow_dispatch` with `dry_run=false, force_release=false` is a silent no-op (worth stating in the input description that manual publishes need `force_release`); bare versions from dev land on the `latest` dist-tag while the "Refuse non-dev publish" guard means this workflow can never serve a future main-branch channel; the new `homepage` URL (`/tree/main/mobile/modules/engine`) 404s until the engine module reaches main.

### Bottom line

The release plumbing is a faithful adaptation of the bluetooth-sdk pipeline and most of its internal logic is correct. Hold the stack from reaching dev until the peer-dependency story is resolved — as written, merging publishes a package nobody can install, on the unrecallable 0.1.0 version number, from the very first push. Fixes 3–5 are small and worth folding in now; 6–8 are optional.

PhilippeFerreiraDeSousa:
@aisraelov All addressed in 997f4b3a1a — point by point:

**1 (blocker) — merge no longer publishes anything.** Took your second option: the `workflowAdded` bootstrap is removed entirely. Landing this stack on dev is inert; the first publish is a supervised `workflow_dispatch` with `force_release=true`, to be run only after `@mentra/crust`, `@mentra/cloud-client`, `@mentra/cloud-runtime`, and `@mentra/miniapp` exist on npm. The procedure is documented in the workflow header, and the `force_release` input description now states that a dispatch without it is a no-op (your smaller note). 0.1.0 stays unburned.

**2 — decided consciously, documented.** Metro/React Native-only consumption is accepted for 0.1.x: the README now states both limits explicitly (workspace consumption until the peers publish; the `default` condition's ESM output does not load under Node). Making the `default` entry Node-loadable is deferred until there's a concrete non-Metro consumer.

**3 — fails closed, empirically verified.** Same gate as the siblings (`hasPreviousVersion &&`); the script now fetches the before-SHA itself at depth 1, and I reproduced your scenario: an unresolvable `event.before` yields `Previous version: (unknown — releases fail closed) / Run release job: false`. Same-version and real-version-change cases verified too.

**4 — E404 is the only absence signal.** The registry check parses `npm view --json` error codes; any non-404 failure hard-fails the step with an explicit refusal to guess, so a flake can neither route a routine release down the `NPM_TOKEN` path nor produce the misleading token error.

**5 — fixed.** `semver` → `dependencies` (`^7.7.4`, matching mobile), `@types/semver` → `devDependencies`; lockfiles regenerated.

**7 + parts of 8** — debris removed (`ENGINE_PACKAGE_JSON`, five unused outputs, `bash -lc` setOutput, `--access public`, `npm@latest` → pinned `npm@11`), `fetch-depth: 0` → depth-1 + targeted before-SHA fetch, and both detect-job lockfile preflights dropped (the npm job's frozen mobile install is the real gate; root-lock drift shouldn't block an engine release). Also from your smaller notes: `homepage` now points at the dev tree.

**6 (shared release pipeline) — agreed, deferred.** With this PR the engine script is now byte-for-byte the siblings' *semantics* (fail-closed, same gate), which removes the divergence risk you called out; folding all three into a `workflow_call` reusable is queued as a follow-up rather than widening this PR into the bluetooth/miniapp pipelines. The dist-tag/main-channel note is noted for whenever a main-branch channel becomes real.

PR #3458, opened 2026-07-16
ASG: owner-keyed wake leases with lifecycle tracing; fresh awake window per W:1 command

PhilippeFerreiraDeSousa:
Two changes that together make the glasses' wake behavior correct and observable: a fresh awake window per wake-flagged phone command (the original scope), and a rework of `WakeLockManager` into **owner-keyed wake leases** after review surfaced that the shared-lock design lets any subsystem revoke any other's protection.

## Part 1 — a fresh MTK awake window per wake-flagged (W:1) command

Root cause of `sendWifiCredentials` results arriving at 16-25s in the field (measured on hardware):

- The glasses' screen timeout is **12s**; with no wakelock held, the MTK suspends when it expires.
- The BES wakes the MTK for wake-flagged commands by pulsing the power key — but **only when the SoC is already asleep** (`lxy_glass_force_wakeup_mtk` is GPIO-gated). A command landing mid-awake-window gets **no extra time**.
- The wifi flow's status machinery (`Thread.sleep` polls, debounce Handler) runs on clocks that **freeze across suspend**, and SDK heartbeats are `wakeup=false`, so a mid-flow suspend stalls the result until the next unrelated wake — the 16-25s tail.

Fix: every W:1 command (string-wrapper `"W":1` in `CommandProcessor`; binary `FLAG_WAKE` in `K900BluetoothManager`) grants a **15s CPU lease** (`AsgConstants.PHONE_WAKE_COMMAND_WINDOW_MS`, sized to outlive the wifi failure verdict at ~12.4s). W:0 traffic (30s heartbeats) grants nothing, so idle battery behavior is unchanged. The BES translates binary-wire FLAG_WAKE to `"W":1` when re-packing onto UART, so the CommandProcessor path covers both wire modes today.

## Part 2 — owner-keyed wake leases (review finding by @aisraelov)

The manager held ONE shared static `WakeLock` per kind. Extend-only protected against shorter *acquires*, but any subsystem's *release* — notably `releaseAllWakeLocks()`, called by camera close and every streaming cleanup — silently revoked every other subsystem's protection. Inventory of what could collide:

| Owner | CPU timeout | Release path (before) |
|---|---|---|
| Camera | 180s | `releaseAllWakeLocks()` ×2 |
| RTMP/SRT/WHIP streaming | 36 min | `releaseAllWakeLocks()` ×4 (WHIP before its final "stopped" notification) |
| BES OTA | 5 min | `releaseCpuWakeLock()` |
| MTK OTA | 10 min | timeout-only |
| Battery announce | 5s screen | timeout-only |
| W:1 commands (new) | 15s | timeout-only |
| Bootstrap | 60s (raw lock outside the manager) | manual |

Concrete self-collisions: `stop_video_recording` (W:1) → camera close releases all locks mid-upload; photo keep-alive close during webhook/BLE fallback; `stop_stream` cleanup killing a concurrent BES OTA's lock — a plausible contributor to the known "MTK sleeps mid-BES-update" stalls.

**Design:** one lease per owner (`PHONE_COMMAND, CAMERA, STREAMING, BES_OTA, MTK_OTA, BATTERY_ANNOUNCE, BOOTSTRAP`), each backed by its own `PowerManager.WakeLock`. The kernel keeps the CPU up while ANY lease is held and sleeps after the last one ends — the OS is the arbiter, no arbitration logic in the manager. `release(owner)` can only end that owner's leases; `releaseAllWakeLocks()` is **deleted** and each caller now releases exactly the lease it owns (which is what every call site meant). Acquires stay extend-only, per owner. Per-owner tags (`AugmentOS:Cpu:CAMERA`, …) make `dumpsys power` / batterystats attribution per-subsystem. `BootstrapActivity`'s raw lock joins the manager as `BOOTSTRAP`.

## Part 3 — BES OTA holds a screen lease, re-armed on segment progress

The 2026-07-08 field failure proved a CPU lease is **not sufficient** for the BES UART transfer: the transfer froze between segments at 80% with the 5-minute CPU lock still held (4:40 remaining) the moment the device's display-sleep transition fired ("Going to sleep due to timeout" — the vendor sleep hook wedges the transfer state machine). The vendor "screen on" state is the protection the transfer actually needs — the same one the streaming services already hold.

`BesOtaManager` now acquires **both** cpu and screen leases at transfer start (5-min initial window covering authorization + handshake) and re-arms them on confirmed segments (`BES_OTA_SEGMENT_LEASE_WINDOW_MS` = 120s rolling, at most every 60s): a live transfer never expires its protection regardless of total duration — closing the >5-minute exposure of the old one-shot lock — while a wedged transfer stops re-arming and lets the device sleep within the window, aligned with the phone's 120s stall watchdog. (The verify-wait timeout / breakpoint-resume recovery inside the transfer state machine is tracked separately.)

## Observability

Every lease transition rides the existing lifecycle trace channel (`BleTraceLogger.logLifecycle`, component `WakeLockManager` — same yellow channel as the service start events, flows into incident logs):

- `wake_acquire` / `wake_extend` — `{owner, kind, ms, held}`
- `wake_release` — with remaining ms at release
- `wake_expire` — deadline watcher (timeouts expire in the OS without callback)

`held` is the full set of currently-held leases, so "why did the MTK sleep at T?" is answerable from one grep: which lease died, when, and what was left holding.

## Test evidence

- `:app:compileDebugJavaWithJavac` clean; all 11 call sites ported, no references to the old API remain.
- Part 1 hardware-verified before the rework: W:1 → lock held (`dumpsys power`), auto-release at 15s, no lock for W:0 pings, wifi join over BLE at 3.3s e2e.
- **On-device verification of the lease manager (Mentra Live hardware, test starting from a sleeping device):**
  - Three concurrent per-owner kernel locks with per-owner tags (`AugmentOS:Cpu:CAMERA`, `AugmentOS:Screen:CAMERA`, `AugmentOS:Cpu:PHONE_COMMAND` in `dumpsys power`); PowerManagerService even attributes the wake to the owner (`Waking up ... details=AugmentOS:Screen:CAMERA`).
  - **Ownership isolation (the review's collision case):** camera close released its own leases while a W:1 command lease was active — the `wake_release owner=CAMERA` trace shows `held="PHONE_COMMAND:cpu"`, i.e. the command lease survived. Under the old shared lock this exact sequence revoked it.
  - W:1 lease duration kernel-verified twice via the wake-lock event history: `ACQ → REL` at exactly 15.01s, lock present in every 2s dumpsys sample in between.
  - Trace channel captured every transition (`wake_acquire` / `wake_release` / `wake_expire`) with the full held-lease set.
  - Still pending: a battery-powered (non-USB) wifi join for the field 16-25s tail, and a real BES OTA under the new progress-coupled leases.

## Notes for reviewers

- 15s per user-action-driven command is the battery cost of part 1; heartbeats unaffected.
- The SDK-side wrong-password/error surfacing is #3459/#3460; this PR is glasses-only and helps every app (manager included) with no phone-side changes.

---

> [!NOTE]
> **High Risk**
> Changes core power/awake behavior across OTA, camera, streaming, and BLE command handling; regressions could cause mid-operation sleep or extra battery drain, though the design directly targets known production stalls.
> 
> **Overview**
> Reworks **`WakeLockManager`** from one shared CPU/screen lock into **per-subsystem leases** (`WakeOwner`: camera, streaming, BES/MTK OTA, phone commands, bootstrap, etc.). Each owner holds its own `PowerManager.WakeLock`; **`release(owner)`** only drops that owner’s leases, and **`releaseAllWakeLocks()`** is removed so camera/stream cleanup can’t revoke another subsystem’s protection. Acquires stay **extend-only** per owner; lease transitions are traced via **`BleTraceLogger`** (`wake_acquire` / `wake_extend` / `wake_release` / `wake_expire` with the full `held` set).
> 
> **Wake-flagged phone commands** (`"W":1` in **`CommandProcessor`**, **`FLAG_WAKE`** on binary wire in **`K900BluetoothManager`**) now grant a **15s** CPU lease (`PHONE_WAKE_COMMAND_WINDOW_MS`) so follow-up work (e.g. wifi credential polls) isn’t cut off by the 12s screen timeout when the BES power pulse doesn’t extend an already-awake window. Binary wake grants fire on the first fragment and again when reassembly completes.
> 
> **BES OTA** holds **CPU + screen** leases (display sleep was wedging UART mid-transfer) and **re-arms** them on segment CRC confirms (`BES_OTA_SEGMENT_LEASE_WINDOW_MS`), with **`mLeaseGate`** so post-cleanup UART confirms can’t re-acquire orphaned leases. Boot, MTK OTA, camera, streaming, and battery-announce paths are ported to the new API; **`WakeLockManagerTest`** adds ownership-isolation coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d63bd10d8c5e7b1b292d293b122270bc47aea375. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
@@ -499,9 +499,12 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
         bWait4Confirm = false;
         // Progress-coupled lease: while segments keep confirming, the transfer's wake
         // protection never expires; a wedged transfer stops re-arming and the device may
-        // sleep once the window runs out.
+        // sleep once the window runs out. Gated on isBesOtaInProgress so a late UART
+        // segment-verify arriving after cleanup() cannot re-acquire leases nothing will
+        // ever release.
         long nowMs = android.os.SystemClock.elapsedRealtime();
-        if (nowMs - lastLeaseArmMs > AsgConstants.BES_OTA_SEGMENT_LEASE_WINDOW_MS / 2) {
+        if (isBesOtaInProgress
+                && nowMs - lastLeaseArmMs > AsgConstants.BES_OTA_SEGMENT_LEASE_WINDOW_MS / 2) {
             lastLeaseArmMs = nowMs;
             WakeLockManager.acquireFull(mContext, WakeLockManager.WakeOwner.BES_OTA,
                     AsgConstants.BES_OTA_SEGMENT_LEASE_WINDOW_MS,
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
@@ -9,6 +9,7 @@ import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.SerialPor
 import com.mentra.asg_client.io.bluetooth.utils.ByteUtil;
 import com.mentra.asg_client.io.ota.interfaces.IBesOtaController;
 import com.mentra.asg_client.service.core.handlers.K900CommandHandler;
+import com.mentra.asg_client.AsgConstants;
 import com.mentra.asg_client.utils.WakeLockManager;
 import java.io.File;
 import java.io.FileInputStream;
@@ -254,9 +255,15 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
             return false;
         }
 
-        // Acquire wakelock to prevent CPU sleep during firmware transfer
-        WakeLockManager.acquireCpu(mContext, WakeLockManager.WakeOwner.BES_OTA, WAKELOCK_TIMEOUT_MS);
-        Log.i(TAG, "BES OTA wakelock acquired for " + WAKELOCK_TIMEOUT_MS + "ms");
+        // The transfer needs the vendor "screen on" state, not just CPU: the display-sleep
+        // hook wedges the UART transfer state machine mid-flight even with a CPU lock held
+        // (2026-07-08 incident, frozen between segments at 80%). Hold BOTH leases; the
+        // initial window covers authorization + handshake, then confirmed segments re-arm
+        // them (see crc32ConfirmSuccess).
+        WakeLockManager.acquireFull(mContext, WakeLockManager.WakeOwner.BES_OTA,
+                WAKELOCK_TIMEOUT_MS, WAKELOCK_TIMEOUT_MS);
+        lastLeaseArmMs = android.os.SystemClock.elapsedRealtime();
+        Log.i(TAG, "BES OTA cpu+screen leases acquired for " + WAKELOCK_TIMEOUT_MS + "ms");
 
         // Set waiting for authorization flag (NOT in OTA mode yet!)
         isWaitingForAuthorization = true;
@@ -482,10 +489,24 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
         sentPos += size;
     }
 
+    // Deadline of the last lease re-arm; segment confirms arrive ~1/s, so re-arm at most
+    // every half window to avoid per-segment lock churn and wake_extend trace spam.
+    private long lastLeaseArmMs = 0;
+
     public void crc32ConfirmSuccess() {
         confirmTimes++;
         confirmSentPos = sentPos;
         bWait4Confirm = false;
+        // Progress-coupled lease: while segments keep confirming, the transfer's wake
+        // protection never expires; a wedged transfer stops re-arming and the device may
+        // sleep once the window runs out.
+        long nowMs = android.os.SystemClock.elapsedRealtime();
+        if (nowMs - lastLeaseArmMs > AsgConstants.BES_OTA_SEGMENT_LEASE_WINDOW_MS / 2) {
+            lastLeaseArmMs = nowMs;
+            WakeLockManager.acquireFull(mContext, WakeLockManager.WakeOwner.BES_OTA,
+                    AsgConstants.BES_OTA_SEGMENT_LEASE_WINDOW_MS,
+                    AsgConstants.BES_OTA_SEGMENT_LEASE_WINDOW_MS);
+        }
         Log.i(
                 TAG,
                 "✅ CRC32 verification successful - confirmTimes="
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
@@ -255,7 +255,7 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
         }
 
         // Acquire wakelock to prevent CPU sleep during firmware transfer
-        WakeLockManager.acquireCpuWakeLock(mContext, WAKELOCK_TIMEOUT_MS);
+        WakeLockManager.acquireCpu(mContext, WakeLockManager.WakeOwner.BES_OTA, WAKELOCK_TIMEOUT_MS);
         Log.i(TAG, "BES OTA wakelock acquired for " + WAKELOCK_TIMEOUT_MS + "ms");
 
         // Set waiting for authorization flag (NOT in OTA mode yet!)
@@ -327,7 +327,7 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
         operationStartTime = 0;
 
         // Release wakelock
-        WakeLockManager.releaseCpuWakeLock();
+        WakeLockManager.release(WakeLockManager.WakeOwner.BES_OTA);
         Log.i(TAG, "BES OTA wakelock released");
 
         isBesOtaInProgress = false;
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
@@ -335,11 +335,13 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
 
         operationStartTime = 0;
 
-        // Release wakelock
-        WakeLockManager.release(WakeLockManager.WakeOwner.BES_OTA);
+        // Flag-clear and lease release are one atomic step against the segment-confirm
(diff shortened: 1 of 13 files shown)


aisraelov:
I found one ownership edge case with the new 15s command grant.

The extend-only behavior works when a shorter acquire arrives: if camera, streaming, or OTA already holds a longer CPU lock, the 15s acquire is correctly ignored. The problem is that every subsystem currently shares the single static sCpuWakeLock. A later releaseCpuWakeLock() or releaseAllWakeLocks() from the original owner releases that sole lock, including the effective protection the W:1 command was relying on.

There are concrete self-collision paths:

- stop_video_recording arrives with W:1 while the camera 180s lock is active, so the 15s acquire is ignored. Video shutdown then calls closeCamera(), which releases all wake locks before the queued integrity check and optional upload finish.
- take_photo extends the shared lock through the camera, but the normal 3s post-capture keep-alive closes the camera and releases all locks while webhook upload or BLE fallback may still be running.
- stop_stream arrives while the long streaming lock is active, then RTMP/SRT/WHIP cleanup releases all locks. WHIP releases before emitting its final stopped notification.
- Any W:1 command, including set_wifi_credentials, can overlap an asynchronous camera keep-alive expiry, stream failure/timeout, or BES OTA cleanup and lose its window early.

I think the smallest safe fix for this PR is a dedicated phone-command CPU wake lock with its own stored WakeLock and deadline, e.g. acquirePhoneCommandWakeLock(...), tagged AugmentOS:PhoneCommandWakeLock. Camera/stream/OTA cleanup should not release that lock; it can expire via the existing 15s timeout. Longer term, WakeLockManager could key independent locks by owner (PHONE_COMMAND, CAMERA, STREAMING, BES_OTA, MTK_OTA), so release(owner) cannot cancel another subsystem lease.

Changing only the Android wake-lock tag would not isolate ownership because the manager would still retain just one shared WakeLock object.

PhilippeFerreiraDeSousa:
Went with the structural fix rather than the dedicated phone-command lock: `WakeLockManager` is now owner-keyed (`5a1b4bd`) — one lease per owner (`PHONE_COMMAND, CAMERA, STREAMING, BES_OTA, MTK_OTA, BATTERY_ANNOUNCE, BOOTSTRAP`), each backed by its own `WakeLock`, so the kernel arbitrates and `release(owner)` structurally cannot revoke another subsystem's lease. `releaseAllWakeLocks()` is deleted; all 11 call sites (including your four collision paths and `BootstrapActivity`'s raw lock) now release exactly what they own. Every lease transition (`wake_acquire/extend/release/expire`) is emitted on the lifecycle trace channel with the full held-lease set, and per-owner tags make dumpsys/batterystats attribution per-subsystem. PR description rewritten with the full inventory — details there.

asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
```diff
@@ -482,10 +491,27 @@ public void addSentSize(int size) {
...
         confirmTimes++;
         confirmSentPos = sentPos;
         bWait4Confirm = false;
+        // Progress-coupled lease: while segments keep confirming, the transfer's wake
+        // protection never expires; a wedged transfer stops re-arming and the device may
+        // sleep once the window runs out. Gated on isBesOtaInProgress so a late UART
+        // segment-verify arriving after cleanup() cannot re-acquire leases nothing will
+        // ever release.
+        long nowMs = android.os.SystemClock.elapsedRealtime();
+        if (isBesOtaInProgress
```
aisraelov: The `isBesOtaInProgress` check is not atomic with cleanup. A UART callback can read `true` here, pause, then `cleanup()` can release the leases and clear state; when the callback resumes it reaches `acquireFull()` and recreates both leases for 120 seconds with no active update left to release them. Please serialize cleanup and re-arming, or use an OTA generation token, so a stale segment confirmation cannot acquire after cleanup.

asg_client/app/src/main/java/com/mentra/asg_client/utils/WakeLockManager.java
```diff
@@ -3,326 +3,288 @@
...
-     */
-    public static boolean releaseScreenWakeLock() {
-        synchronized (LOCK) {
-            try {
-                if (sScreenWakeLock != null && sScreenWakeLock.isHeld()) {
-                    sScreenWakeLock.release();
-                    Log.d(TAG, "Screen wake lock released");
+    private static void scheduleExpiryCheck(WakeOwner owner, boolean screen, long deadlineMs) {
+        long delayMs = Math.max(0, deadlineMs - SystemClock.elapsedRealtime()) + 250;
+        sExpiryHandler.postDelayed(() -> {
```
aisraelov, line 230: This callback is scheduled on the main looper 250 ms after the kernel lock deadline. If the expiring CPU lease is the last partial wake lock, the device can suspend before this callback runs, so `wake_expire` is delayed until the next wake and may be suppressed by a later reacquire. That makes the trace unreliable for the W:1 expiration this PR is intended to diagnose. Please record the absolute deadline in the acquire/extend event and treat later expiry detection as observational, or use a mechanism that does not depend on the main looper after dropping the final lock.

aisraelov (commented):
Two follow-up findings from manual review:

asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
PhilippeFerreiraDeSousa: Fixed in 9d7a44d with a lease gate: the segment-confirm re-arm (flag check + acquire) and cleanup's flag-clear + release are now two critical sections on the same monitor, so a confirm that read `isBesOtaInProgress == true` either completes its acquire before cleanup releases (and cleanup then releases what it acquired), or observes `false` and skips. Went with the monitor over a generation token — same guarantee, less state.

asg_client/app/src/main/java/com/mentra/asg_client/utils/WakeLockManager.java
PhilippeFerreiraDeSousa: Agreed the watcher can't be trusted across a suspend — fixed in 9d7a44d exactly along the lines you suggested: `wake_acquire`/`wake_extend` now advertise the authoritative `deadlineWallMs` (so the expiry moment is derivable from the grant event alone, and matches logcat's wall-clock timestamps), the watcher is documented as observational, and `wake_expire` carries `lateMs` so a post-suspend late fire is self-describing rather than misleading. Deliberately did not switch to an AlarmManager wakeup: waking the SoC just to log an expiry would spend battery on observability — for the W:1 diagnosis, grant-time deadline + optional late confirmation is sufficient.

aisraelov (approved):

PR #3462, opened 2026-07-16
Bluetooth SDK: harden BLE wire v2 negotiation triggers and retry

PhilippeFerreiraDeSousa:
## Background

While investigating wifi latency on Mentra Live with live traces (example app + BES 17.26.7.20), we mapped the full BLE wire v2 negotiation and found it works — the BES answers the phone's handshake itself and translates v2 binary frames to K900 string frames on the UART leg — but the SDK-side triggering is fragile in ways that fail **silently**: a session that misses its one negotiation window stays on the legacy string wire with no log or status anywhere.

## Three gaps, three fixes (Android + iOS, which mirror each other)

**1. `wire_caps` arriving in `version_info_3` were never parsed.**
The flexible `version_info*` parser only called `parsePeerWireCaps` inside its `build_number` branch — but the glasses put `wire_caps` in `version_info_3` (firmware chunk) while `build_number` travels in `version_info_1`. The caps that travel with the version exchange were silently discarded, leaving `glasses_ready` as the only real caps source. Now every `version_info` chunk parses caps and attempts the handshake; `parsePeerWireCaps` no-ops without a `wire_caps` key and `maybeSendWireHandshake` gates on caps + build + not-yet-queued, so per-chunk calls are idempotent.

**2. `glasses_ready` parsed caps but never attempted the handshake.**
If the build number is already known when caps (re-)arrive — e.g. the asg service restarts mid-session after an OTA install or a crash, and `glasses_ready` re-advertises the caps — no trigger was left to negotiate. `glasses_ready` now calls `maybeSendWireHandshake()` after parsing caps.

**3. The handshake was one-shot with no retry.**
`wireHandshakeQueued` was cleared only on success or full BLE disconnect, and the handshake + reply are fire-and-forget binary frames with **no ACK tracking** — one lost frame stranded the session on the legacy wire forever, silently. The flag now re-arms after a 5s grace period and retries, capped at **3 automatic attempts per session** (counter resets on disconnect and on activation) so a peer that advertises binary but never answers doesn't get pinged forever. Each re-arm logs, so a stuck negotiation is finally visible.

## Impact

Legacy string framing is fully functional (it's what v39-era pairings use), so these gaps cost the v2 wins (compact JSON framing, LE lengths) rather than correctness — which is exactly why they've been invisible. Fixing the triggers makes negotiation robust to chunk ordering, service restarts, and single lost frames.

## Test evidence

- `scripts/check-android-compile.sh bluetooth-sdk` passes.
- Gap 1 and the silent-stall behavior were observed on hardware during the investigation (glasses advertising `binary:true, proto:2` in `version_info_3`, phone never negotiating within that exchange). iOS mirrors the Android logic line-for-line and needs a CI/Xcode pass.
- Suggested on-device check: connect the example app to July-firmware glasses and confirm `BLE_TRACE ... proto=v2` wire metrics in the Console tab after the version exchange, then restart the asg service mid-session (adb reinstall) and confirm renegotiation on the next `glasses_ready`.

## Notes for reviewers

- One of several independent pick-and-choose PRs from the wifi/OTA latency investigation (#3458, #3459, #3460, #3461). Lowest urgency of the set.

---

> [!NOTE]
> **Medium Risk**
> Touches core BLE framing negotiation and send-queue lifecycle on both platforms; mistakes could affect protocol selection or drop in-flight writes, though legacy wire remains the fallback.
> 
> **Overview**
> **Hardens Mentra Live BLE wire v2 negotiation** on Android and iOS so sessions are less likely to stay on the legacy string wire without any visible failure.
> 
> **Negotiation triggers:** `wire_caps` are now parsed and `maybeSendWireHandshake()` runs on **every** `version_info*` chunk (not only when `build_number` arrives), fixing caps that live in `version_info_3` while build number is in `version_info_1`. **`glasses_ready`** is treated as a remote wire-session reset: negotiation state is cleared, caps are re-parsed, and the handshake is attempted again after mid-link ASG restarts.
> 
> **Retries and session safety:** Unanswered v2 handshakes **re-arm after 5s**, up to **3 attempts per session**, with logging when re-arming. A shared **`resetWireNegotiationState()`** bumps **`wireSessionGeneration`** and clears v2/caps state on disconnect, kill, and `glasses_ready`; handshake replies from a prior epoch are ignored. **Outbound queues** are cleared on disconnect (Android `sendQueue`; iOS `commandQueue.removeAll()`), and iOS queued writes carry a **generation** stamp so stale payloads cannot drain into a new session.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4dec86b6432bc1fae311d338fc7b96c6588cf2d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt
@@ -788,16 +788,12 @@ class MentraLive : SGCManager() {
         val isEqual = state == connectionState
         if (isEqual) {
             if (state == ConnTypes.DISCONNECTED) {
-                incomingChunkReassembler.clear()
-                peerWireProtocolVersion = 0
-                useBinaryWireProtocol = false
-                wireHandshakeQueued = false
-                wireHandshakeAttempts = 0
-                wireSessionGeneration++
-                peerK900Le = false
-                peerWireCapsBinary = false
-                peerFilePayloadV2 = false
-                BleJsonCompact.resetSession()
+                resetWireNegotiationState()
+                // Queued writes are session-bound: transmitting them into the NEXT session
+                // (e.g. a stale handshake whose reply activates v2 before the new session
+                // negotiated) is the bug class this clear removes. Higher layers re-send
+                // what still matters via their own ACK/retry tracking.
+                sendQueue.clear()
             }
             return
         }
@@ -825,16 +821,9 @@ class MentraLive : SGCManager() {
             DeviceStore.apply("glasses", "connected", false)
             DeviceStore.apply("glasses", "signalStrength", -1)
             DeviceStore.apply("glasses", "signalStrengthUpdatedAt", 0L)
-            incomingChunkReassembler.clear()
-            peerWireProtocolVersion = 0
-            useBinaryWireProtocol = false
-            wireHandshakeQueued = false
-            wireHandshakeAttempts = 0
-            wireSessionGeneration++
-            peerK900Le = false
-            peerWireCapsBinary = false
-            peerFilePayloadV2 = false
-            BleJsonCompact.resetSession()
+            resetWireNegotiationState()
+            sendQueue.clear() // see the disconnect reset above: stale writes die with the session
+
             // Drop OTA caches when fully disconnected — avoids leaking session/step state
             // from a previous pairing into the next one.
             resetOtaCache()
@@ -3504,10 +3493,12 @@ class MentraLive : SGCManager() {
                 // Glasses SOC has booted and is ready for communication
                 Bridge.log("LIVE: 🎉 Received glasses_ready message - SOC is booted and ready!")
 
-                // Negotiate wire capabilities advertised in the glasses_ready handshake.
-                // Also attempt the v2 handshake here: when the glasses' service restarts
-                // mid-session (OTA install, crash restart) the build number is already
-                // known, and glasses_ready is the message that re-advertises the caps.
+                // glasses_ready is a REMOTE wire-session reset: the glasses ran
+                // onTransportReset() before sending it, so their side is back on legacy
+                // framing regardless of what this side negotiated earlier. Start a fresh
+                // wire epoch (clears v2-active state that would otherwise gate the
+                // handshake off), then negotiate from the caps this message advertises.
+                resetWireNegotiationState()
                 parsePeerWireCaps(json)
                 maybeSendWireHandshake()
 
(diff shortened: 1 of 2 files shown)


mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt
```diff
@@ -3493,7 +3505,11 @@ class MentraLive : SGCManager() {
                 Bridge.log("LIVE: 🎉 Received glasses_ready message - SOC is booted and ready!")
 
                 // Negotiate wire capabilities advertised in the glasses_ready handshake.
+                // Also attempt the v2 handshake here: when the glasses' service restarts
+                // mid-session (OTA install, crash restart) the build number is already
+                // known, and glasses_ready is the message that re-advertises the caps.
                 parsePeerWireCaps(json)
+                maybeSendWireHandshake()
```
aisraelov, line 3508: **[P2] Treat `glasses_ready` as a remote wire-session reset.** The peer calls `onTransportReset()` before sending `glasses_ready`, which resets its binary state to v1. If the phone already has `peerWireProtocolVersion >= v2`, this call immediately returns in `maybeSendWireHandshake()`, because local state is only cleared on BLE disconnect. A mid-link ASG restart or duplicate readiness exchange therefore leaves the peer outbound path on legacy framing while the phone incorrectly considers v2 active, so the restart-recovery path described by this PR still does not work. Please force or re-arm a bounded handshake for a new `glasses_ready` epoch, and mirror it on iOS, instead of relying on the existing active-state guard.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt
```diff
@@ -7234,6 +7258,37 @@ class MentraLive : SGCManager() {
...
+                                            WIRE_HANDSHAKE_RETRY_GRACE_MS +
+                                            "ms (attempt " +
+                                            wireHandshakeAttempts +
+                                            "/" +
+                                            WIRE_HANDSHAKE_MAX_ATTEMPTS +
+                                            ") - re-arming"
+                            )
+                            wireHandshakeQueued = false
+                            if (wireHandshakeAttempts < WIRE_HANDSHAKE_MAX_ATTEMPTS) {
+                                maybeSendWireHandshake()
```
aisraelov, line 7305: **[P2] Bind queued retry frames to the session generation.** The generation check only protects this delayed callback. Once it calls `maybeSendWireHandshake()`, the retry frame is placed in the Android `sendQueue` without a generation. The normal GATT disconnect path removes `processSendQueueRunnable` but does not clear `sendQueue`; that only happens in `destroy()`. A disconnect between enqueue and dequeue therefore leaves this old handshake to be transmitted after reconnect, and its reply can activate v2 before the new session has established its build and capabilities. Please tag queued handshake writes with `wireSessionGeneration` and discard stale entries, or clear session-bound writes on disconnect.

aisraelov (changes requested):
Requesting changes for two session-boundary issues in the negotiation retry/restart flow. The ordering fix for `version_info*` looks correct, and both the current head and its merge with current `dev` compile successfully; these two state transitions remain unsafe.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt
PhilippeFerreiraDeSousa: Fixed in fb268a0 — went further than re-arming: every `glasses_ready` now starts a fresh wire epoch via the extracted `resetWireNegotiationState()` (same helper the disconnect resets use): negotiated state cleared, session generation bumped so old scheduled callbacks stand down, then caps from THIS message are parsed and the handshake re-attempts, bounded by the per-session attempt cap. A mid-link ASG restart or duplicate readiness exchange now converges back to v2 instead of wedging on stale v2-active state. Mirrored on iOS.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt
PhilippeFerreiraDeSousa: Fixed in fb268a0 with your second option, applied broadly: the disconnect reset now clears the send queue on both platforms (Android `sendQueue.clear()`, iOS `CommandQueue.removeAll()`). Queued writes are session-bound by nature — transmitting any of them into the next session is the same bug class as the stale handshake, so dropping them all on disconnect is both simpler and stricter than generation-tagging just the handshake frames. Higher layers re-send what still matters through their own ACK/retry tracking. The epoch reset from the other thread additionally bumps the generation on every `glasses_ready`, so even a frame that slipped through would face a fresh negotiation rather than activating v2 silently.

aisraelov (approved):

PR #3469, opened 2026-07-16
Allow a per-call timeout override on CameraModule.takePhoto()

isaiahb:
SDK-side plumbing for OS-1701 (sub-issue of OS-1687).

## Problem

`session.sendRequest()` defaults to a 60s timeout for every request type, including photo capture. A caller that speculatively fires a capture and may abandon it before it resolves (e.g. a miniapp racing a capture against a short latency cap for non-visual queries) has no way to bound its own wait: the abandoned `takePhoto()` promise sits against the full 60s ceiling.

## Scope (what this does and does not fix)

This change bounds only the miniapp-side promise. When the shorter timeout fires, `sendRequest()` removes the local pending entry and rejects; nothing is sent to the host, and the host-side capture pipeline (`LocalMiniappRuntime.handlePhoto()` awaiting `PhonePhotoCoordinator.takePhoto()`) keeps running under its own timeouts until it completes or expires.

The cross-request stall originally observed alongside this (an unrelated `location.getOnce()` freezing behind an in-flight capture) turned out to be the Android Expo-queue head-of-line blocking fixed in #3474, not the abandoned promise itself. Post #3474 the JS layers are concurrent: `handlePhoto` dispatch is fire-and-forget and the coordinator tracks requests in a requestId-keyed map, so an abandoned capture holds only its own coordinator entry, its cloud waiter, and the glasses camera, which arbitrates its own request queue on-device.

Propagating a correlated cancel to the host so an abandoned capture also releases those host-side resources is deliberately out of scope for this PR and tracked as a follow-up (PHOTO_CANCEL, see OS-1714 review discussion).

## Change

`TakePhotoOptions` gains an optional `timeoutMs`, passed through to `sendRequest()`. Unset (the default for a normal user-triggered capture) keeps the existing 60s ceiling. Scoped to the photo-capture path only; no other request type's default changes. Unit tests assert the timeout is forwarded as the second `sendRequest` argument and that the options argument is omitted when unset.

## Consumer

The Mentra AI miniapp's `PhotoManager.takePhoto()` will pass a bounded `timeoutMs` for its speculative non-visual capture once this lands.

---

> [!NOTE]
> **Low Risk**
> Backward-compatible optional API on photo capture only; default timeout path is unchanged.
> 
> **Overview**
> Adds optional **`timeoutMs`** on **`TakePhotoOptions`** so **`takePhoto()`** can pass a custom timeout into **`session.sendRequest()`** instead of always using the default **60s** ceiling.
> 
> When **`timeoutMs`** is omitted, behavior is unchanged (no second argument to **`sendRequest`**). When set, only that capture uses the override—intended for speculative captures that may be abandoned so the promise settles quickly and does not block other in-flight requests on the shared transport.
> 
> Tests cover forwarding **`timeoutMs`** in the options argument and passing **`undefined`** when it is unset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85a0206d61c88b8ceb37500aead7dbf2ee648713. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Add a per-call timeout override to `CameraModule.takePhoto()` to let speculative captures fail fast and avoid blocking other requests. Keeps the 60s default for normal captures and fixes OS-1701’s pipeline stall.

- **New Features**
  - Added optional `timeoutMs` to `TakePhotoOptions`; forwarded as the `sendRequest` options arg and omitted when unset.
  - Default 60s unchanged; photo-only; enables the Mentra AI miniapp to cap abandoned captures (~1–2.5s) per OS-1701.
  - Tests now record `sendRequest` options and assert both forwarding and omission to prevent regressions.

<sup>Written for commit 85a0206d61c88b8ceb37500aead7dbf2ee648713. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3469?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mobile/modules/miniapp/src/modules/camera.test.ts
+++ b/mobile/modules/miniapp/src/modules/camera.test.ts
@@ -8,18 +8,20 @@ import {CameraModule, type CameraFovResult, type PhotoTaken} from "./camera"
 
 function mockSession<T>(result: T) {
   const requestCalls: object[] = []
+  const requestOptions: (object | undefined)[] = []
   const session = {
     _hasManifestPermission: () => true,
     sendOneShot: () => {
       throw new Error("setFov should use sendRequest")
     },
-    sendRequest: (payload: object) => {
+    sendRequest: (payload: object, options?: object) => {
       requestCalls.push(payload)
+      requestOptions.push(options)
       return Promise.resolve(result)
     },
   } as unknown as MiniappSession
 
-  return {session, requestCalls}
+  return {session, requestCalls, requestOptions}
 }
 
 describe("CameraModule", () => {
@@ -74,6 +76,24 @@ describe("CameraModule", () => {
     await camera.takePhoto({mode: "text"})
 
     expect(requestCalls[0]).toMatchObject({mode: "text"})
+  })
+
+  test("takePhoto forwards timeoutMs as the sendRequest options argument", async () => {
+    const {session, requestOptions} = mockSession({})
+    const camera = new CameraModule(session)
+
+    await camera.takePhoto({timeoutMs: 1_200})
+
+    expect(requestOptions).toEqual([{timeoutMs: 1_200}])
+  })
+
+  test("takePhoto omits the options argument when timeoutMs is unset", async () => {
+    const {session, requestOptions} = mockSession({})
+    const camera = new CameraModule(session)
+
+    await camera.takePhoto({})
+
+    expect(requestOptions).toEqual([undefined])
   })
 
   test("warmUp forwards size and default duration", async () => {
--- a/mobile/modules/miniapp/src/modules/camera.ts
+++ b/mobile/modules/miniapp/src/modules/camera.ts
@@ -42,6 +42,16 @@ export interface TakePhotoOptions {
    * manual exposure; ignored otherwise.
    */
   exposureTimeNs?: number
+  /**
+   * Override the SDK's default 60s request timeout for this capture only.
+   * Leave unset for a normal user-triggered capture (the full 60s ceiling is
+   * appropriate there). Callers that speculatively fire a capture and may
+   * abandon it if the result isn't needed (e.g. a short-lived non-visual
+   * assistant turn racing a capture against a ~1s latency cap) should pass a
+   * short value here so the abandoned `takePhoto()` promise settles quickly
+   * instead of sitting on the default 60s ceiling.
+   */
+  timeoutMs?: number
 }
 
 export interface PhotoTaken {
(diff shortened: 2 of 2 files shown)


aisraelov:
Blocking concern: `timeoutMs` currently only bounds the miniapp-side promise; it does not cancel or shorten the phone-side photo operation.

`CameraModule.takePhoto()` passes the value as the second argument to `session.sendRequest()`. When that timer fires, `sendRequest()` only removes the entry from `pendingRequests` and rejects locally. The already-dispatched `LocalMiniappRuntime.handlePhoto()` continues awaiting `PhonePhotoCoordinator.takePhoto()`, whose BLE/cloud pipeline remains active under its own 30–45s timeouts. No deadline or cancellation is sent to the runtime/coordinator.

That means an abandoned speculative capture still owns the host-side request state and camera pipeline. If that in-flight host work is what causes the location stall described in this PR, this change cannot release it; it only causes the caller to stop waiting for its result.

Please propagate a correlated cancellation/deadline to the phone runtime and have `PhonePhotoCoordinator` release its active request and cloud waiter when the SDK timeout fires. A focused regression test should start a hanging photo, trigger the short timeout/cancel, verify the coordinator no longer owns the request, and verify a subsequent unrelated request can complete promptly. The camera module test should also assert the second `sendRequest` argument, since the current mock records only the payload and would pass if timeout forwarding were removed.

isaiahb:
The mechanics you describe are right: the timeout only settles the miniapp-side promise, nothing is sent to the host, and the coordinator pipeline keeps running. The stall framing was stale though; the cross-request freeze turned out to be the Android Expo-queue head-of-line blocking fixed in #3474, and post-#3474 an abandoned capture holds only its own coordinator entry and cloud waiter (the glasses arbitrate their own camera queue). I've reframed the PR description to claim only the promise bound, added the sendRequest-args tests you asked for (85a0206d6), and filed OS-1723 for the correlated PHOTO_CANCEL so the host releases those resources too; a full cancellation protocol felt disproportionate for this one-file PR.

aisraelov (approved):
Reviewed the narrowed SDK-side timeout scope and the follow-up test changes. The per-call value is forwarded into the existing `sendRequest` timeout without changing the default path, the caller/host cancellation boundary is documented, and the change merges cleanly with current `dev`. Verified the synthesized merge with the full Miniapp SDK suite (225 tests), typecheck, and build.

PR #3090, opened 2026-06-04
Bluetooth SDK should use promises

PhilippeFerreiraDeSousa:
## Summary

This is a breaking Bluetooth SDK ergonomics cleanup. Request/response-style operations now return typed results from correlated glasses/ASG feedback instead of requiring apps to call one method and separately wire listeners for the final result.

- Added request-id based pending result handling in the React Native module, Android SDK, and iOS SDK.
- `requestPhoto(...)`, `startVideoRecording(...)`, `stopVideoRecording(...)`, `rgbLedControl(...)`, `queryGalleryStatus()`, `checkForOtaUpdate()`, `startOtaUpdate()`, `requestWifiScan()`, and `requestVersionInfo()` now return corresponding response/result types.
- Gallery/button settings promises return `SettingsAckSuccessEvent` from correlated ASG acks; raw `settings_ack` listener events remain `SettingsAckEvent`. Local settings state updates only after the public promise success boundary, and FOV now returns a dedicated success-only `CameraFovResult` after ASG reports camera hardware application after the restart cooldown.
- `requestPhoto(...)` now resolves on terminal success for the full photo action: capture finished and the photo reached the configured target, either by direct glasses upload or phone Bluetooth fallback relay. It rejects on glasses/phone terminal `photo_response` errors, send failure, or timeout.
- `photo_status` remains the intermediate progress stream for accepted/configuring/capturing/captured/uploading/BLE-transfer states.
- Phone-side BLE fallback upload now emits terminal `photo_response` success/error after relay upload completes, so Android/iOS/RN SDK promises settle from the same final lifecycle as ASG direct upload.
- Successful `photo_response` payloads can include webhook-returned metadata (`photoUrl`, `statusUrl`, `contentType`, `fileSizeBytes`) in addition to `uploadUrl`. The ASG client still accepts older webhook aliases (`mimeType`, `bytes`, `size`) when normalizing terminal responses, but SDK consumers see the canonical fields.
- Local miniapp `session.camera.takePhoto(...)` remains end-to-end: it resolves from the phone/cloud upload result, while Bluetooth terminal errors short-circuit the in-flight request. `PhotoTaken` now includes `requestId`.
- Local miniapp `session.camera.setFov(...)` uses the same request/result model and resolves from the Bluetooth SDK ASG-backed hardware-applied acknowledgement; the local settings store is updated only after that success result returns.
[...]

PR #3474, opened 2026-07-17
Suspend BluetoothSdk glasses round-trips instead of blocking the Expo queue (OS-1714)

isaiahb:
Fixes OS-1714: on Android, every plain Expo AsyncFunction of every module shares one HandlerThread (expo.modules.AsyncFunctionQueue). MentraBluetoothSdk's PendingResponse parked that thread on a CountDownLatch for the full glasses round-trip, so any in-flight photo capture (or the 15 other latch-backed commands: warm-up, gallery, WiFi, hotspot, stream, video, version, OTA, settings) froze Crust.mentraJsDispatchToJs and with it ALL event delivery into miniapp background JS until the glasses responded.

Changes:
- PendingResponse: CountDownLatch to CompletableDeferred; await() suspends under withTimeoutOrNull, timeout/error semantics unchanged.
- The 16 latch-backed entry points become suspend funs (compiler-enforced caller audit); the OTA manifest poll's Thread.sleep becomes delay().
- Their BluetoothSdkModule bindings become Coroutine AsyncFunctions via new parameterized SdkCoroutineFunction helpers (same withExpoSdkError wrapping). JS-visible API is unchanged.
- iOS needs no changes: it already awaits a CheckedContinuation (verified), so this brings Android to parity with iOS.

Note for external consumers of the public Maven artifact: the 16 entry points change from blocking to suspend, which is a Kotlin API break (and the right shape to match the iOS SDK's async functions).

On-device verification (Pixel 8 + Mentra Live, APK from this branch, Mentra AI dev-connected, continuous TTS markers + mid-stream visual query, evidence attached to OS-1714):
- Before (dev): transcript delivery into miniapp JS froze for the exact photo window (3-4s real capture), zero events, then a 25-event FIFO burst the second photo_response arrived.
- After (this branch): steady 3-7 transcript deliveries/sec straight through a 3.7s in-flight capture, max inter-event gap 553ms across 54 events, delivery rate identical to the RN fan-out rate second-for-second. Visual query itself still works end to end (photo captured, answered, spoken).

Compile check: ./scripts/check-android-compile.sh bluetooth-sdk passes.

Spec/writeup: notes/miniapp-event-delivery-modulesqueue-hol.md (PR #3473). Linear: OS-1714 (sub-issue of OS-1687), related OS-1701.

---

> [!NOTE]
> **High Risk**
> Touches core Bluetooth request/response paths, public Kotlin API (`suspend` break for Maven consumers), and concurrency around WiFi scan, stream preemption, and OTA cancellation—high impact if regressions slip through.
> 
> **Overview**
> Fixes **OS-1714** by stopping glasses round-trips from blocking Android’s shared **`expo.modules.AsyncFunctionQueue`**, so long BLE waits (e.g. photo capture) no longer freeze miniapp JS event delivery.
> 
> **`PendingResponse`** now uses **`CompletableDeferred`** with a **`suspend`** **`await()`** under **`withTimeoutOrNull`** instead of **`CountDownLatch`**. The affected **`MentraBluetoothSdk`** entry points are **`suspend fun`**, and **`BluetoothSdkModule`** wires them through new parameterized **`SdkCoroutineFunction`** helpers (JS API unchanged). OTA version polling uses **`delay()`** instead of **`Thread.sleep`**.
> 
> **WiFi scan** lets concurrent callers join one in-flight scan with a waiter count; cancellation only detaches that caller, and the last waiter clears the shared entry. **Stream start** tracks send order with a sequence and rejects older pending starts with **`stream_preempted`** when a newer start’s status arrives. **`getFreshGlassesStatus`** rethrows **`CancellationException`** so cancelled OTA flows cannot continue with stale status.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 538166e08e4fe7707d57ca843898f4f53cedbd81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes OS-1714 by suspending Bluetooth glasses round-trips instead of blocking the shared `expo.modules.AsyncFunctionQueue`, so long operations (e.g., photo capture) no longer freeze miniapp background JS events. Matches iOS; JS API unchanged.

- **Bug Fixes**
  - Replaced latch-backed `PendingResponse` with `CompletableDeferred` + `withTimeoutOrNull`; timeout/error behavior unchanged.
  - Converted SDK entry points to `suspend fun` and bound them as coroutine `AsyncFunction`s via new `SdkCoroutineFunction` helpers, so the shared Expo queue suspends instead of blocking.
  - Swapped OTA polling `Thread.sleep` for `delay()`; `checkForOtaUpdate` runs on `Dispatchers.IO`.
  - Removed camera-command mutex and photo `requestId` dedup; firmware already queues requests and IDs are unique by contract.
  - Fixed OTA cancellation: `getFreshGlassesStatus` now rethrows `CancellationException` so `startOtaUpdate` cannot proceed after cancellation.
  - Hardened WiFi scan concurrency/cancellation: concurrent callers join one scan; cancellation detaches only that waiter; last waiter clears the shared entry to avoid rejecting joiners or wedging future scans.
  - Stream starts: fail fast when a newer start preempts an older one instead of waiting for the 30s timeout; preserves send order with a sequence and lock.

- **Migration**
  - Public Kotlin API change: affected methods are now `suspend`. Call from coroutines. No changes on iOS or to the JS-visible API.

<sup>Written for commit 538166e08e4fe7707d57ca843898f4f53cedbd81. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -758,7 +753,7 @@ class MentraBluetoothSdk private constructor(
         deviceManager.setSystemTime(timestampMs)
     }
 
-    fun requestPhoto(request: PhotoRequest): PhotoResponseEvent {
+    suspend fun requestPhoto(request: PhotoRequest): PhotoResponseEvent {
```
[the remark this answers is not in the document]
isaiahb: Agreed and fixed: added a camera-scoped suspending Mutex (kotlinx.coroutines.sync) around requestPhoto and warmUpCamera, so concurrent captures queue FIFO toward the glasses exactly as the old accidental serialization did, without holding the Expo queue thread. Scoped to camera commands rather than all glasses commands on purpose: the old global serialization of unrelated commands (WiFi scan behind a 15s photo, etc.) was part of the bug, not behavior to preserve. Compile check passes.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -769,14 +772,16 @@ class MentraBluetoothSdk private constructor(
         val pending = PendingResponse<PhotoResponseEvent>("photo request ${routedRequest.requestId}")
         pendingPhotoRequests[routedRequest.requestId] = pending
         try {
-            deviceManager.requestPhoto(routedRequest)
-            return pending.await()
+            cameraCommandMutex.withLock {
+                deviceManager.requestPhoto(routedRequest)
+                return pending.await()
+            }
```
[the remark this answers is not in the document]
isaiahb: Valid, fixed: requestPhoto now uses putIfAbsent and rejects a duplicate in-flight requestId with request_in_flight, mirroring warmUpCamera. Reachable in practice since the coordinator's 4-hex BLE correlation ids wrap and callers may supply their own ids. Compile check passes.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: The glasses have a photo request queue and implement the behavior as to whether wait or return camera busy! Do not make up a behavior in the Bluetooth SDK!

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -767,16 +770,29 @@ class MentraBluetoothSdk private constructor(
             "NATIVE: PHOTO PIPELINE [3b/6] MentraBluetoothSdk.requestPhoto requestId=${routedRequest.requestId}"
         )
         val pending = PendingResponse<PhotoResponseEvent>("photo request ${routedRequest.requestId}")
-        pendingPhotoRequests[routedRequest.requestId] = pending
+        // Reject duplicate in-flight requestIds instead of blind-overwriting
+        // (mirrors warmUpCamera): with suspending callers two overlapping
+        // requests sharing an id would otherwise clobber each other's pending
```
PhilippeFerreiraDeSousa: request ids are assumed unique. I know we do not enforce unicity at the Bluetooth SDK interface but we have to accept assumption to keep complexity in check. Those request ids are mostly made from timestamps.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -1079,12 +1097,12 @@ class MentraBluetoothSdk private constructor(
...
 
-    internal fun startOtaUpdate(otaVersionUrl: String): OtaStartAckEvent =
+    internal suspend fun startOtaUpdate(otaVersionUrl: String): OtaStartAckEvent =
         startOtaCommand(otaVersionUrl)
 
-    internal fun sendOtaQueryStatus(): OtaQueryResult = queryOtaStatus()
+    internal suspend fun sendOtaQueryStatus(): OtaQueryResult = queryOtaStatus()
 
-    private fun getFreshGlassesStatus(): GlassesStatus {
+    private suspend fun getFreshGlassesStatus(): GlassesStatus {
```
aisraelov, line 1120: **P1 — Preserve cancellation before OTA side effects.** `requestVersionInfo()` is now suspending, so the broad `catch (Throwable)` later in `getFreshGlassesStatus()` can catch `CancellationException` (including Expo `ContextDestroyedException`). In the no-URL `startOtaUpdate()` path, that catch returns stale status and execution proceeds into `startOtaCommand()`, which calls `deviceManager.sendOtaStart(...)` before the next suspension observes that the coroutine is cancelled. An app-context teardown or native caller cancellation during the version request can therefore still start OTA. Please rethrow `CancellationException` before the fallback and cover this path with a cancellation test.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -643,7 +646,7 @@ class MentraBluetoothSdk private constructor(
         return GlassesMediaVolumeSetResult.fromMap(deviceManager.setGlassesMediaVolumeBlocking(level))
     }
 
-    fun requestWifiScan(): List<WifiScanResult> {
+    suspend fun requestWifiScan(): List<WifiScanResult> {
```
aisraelov, line 654: **P2 — Do not let one waiter cancel the shared Wi-Fi scan.** This method intentionally lets concurrent callers join the same `PendingResponse`. Now that `await()` is cancellation-aware, cancellation of the first/owner coroutine reaches the existing `catch (Throwable)`, which calls `pending.reject(error)`. That completes the shared deferred with `CancellationException`, fails every independent joined caller, and the `finally` clears `pendingWifiScan` even though the glasses scan is still running. This is reachable for public SDK consumers using separate coroutine scopes. Please treat cancellation separately and decouple the shared scan lifetime from the first waiter, while still propagating real request failures/timeouts to all joiners.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -758,7 +761,7 @@ class MentraBluetoothSdk private constructor(
         deviceManager.setSystemTime(timestampMs)
     }
 
-    fun requestPhoto(request: PhotoRequest): PhotoResponseEvent {
+    suspend fun requestPhoto(request: PhotoRequest): PhotoResponseEvent {
```
aisraelov, line 786: **P2 — Update the public Android API examples and migration surface in this PR.** These new `suspend` signatures are a source/binary break for the published Maven facade, but the official docs still invoke them from non-suspending functions or unscoped snippets. For example, `mintlify-docs/mentra-live/android.mdx:93-95` and `quickstart.mdx:272-274` call `requestVersionInfo()` inside ordinary `fun refreshStatus()`, while `camera-streaming.mdx:33`, `:284-290`, and `:418-426` call the photo/settings/stream methods without a coroutine context. Those examples no longer compile. Please update the Android docs/samples to use an explicit lifecycle coroutine scope (and document the migration/versioning); if Java is supported, preserve an ergonomic Java async wrapper because a Kotlin suspend method only exposes a Continuation-shaped JVM signature.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -824,13 +842,13 @@ class MentraBluetoothSdk private constructor(
         }
     }
 
-    fun startStream(request: StreamRequest): StreamStatusEvent =
+    suspend fun startStream(request: StreamRequest): StreamStatusEvent =
```
aisraelov, line 852: **P2 — Guard overlapping stream starts now that the Expo queue no longer serializes them.** Two `startStream` calls can now both enter this path and populate `pendingStreamStarts` before either response arrives. `matchingStreamStart()` only routes a streamId-less status when the map has exactly one entry; with two pending starts, older/id-less glasses statuses match neither request and both calls wait for the 30-second timeout. A duplicate caller-supplied streamId also still blind-overwrites the first pending deferred. The engine coordinator serializes its own transitions, but this remains directly reachable through the public React Native/native SDK. Please add a one-in-flight guard or suspending stream-transition mutex (and use `putIfAbsent` for correlation IDs), or otherwise make concurrent starts explicitly routable.

PhilippeFerreiraDeSousa:
The issue was introduced by https://github.com/Mentra-Community/MentraOS/pull/3090 and affects Bluetooth SDK releases 0.1.9 - 0.1.20

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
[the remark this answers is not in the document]
isaiahb: You're right, this was me over-applying a bot suggestion without checking whether the firmware already owns this. Reverted the mutex in 6256d2a05, the glasses' own queue/busy behavior is the only source of truth here.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
isaiahb: Fair, this was solving for a case that shouldn't exist given the uniqueness assumption. Reverted the guard back to the original behavior in 6256d2a05.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
isaiahb: Fixed in ebb342fbe: getFreshGlassesStatus now rethrows CancellationException before the stale-status fallback, so a cancelled caller can no longer proceed into startOtaCommand/sendOtaStart. On the test: the module's existing unit tests only cover data classes and protocol helpers; exercising this path needs scaffolding to build the SDK against a fake DeviceManager, which I'd rather add as a follow-up than bolt onto this PR.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
isaiahb: Fixed in ebb342fbe. Cancellation of any one waiter (starter or joiner) now detaches only that waiter and leaves the shared deferred alone, so scan_complete still resolves for the others. A waiter count makes the last waiter out clear the shared entry, so an abandoned scan can't strand it or wedge future scans. Real failures and timeouts still reject the shared pending so joiners fail fast, same as before.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
isaiahb: Agreed, the published Android examples need updating for the suspend signatures plus a note on the Maven versioning impact. Handling the docs/migration surface as its own follow-up PR rather than growing this one.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
isaiahb: The correlation gap is real: with two id-less starts pending, matchingStreamStart can't route either status. But before adding client-side serialization here I want Philippe's read, since he rejected exactly that pattern (mutex + putIfAbsent) on requestPhoto because the glasses own queue/busy behavior. Will fix according to his answer below.

isaiahb: @PhilippeFerreiraDeSousa question on the stream-start path: now that startStream suspends instead of blocking the Expo queue, two concurrent starts can both be pending. matchingStreamStart only routes an id-less stream status when exactly one start is pending, so with two in flight both time out after 30s. Do the glasses handle overlapping stream starts themselves like the photo queue? If so I'll fix just the correlation routing; if not, what should the second start do at the SDK layer?

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -677,15 +683,23 @@ class MentraBluetoothSdk private constructor(
...
+    }
+
+    /** Last waiter (starter or joiner) out clears the shared scan entry so a
+     * cancelled starter can't strand it, and an abandoned scan can't wedge
+     * future ones. handleWifiScanResultsForRequests may clear it first. */
+    private fun releaseWifiScanWaiter(request: PendingWifiScan) {
+        synchronized(oneShotLock) {
+            request.waiters--
+            if (request.waiters <= 0 && pendingWifiScan === request) {
+                pendingWifiScan = null
```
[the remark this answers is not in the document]
isaiahb: Real in theory but pre-existing: the old code had the same window via the 20s timeout path (owner times out, entry cleared, glasses still scanning, next scan's entry gets the stale scan_complete). Cancellation widens that window; it doesn't create it. It also can't be fixed correctly on the phone: the glasses' scan events carry no correlation id, so the handler cannot attribute a scan_complete to a specific scan, and any client-side guess would be inventing protocol behavior in the SDK. The proper fix is a scan correlation id at the protocol level, which is a glasses-side change. Given the payload is a current-environment network list, leaving this as-is and flagging the protocol gap.

@PhilippeFerreiraDeSousa flagging the underlying gap for your queue rather than working around it here: WiFi scan results / scan_complete carry no correlation id on the wire, so the phone can only associate them with "the current scan" heuristically. If a scan id ever gets added to the protocol, the SDK association becomes exact and this whole class of stale-association goes away.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -824,13 +819,13 @@ class MentraBluetoothSdk private constructor(
...
 
-    fun startStream(request: StreamRequest): StreamStatusEvent =
+    suspend fun startStream(request: StreamRequest): StreamStatusEvent =
         startStream(request, startSdkKeepAlive = true)
 
-    internal fun startExternallyManagedStream(request: StreamRequest): StreamStatusEvent =
+    internal suspend fun startExternallyManagedStream(request: StreamRequest): StreamStatusEvent =
         startStream(request, startSdkKeepAlive = false)
 
-    private fun startStream(
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Verified on hardware (Mentra Live, two start_stream commands 2s apart via the glasses' command intent): the glasses preempt — `stopAllServices()` runs before every start, last request wins, no queue. Observed: `initializing{testA}` → `stopped` (no streamId) → `initializing{testB}` → `reconnecting{testB}`. Lifecycle statuses echo streamId, so the winner always resolves via its id-matched `streaming`; "both calls hit the 30s timeout" is wrong for the winner. The loser is real: its only verdict is the id-less `stopped`, so it ran out the 30s timeout with a misleading `request_timeout`.

Fixed in 538166e08e by decoding the glasses' semantics instead of serializing: pending starts carry a registration sequence assigned in the same critical section as the BLE hand-off (so seq order provably equals wire order), and since commands are FIFO and every start stops everything older, any id-matched status for start X rejects every earlier pending start immediately with `stream_preempted`. Id-less routing unchanged.

On the duplicate-streamId overwrite: declined — request ids are assumed unique at the SDK boundary (timestamp/UUID-derived), same policy as the reverted photo putIfAbsent guard.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
PhilippeFerreiraDeSousa: Answered by hardware test (two start_stream 2s apart via adb on a Mentra Live): the glasses do NOT queue stream starts like photos — `handleStartCommand` calls `stopAllServices()` first, so it's preemption, last-wins. Observed: `initializing{testA}` → `stopped` (no streamId) → `initializing{testB}` → `reconnecting{testB}`.

So no serialization in the SDK, correlation routing only — pushed in 538166e08e. Commands are FIFO and each start stops everything older, so an id-matched status for start X is proof every earlier pending start was preempted; those now reject fast with `stream_preempted` instead of timing out. The second start needs nothing special — it resolves normally via its own id-echoed `streaming`.

Two follow-ups stacked on this branch: #3487 brings iOS to parity (it was missing the id-less-STOPPED guard, so a start replacing a running stream got falsely rejected `stream_start_failed` on iOS — the hardware test emits exactly that sequence — plus the same preemption routing, and aligns Android's ERROR handling to iOS's tolerate-transient-errors semantics with an id-presence discriminator so command-level rejections still fail fast). #3488 closes the protocol gap underneath: streamId echoed on `stopping`/`stopped` and on command-level start rejections, plus WiFi scan correlation ids.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Confirmed your read, in code and on the wire: `wifi_scan_result` carries only `timestamp` + `networks`/`networks_neo` — no correlation id — so phone-side attribution is heuristic by construction, the stale-association window predates this PR (the timeout path had it), and severity is low since the payload is a current-environment network list. Agreed it can't be fixed correctly on the phone.

Protocol fix opened as #3488 (stacked on this branch): `request_wifi_scan` gains an optional `scanId`, bound to the scan that produced the results and echoed in every chunk, and the phone matches by scanId when echoed (stale ids ignored, chunks accumulated until `scanComplete` — which also fixes the phone resolving on the first 4-network chunk). Old-firmware fallback keeps today's behavior unchanged. Same PR carries the streamId echo on stop/rejection statuses.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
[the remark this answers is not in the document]
isaiahb: Clean solution, thanks for the hardware test and the fix. The seq-in-critical-section trick making registration order provably equal wire order is a nice way to get fast preemption verdicts without serializing anything. Pulled 538166e08e locally, and will review #3487 and #3488.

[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Heads up: I pushed one more commit on top (e4034912bb) — a forward-compat guard on the same rejectPreemptedStreamStarts. Bugbot flagged (correctly) that once glasses firmware id-stamps command-level preflight rejections (missing URL / low battery / no WiFi — that echo lands in #3488), an id-carrying ERROR arriving *before* stopAllServices would wrongly preempt a still-live older start and strand its keep-alive. The guard restricts preemption to the states the glasses only reach after stopAllServices ran (initializing/reconnecting/reconnected/streaming). It was originally sitting in #3488; moved it here so the phone-side guard is unconditional rather than coupled to the firmware change. Compile-checked, and the iOS mirror is in #3487.

PR #3478, opened 2026-07-17
ASG: dead-man response watchdog for the BES OTA transfer (clean abort, no resume)

PhilippeFerreiraDeSousa:
> **Stacked on #3458** (uses its owner-keyed `WakeLockManager` and `AsgConstants` entries). Merge #3458 first; this PR then retargets to `dev` automatically.

## Problem

The BES firmware transfer is **response-driven**: every inbound OTA response (`RCMD_SEND_DATA` ack, `RCMD_SEGMENT_VERIFY` confirm) triggers the next send — there is no wait loop anywhere. So a single lost response stalls the state machine **silently and forever**: nothing is blocked, nothing times out, nothing logs.

That's exactly the 2026-07-08 field incident: the device slept mid-transfer, the response between segments 56/57 was lost, and the transfer froze at 80% with no further log line. The phone's 120s stall watchdog produced the "Update failed" screen, but the glasses side never recovered, never cleaned up, and held its OTA mode and wake lock until timeout.

#3458's screen lease removes the *sleep* cause; this PR adds the safety net for **any** lost-response cause (BES crash, UART drop, ESD glitch).

## Change — one dead-man timer, nothing else

Mirrored on the class's existing authorization-timeout pattern (Handler + `removeCallbacks`, cancelled in `cleanup()`):

- **Armed** when authorization is granted (transfer starts), **re-armed** on every inbound OTA response, **cancelled** in `cleanup()` — the single funnel every terminal path already goes through.
- **If it fires** (30s with no response — normal inter-response gaps are sub-second; the window sits well below the phone's 120s watchdog so the glasses clean up first): abort through the **existing** failure path (`BesOtaProgressEvent.createFailed` + `cleanup()`), and emit a `bes_ota_response_timeout` lifecycle trace with the transfer position (`sentPos`, `confirmedSegments`).

## Deliberately NOT in this PR

**No resume. No retry. No state carried across attempts.** An aborted transfer is transactionally safe: the BES only swaps firmware slots after receiving a complete, CRC-valid image, so the device stays on its current bootable firmware and the phone re-runs the whole OTA from scratch — the proven path (the 07-08 device recovered exactly this way). Breakpoint-resume style recovery inside the transfer is rarely-exercised code on the one flow where a malfunction bricks a customer device; simplicity is the reliability feature here.

## Test evidence

- `:app:compileDebugJavaWithJavac` and `:app:compileDebugUnitTestJavaWithJavac` clean.
- Watchdog lifecycle is pinned to existing, exercised mechanisms only (auth-timeout Handler pattern, `createFailed`, `cleanup()`).
- On-device check planned with the next OTA test session: pull the BES mid-transfer (or kill UART) and confirm abort at 30s with the `bes_ota_response_timeout` trace, then a successful full retry from the phone.

## Notes for reviewers

- Companion to #3458 (wake leases). Together they close the 07-08 incident: the screen lease prevents the sleep-kill; this watchdog guarantees a stalled transfer of any cause ends in a clean, retryable failure instead of a silent wedge.

---

> [!NOTE]
> **Medium Risk**
> Changes firmware OTA failure and cleanup timing on customer devices; logic is constrained to existing failure/cleanup paths with tests for the watchdog race, but field UART edge cases remain.
> 
> **Overview**
> Adds a **30s dead-man watchdog** so a response-driven BES OTA transfer cannot stall silently when a UART/BES response is lost. `AsgConstants.BES_OTA_RESPONSE_TIMEOUT_MS` defines the window (below the phone’s 120s stall watchdog so glasses clean up first).
> 
> In **`BesOtaManager`**, a Handler-based timer is armed at authorization grant and **re-armed on every inbound OTA response**; `cleanup()` cancels it with other terminal paths. If it fires, the existing failure path runs (`BesOtaProgressEvent.createFailed`, `cleanup()`), plus a `bes_ota_response_timeout` **`BleTraceLogger`** event with `sentPos` / `confirmedSegments`. **`mTransferGate`** serializes watchdog abort vs UART receive so a late response cannot race into a false abort (deadline re-check on the dispatched callback).
> 
> **Robolectric tests** (`BesOtaResponseWatchdogTest`) cover timeout abort, re-arm standing down a due callback, and no timer after cleanup. **`bes-ota.md`** documents the new stall behavior and OTA-mode recovery via watchdog cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc30001f610317e1be02ea170095c70c95a12a58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
@@ -352,6 +352,17 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
 
     /** Cleanup and reset state */
     private void cleanup() {
+        // Every terminal path funnels here, gated or not (watchdog and receive dispatch
+        // hold mTransferGate; abortIfInProgress and the auth failure paths do not).
+        // Taking the gate here - reentrant for the former - serializes ALL cleanups with
+        // receive processing and the watchdog re-arm, so no caller can tear the watchdog
+        // state down mid-rearm on another thread.
+        synchronized (mTransferGate) {
+            cleanupLocked();
+        }
+    }
+
+    private void cleanupLocked() {
         // Cancel authorization timeout if pending
         if (authTimeoutHandler != null && authTimeoutRunnable != null) {
             authTimeoutHandler.removeCallbacks(authTimeoutRunnable);
@@ -812,7 +823,7 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
         if (responseWatchdogRunnable != null) {
             responseWatchdogHandler.removeCallbacks(responseWatchdogRunnable);
         }
-        responseWatchdogRunnable = () -> {
+        Runnable watchdog = () -> {
           synchronized (mTransferGate) {
             if (!isBesOtaInProgress) {
                 return;
@@ -838,8 +849,10 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
             cleanup();
           }
         };
-        responseWatchdogHandler.postDelayed(
-                responseWatchdogRunnable, AsgConstants.BES_OTA_RESPONSE_TIMEOUT_MS);
+        // Post the local reference: the field is only bookkeeping for removeCallbacks, so
+        // a concurrent teardown nulling it can never turn this into postDelayed(null).
+        responseWatchdogRunnable = watchdog;
+        responseWatchdogHandler.postDelayed(watchdog, AsgConstants.BES_OTA_RESPONSE_TIMEOUT_MS);
     }
 
     private void dealOtaRecvCmd(BesOtaMessage msg) {
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
@@ -51,6 +51,12 @@ public class BesOtaManager implements IBesOtaController, BesOtaUartListener, Bes
     // machine silently forever (2026-07-08 incident: frozen at 80%, no further log line).
     private android.os.Handler responseWatchdogHandler;
     private Runnable responseWatchdogRunnable;
+    // Absolute deadline (elapsedRealtime) of the current watchdog window. The re-arm from
+    // the UART receive thread cannot removeCallbacks() a callback that has ALREADY been
+    // dispatched on the main looper; the dispatched callback re-checks this deadline and
+    // aborts only if no response re-armed it in the meantime, so a response racing the
+    // timeout can never produce a false failure while its handler is mid-flight.
+    private volatile long responseWatchdogDeadlineMs;
 
     private String filePath;
     private boolean bInit = false;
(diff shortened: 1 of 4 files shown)


asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
```diff
@@ -750,7 +776,49 @@ public void onParseResult(byte cmd, byte[] data) {
...
+                extra.put("fileLen", fileLen);
+                extra.put("confirmedSegments", confirmTimes);
+                BleTraceLogger.logLifecycle(mContext, "BesOtaManager", "bes_ota_response_timeout", extra);
+            } catch (Exception ignored) {
+                // Trace logging must never affect the abort itself.
+            }
+            EventBus.getDefault().post(BesOtaProgressEvent.createFailed(
+                    "No response from BES for "
+                            + (AsgConstants.BES_OTA_RESPONSE_TIMEOUT_MS / 1000) + "s"));
+            cleanup();
```
aisraelov: **[P2] Serialize the timeout decision with receive handling**

The volatile deadline makes a re-arm visible, but it does not make the deadline check and abort atomic. The main callback can read the old expired deadline at line 796, then `RecvThread` can re-arm the watchdog, and the callback will still post failure and call `cleanup()` while `dealOtaRecvCmd()` continues. That cleanup nulls `fileData` and disables OTA mode, so a valid segment response can still hit a failed send or null-state crash during the CRC/file-buffer work.

Please serialize the timeout transition/cleanup with the receive state machine (for example, run both on one handler/executor, or guard the full transition so the losing side immediately stops processing) and add a deterministic Robolectric race test.

aisraelov (changes requested):
Requesting changes for the remaining watchdog/receive concurrency race.

Non-blocking documentation follow-up: please update `asg_client/docs/features/bes-ota.md` to describe the new 30-second automatic abort, cleanup, `bes_ota_response_timeout` trace, and full-transfer retry behavior. Its troubleshooting section currently still describes a silent stall and manual app restart.

Validation on this head: ASG Java compile, unit-test source compile, and focused `*BesOta*` tests passed. The stacked prerequisite #3458 still needs to merge first.

asg_client/app/src/main/java/com/mentra/asg_client/io/bes/BesOtaManager.java
PhilippeFerreiraDeSousa: Fixed in 65ab5e7 by full serialization rather than a finer-grained guard: the watchdog's entire decide+abort and the whole receive dispatch (`dealOtaRecvCmd`) now run inside a single transfer gate. Either the timeout wins the monitor and aborts first — the late response then sees the transfer inactive at the gate and stops before touching `fileData` — or the response wins, re-arms, and the already-dispatched callback re-checks the deadline under the same monitor and stands down. The torn interleaving you describe is structurally impossible now. Added the deterministic Robolectric tests (`BesOtaResponseWatchdogTest`): timeout-with-no-response aborts + reports failure; a response racing an already-dispatched callback stands it down; a stale re-arm after cleanup leaves no timer. Docs follow-up included in the same commit — the troubleshooting section now describes the 30s abort, cleanup, `bes_ota_response_timeout` trace, and full-transfer retry instead of the silent stall + manual restart.

aisraelov (approved):

PhilippeFerreiraDeSousa:
Mistakenly merged into claude/asg-wake-window-per-w1-command instead of dev, it was remerged manually into dev

PR #3487, opened 2026-07-17
Bring iOS stream-start semantics to parity with Android

PhilippeFerreiraDeSousa:
Follow-up to #3474 (stacked on `isaiah/os-1714-suspend-bluetooth-asyncfunctions`). Brings iOS and Android stream-start semantics to parity in both directions, prompted by the overlapping-start review threads on that PR.

Background (hardware-verified on a Mentra Live via the glasses' command intent): the glasses preempt on `start_stream` — `stopAllServices()` runs before every start, last request wins, no queue. Lifecycle statuses echo `streamId`; `stopping`/`stopped` and command-level rejections (missing URL, battery low, no WiFi) arrive with no `streamId` on current firmware. A start that replaces a running publisher emits exactly `stopped` (id-less) → `initializing{new}` → `streaming{new}`.

Changes:

- **Ignore streamId-less stopped for a pending iOS stream start.** iOS `matchingStreamStart` was missing the guard Android already has, so the id-less stop-ack of the *previous* stream was attributed to the single pending start and rejected it with `stream_start_failed` even though the new stream goes live moments later. Real shipped-iOS bug; the hardware test emits the exact sequence. The guard has one deliberate exception (below).
- **Reject preempted iOS stream starts instead of timing them out.** Mirror of the Android preemption rejection from #3474: per-start sequence numbers assigned in the same await-free MainActor turn as the BLE hand-off; any id-matched status for start X rejects all lower-seq pending starts with `stream_preempted`.
- **Keep Android stream starts pending through transient stream errors.** Android rejected on the first `error` status; iOS deliberately tolerates them because the glasses publisher auto-retries transient transport errors. Android now matches iOS — with a discriminator both platforms need (found in review): an **id-less** `error` is the glasses' command-level rejection (missing URL, low battery, no WiFi), emitted before any publisher exists and never followed by another status, so it still rejects immediately with the glasses' `errorDetails`. Only **id-carrying** errors stay pending.
- **Fatal-error fast-fail preserved (both platforms).** A fatal (non-retryable) publisher error emits `error{streamId}` then winds down with an id-less `stopped`, bypassing the reconnect machinery. The error event is stashed on the pending start; the subsequent id-less `stopped` is then attributed to that start and rejects with the stashed error details. Without this, tolerating errors would have converted fatal start failures into 30s timeouts. The replace-running-publisher stop-ack is unaffected (its stopped arrives before any status for the pending id, so no stash exists).

Verification:
- Android: `scripts/check-android-compile.sh bluetooth-sdk` — BUILD SUCCESSFUL.
- iOS: SPM export + `xcodebuild -scheme MentraBluetoothSDK` — BUILD SUCCEEDED (intermediate and final states).
- On-device: the preemption/parity paths follow the status sequences captured on hardware for #3474; a live two-start run on iOS is still worth doing before release.

Related: OS-1714, PR #3474 discussion threads r3605423618 / r3605370232.

---

> [!NOTE]
> **Medium Risk**
> Changes core async stream start/stop completion logic on both mobile SDKs; wrong attribution of ambiguous statuses could still mis-resolve overlapping calls, but behavior is scoped to concurrent stream commands and is intended to fix known timeout/false-failure bugs.
> 
> **Overview**
> Aligns **Android and iOS** Bluetooth SDK stream **start/stop** promise handling with glasses FIFO behavior (last `start_stream` wins, overlapping commands) and upcoming firmware status shapes.
> 
> Both platforms track **send-order sequence numbers** on pending starts (and stops), reject older starts with **`stream_preempted`** or **`stream_stopped`** when a newer start or a stop settles them, and route **`stop_stream`** through the same ordering critical section as starts so stop acks can clear superseded starts instead of 30s timeouts.
> 
> **Status matching** is tightened: ignore **streamId-less `stopped`** as a verdict on a lone pending start unless a **`willRetry` error** was stashed; attribute ambiguous **id-less errors** to the newest pending start; treat **`reconnected`** as success; on **`stopped`/`reconnect_failed`**, reject using stashed **`lastError`** when a fatal publisher error preceded an id-less wind-down.
> 
> **`StreamStatusEvent`** gains optional **`willRetry`** (parsed from glasses maps, forwarded in `values`). Stream errors with **`willRetry: true`** keep the start pending; terminal or legacy errors fail fast (preserving pre-#3487 behavior when the flag is absent).
> 
> iOS adds **`PendingStreamStart`** and mirrors Android’s preemption/supersede helpers; preemption on iOS is triggered only for id-matched **post-`stopAllServices`** lifecycle states (`initializing`, `reconnecting`, `reconnected`, `streaming`), avoiding false preemption from preflight errors or previous-stream stop-acks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85ac3f05b447ee9c636617bbedd8a78185d9e40c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

**Update (review round):** two refinements from the Codex findings, both verified in the service code:
- A matched pending start now also resolves on `reconnected` — the async start-failure path schedules a reconnect and reports success via `onReconnected`, never `onStreamStarted` (bf2c6a9b62).
- The error-tolerance rule is now discriminated on the wire instead of by id-presence alone: an id-carrying `error` rejects immediately with the glasses' `errorDetails` unless it carries the new optional `willRetry: true` (parsed here, emitted by #3488 exactly where the glasses schedule a reconnect). Terminal id-carrying errors like `camera_busy` — which emit nothing further — fast-fail again instead of hanging 30s. Old firmware never sends the flag, so its errors always fast-fail: the shipped pre-#3487 Android behavior, which also fixes the same latent camera_busy hang iOS had at base (5f6859a62d).

### DIFF
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
@@ -114,6 +114,11 @@ class MentraBluetoothSdk private constructor(
     private data class PendingStreamStart(
         val seq: Long,
         val pending: PendingResponse<StreamStatusEvent>,
+        // Set when an id-carrying error arrives: a fatal publisher error never
+        // reaches the reconnect machinery and winds down with a streamId-less
+        // stopped, so the stash both attributes that stopped to this start and
+        // preserves the real error details for the rejection.
+        var lastError: StreamStatusEvent? = null,
     )
 
     private data class PendingStreamStop(
@@ -1560,11 +1565,27 @@ class MentraBluetoothSdk private constructor(
                     pendingStreamStarts.remove(streamId, start)
                     start.pending.resolve(event)
                 }
-                StreamState.ERROR,
                 StreamState.RECONNECT_FAILED,
                 StreamState.STOPPED -> {
                     pendingStreamStarts.remove(streamId, start)
-                    start.pending.reject(streamStatusException(event, "stream_start_failed"))
+                    start.pending.reject(
+                        streamStatusException(start.lastError ?: event, "stream_start_failed")
+                    )
+                }
+                StreamState.ERROR -> {
+                    if (event.streamId.isNullOrBlank()) {
+                        // An id-less error is the glasses' command-level rejection
+                        // (missing URL, low battery, no WiFi), emitted before any
+                        // publisher starts; nothing further follows for this start.
+                        pendingStreamStarts.remove(streamId, start)
+                        start.pending.reject(streamStatusException(event, "stream_start_failed"))
+                    } else {
+                        // The glasses publisher automatically retries transient transport
+                        // errors, so keep the start pending for a later `streaming`.
+                        // Stash the details for the streamId-less `stopped` that a fatal
+                        // publisher error winds down with instead.
+                        start.lastError = event
+                    }
                 }
                 else -> Unit
             }
@@ -1621,15 +1642,19 @@ class MentraBluetoothSdk private constructor(
             return streamId to start
         }
         if (pendingStreamStarts.size == 1) {
+            val entry = pendingStreamStarts.entries.first()
             // A streamId-less STOPPED is the glasses' stop-ack for a PREVIOUS
             // stream (their stop ack carries no streamId), not a verdict on the
             // pending start — a start_stream that replaces a running publisher
             // emits exactly this sequence (stopped -> initializing -> streaming).
             // Attributing it here would reject a start that is about to succeed.
-            if (event.state == StreamState.STOPPED) {
+            // After an id-carrying error for the pending start, though, the
+            // stopped is a fatal publisher's wind-down (the stop-ack always
+            // precedes any status for the new start), so it is that start's
+            // verdict.
+            if (event.state == StreamState.STOPPED && entry.value.lastError == null) {
                 return null
             }
-            val entry = pendingStreamStarts.entries.first()
             return entry.key to entry.value
         }
         return null
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
@@ -1680,6 +1680,18 @@ class MentraBluetoothSdk private constructor(
             val start = pendingStreamStarts[streamId] ?: return null
             return streamId to start
         }
+        if (event.state == StreamState.ERROR) {
+            // An id-less error is a command-level preflight rejection (missing
+            // URL, low battery, no WiFi) emitted synchronously, while lifecycle
+            // statuses arrive async — so with overlapping starts the wire is
+            // genuinely ambiguous about which start it answers. Attribute it to
+            // the newest pending start: the common case is a later start
+            // failing preflight while an earlier one is already emitting
+            // lifecycle statuses, and at worst this swaps which start carries
+            // the error instead of letting both time out.
+            val entry = pendingStreamStarts.entries.maxByOrNull { it.value.seq } ?: return null
+            return entry.key to entry.value
+        }
         if (pendingStreamStarts.size == 1) {
             val entry = pendingStreamStarts.entries.first()
             // A streamId-less STOPPED is the glasses' stop-ack for a PREVIOUS
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
@@ -1561,7 +1561,11 @@ class MentraBluetoothSdk private constructor(
                 rejectPreemptedStreamStarts(start.seq)
             }
             when (event.state) {
-                StreamState.STREAMING -> {
+                // `reconnected` also proves the stream is live: when an async start
+                // failure triggers the glasses' publisher retry, a successful retry
+                // reports `reconnected` instead of `streaming`.
+                StreamState.STREAMING,
+                StreamState.RECONNECTED -> {
                     pendingStreamStarts.remove(streamId, start)
                     start.pending.resolve(event)
                 }
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
@@ -123,6 +123,10 @@ class MentraBluetoothSdk private constructor(
 
     private data class PendingStreamStop(
         val streamId: String?,
+        // The stop's slot in the shared streamStartSeq order (drawn inside
+        // streamStartOrderLock at send time): every pending start with a
+        // lower seq reached the glasses' FIFO before this stop.
+        val seq: Long,
         val pending: PendingResponse<StreamStatusEvent>,
     )
 
@@ -919,18 +923,25 @@ class MentraBluetoothSdk private constructor(
         val targetStreamId = synchronized(streamKeepAliveLock) {
             activeStreamKeepAlive?.streamId
         }
-        synchronized(oneShotLock) {
-            if (pendingStreamStop != null) {
-                throw BluetoothSdkException(
-                    "request_in_flight",
-                    "A stream stop command is already waiting for a glasses response.",
-                )
-            }
-            pendingStreamStop = PendingStreamStop(targetStreamId, pending)
-        }
-        stopStreamKeepAliveMonitor()
         try {
-            deviceManager.stopStream()
+            // The seq draw and the BLE hand-off share startStream's ordering
+            // critical section: the stop takes its own slot in the send order,
+            // so its ack can separate the pending starts the glasses consumed
+            // before it (lower seq) from starts sent after it (higher seq).
+            synchronized(streamStartOrderLock) {
+                synchronized(oneShotLock) {
+                    if (pendingStreamStop != null) {
+                        throw BluetoothSdkException(
+                            "request_in_flight",
+                            "A stream stop command is already waiting for a glasses response.",
+                        )
+                    }
+                    pendingStreamStop =
+                        PendingStreamStop(targetStreamId, streamStartSeq.incrementAndGet(), pending)
+                }
+                stopStreamKeepAliveMonitor()
+                deviceManager.stopStream()
+            }
             return pending.await(STREAM_STOP_TIMEOUT_MS)
         } finally {
             synchronized(oneShotLock) {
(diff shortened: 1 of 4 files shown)


mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -1560,11 +1565,27 @@ class MentraBluetoothSdk private constructor(
...
+                        // (missing URL, low battery, no WiFi), emitted before any
+                        // publisher starts; nothing further follows for this start.
+                        pendingStreamStarts.remove(streamId, start)
+                        start.pending.reject(streamStatusException(event, "stream_start_failed"))
+                    } else {
+                        // The glasses publisher automatically retries transient transport
+                        // errors, so keep the start pending for a later `streaming`.
+                        // Stash the details for the streamId-less `stopped` that a fatal
+                        // publisher error winds down with instead.
+                        start.lastError = event
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Confirmed and fixed in bf2c6a9b62: a matched pending start now resolves on RECONNECTED exactly like STREAMING. Verified in RtmpStreamingService that the async start-failure path calls scheduleReconnect and a successful retry reports via onReconnected (never onStreamStarted), so `reconnected` is a legitimate success verdict for a start. Reconnected events for established streams find no pending entry and stay no-ops.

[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Confirmed — camera_busy emits an id-carrying error and returns with no follow-up, so stash-and-wait hangs it for the full 30s. The reliable discriminator can only come from the emitter, so it's now on the wire: 5f6859a62d rejects id-carrying errors immediately with the glasses' errorDetails unless the status carries the new optional `willRetry: true`, which #3488 (1a01e414e5) emits exactly at the call sites that schedule a reconnect (the two start_error/start_exception paths in RTMP and SRT; every other onStreamError site was enumerated and is terminal — WHIP's reconnect path never emits errors at all). Old firmware never sends the flag, so its errors always fast-fail — the shipped pre-#3487 Android behavior, which also fixes the same latent hang iOS had at base.

aisraelov (commented):
I found four additional issues in the current head: two request-correlation paths that leave startStream pending until timeout, one public SDK compatibility break, and one Android map race. Details are inline.

mobile/modules/bluetooth-sdk/ios/Source/MentraBluetoothSDK.swift
```diff
@@ -1447,12 +1493,43 @@ public final class MentraBluetoothSDK {
...
         if pendingStreamStarts.count == 1, let entry = pendingStreamStarts.first {
+            // A streamId-less STOPPED is the glasses' stop-ack for a PREVIOUS
+            // stream (their stop ack carries no streamId), not a verdict on the
+            // pending start — a start_stream that replaces a running publisher
+            // emits exactly this sequence (stopped -> initializing -> streaming).
+            // Attributing it here would reject a start that is about to succeed.
+            // After a stashed `willRetry` error for the pending start, though,
+            // the stopped is the retrying publisher's wind-down (the stop-ack
+            // always precedes any status for the new start), so it is that
+            // start's verdict.
```
aisraelov, line 1579: **P1 — Do not drop id-less rejections while starts overlap.** This fallback only runs when exactly one start is pending. On current firmware, command-level rejections are id-less. If start A is still initializing when overlapping start B fails preflight (low battery, no Wi-Fi, or invalid URL), both entries remain in the dictionary, so B’s error matches neither entry and B waits the full 30-second timeout. A can still succeed because these validations occur before `stopAllServices()`. PR #3488 fixes new firmware by echoing IDs, but the advertised old-firmware fallback still needs an explicit ambiguous/ordered attribution strategy instead of silently dropping this verdict. Android has the same count-based behavior.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -1621,15 +1653,19 @@ class MentraBluetoothSdk private constructor(
...
             // stream (their stop ack carries no streamId), not a verdict on the
             // pending start — a start_stream that replaces a running publisher
             // emits exactly this sequence (stopped -> initializing -> streaming).
             // Attributing it here would reject a start that is about to succeed.
-            if (event.state == StreamState.STOPPED) {
+            // After a stashed `willRetry` error for the pending start, though,
+            // the stopped is the retrying publisher's wind-down (the stop-ack
+            // always precedes any status for the new start), so it is that
+            // start's verdict.
+            if (event.state == StreamState.STOPPED && entry.value.lastError == null) {
```
aisraelov, line 1709: **P1 — Reject a pending start when the caller explicitly stops it.** This guard also ignores the id-less `stopped` emitted by current firmware when `stopStream()` is called while `startStream()` is still initializing. The stop promise resolves later in this handler, but the start promise receives no further status and waits the full 30-second timeout. The locally pending stop operation needs to distinguish this from the automatic stop-ack produced while replacing an already-running publisher. The iOS implementation mirrors the same regression.

mobile/modules/bluetooth-sdk/ios/Source/streaming/StreamModels.swift
```diff
@@ -346,7 +346,15 @@ public enum StreamStatus: CustomStringConvertible, Equatable {
     )
     case reconnected(streamId: String?, attempt: Int, timestamp: Int?, resolvedConfig: StreamResolvedConfig?)
     case reconnectFailed(streamId: String?, maxAttempts: Int, timestamp: Int?, resolvedConfig: StreamResolvedConfig?)
-    case error(streamId: String?, errorDetails: String, timestamp: Int?, resolvedConfig: StreamResolvedConfig?)
+    // `willRetry` is true when the glasses will retry the failed publisher
+    // themselves (emitting side lands in PR #3488); absent on older firmware.
+    case error(
+        streamId: String?,
+        errorDetails: String,
+        willRetry: Bool?,
```
aisraelov: **P2 — Preserve the public SDK shape or mark this as a breaking release.** Adding `willRetry` to this public enum case changes every construction and pattern match from four associated values to five, so existing Swift package consumers no longer compile. The Android public `StreamStatus.Error` data class also changes constructor/component JVM signatures, breaking positional source callers and already-compiled consumers. The field is optional on the wire, but these model changes are not API-additive; consider keeping retry metadata on `StreamStatusEvent` or providing compatibility entry points.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -1621,15 +1653,19 @@ class MentraBluetoothSdk private constructor(
             return streamId to start
         }
         if (pendingStreamStarts.size == 1) {
+            val entry = pendingStreamStarts.entries.first()
```
aisraelov: **P2 — Avoid a size/first race on the ConcurrentHashMap.** `size == 1` and `entries.first()` are separate operations. A timeout or another status can remove the pending entry between them, causing `NoSuchElementException` on the BLE event path. This PR moves `first()` ahead of the stopped early return, extending the race to the common id-less replacement-stop event. Please take an atomic/locked snapshot, or at minimum use `firstOrNull()` and revalidate.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
PhilippeFerreiraDeSousa: Fixed in c49422a3a4: the stop now draws its own slot from the shared start-order counter inside the same critical section that hands it to BLE, and its ack rejects every pending start with a lower seq using a dedicated `stream_stopped` code — FIFO makes that attribution exact. Starts sent after the stop keep waiting for their own statuses; the replace-running-publisher guard is untouched. Both platforms.

mobile/modules/bluetooth-sdk/ios/Source/MentraBluetoothSDK.swift
PhilippeFerreiraDeSousa: Fixed in 4bb5b63a2c, with the ambiguity stated rather than papered over: preflight rejections are emitted synchronously while lifecycle statuses arrive async, so with overlapping starts the wire genuinely cannot say which start an id-less error answers. The deterministic choice is the NEWEST pending start — matches the common case (later start fails preflight while the earlier one is already emitting lifecycle statuses) and at worst swaps which start carries the error instead of both timing out. Both platforms; other id-less states keep the size==1 heuristic.

mobile/modules/bluetooth-sdk/ios/Source/streaming/StreamModels.swift
PhilippeFerreiraDeSousa: Agreed and fixed in eadeffb32e: StreamStatus.Error is back to its exact 4-param shape (Kotlin ctor and Swift case arity verified byte-identical to the pre-change signatures), and willRetry now lives additively on StreamStatusEvent — body property outside equals/hashCode/copy on Kotlin, private(set) stored property on Swift, populated from the raw values map and round-tripped. No public constructor or init signature changed.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
PhilippeFerreiraDeSousa: Fixed in 44f2d51a0e — firstOrNull with a null bail on the heuristic path, and the new newest-attribution branch uses maxByOrNull, so no id-less path can throw if the map empties between the size check and the read.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/MentraBluetoothSdk.kt
```diff
@@ -1556,15 +1572,40 @@ class MentraBluetoothSdk private constructor(
                 rejectPreemptedStreamStarts(start.seq)
             }
             when (event.state) {
-                StreamState.STREAMING -> {
+                // `reconnected` also proves the stream is live: when an async start
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Real catch, and it exposed a bad split on my part. The Android `rejectPreemptedStreamStarts` is defined in the base PR #3474, so its guard belongs there too — not gated behind #3488's firmware change (a phone on #3474/#3487 will meet #3488 firmware that id-stamps preflight errors on an independent update cadence, which is exactly the window you describe). I've hoisted the restriction to the base (commit e4034912bb on #3474): preemption now fires only for INITIALIZING/RECONNECTING/RECONNECTED/STREAMING on both platforms — iOS via #3487's call-site guard, Android now unconditionally in the base. #3488's duplicate commit was dropped in a rebase. So Android and iOS are symmetric and the guard no longer depends on firmware version.

PR #3488, opened 2026-07-17
Add correlation ids to stream stop statuses and WiFi scans

PhilippeFerreiraDeSousa:
Follow-up to #3474 (stacked on `isaiah/os-1714-suspend-bluetooth-asyncfunctions`). Closes the protocol correlation gaps surfaced in that PR's review: stream stop/rejection statuses and WiFi scan results carried no id on the wire, so the phone could only attribute them heuristically. Every new field is additive and optional — old phone + new glasses ignores them, new phone + old glasses falls back to exactly today's heuristics. No firmware version gating.

Hardware background (Mentra Live via the glasses' command intent): a stream started with `streamId:testA` emitted its preemption `stopped` with **no** streamId; command-level start rejections (missing URL, battery low, no WiFi) are id-less too; `wifi_scan_result` chunks carry only `timestamp` + `networks`.

Glasses (`asg_client`):

- **Carry streamId on stream stop and rejection statuses.** `handleStartCommand` hoists the `streamId` parse ahead of every early rejection and echoes it in those error statuses; `handleStopCommand`'s `stopping` carries the active stream's id. Root cause of the id-less `stopped` fixed in all three services: RTMP/SRT cleared `mCurrentStreamId` (timeout-cancel + state reset) before `onStreamStopped` fired — the id is now captured at stop entry; WHIP's `notifyStopped` read the field inside a main-handler runnable that could run after a newer start overwrote it — captured at call time.
- **Echo request scanId in every wifi_scan_result chunk.** `request_wifi_scan` accepts an optional `scanId`, captured as a local and closed over by the scan callback, so each scan's chunks are stamped with the id of the request that started *that* scan (no shared mutable field — overlapping scans can't cross-stamp, which was a review finding against the first cut). The network-manager-unavailable path now sends a terminal empty `scanComplete` message instead of letting the phone wait out its timeout.

Phone (`mobile/modules/bluetooth-sdk`, Android + iOS mirrored):

- **Correlate WiFi scans with a generated scanId.** `requestWifiScan` sends `scan-<uuid>`; results carrying a different scanId are ignored (the stale-association from #3474 discussion r3606439346), matching chunks accumulate (deduped by SSID) and resolve on `scanComplete` — fixing the first-chunk truncation where the phone resolved on the first 4-network chunk. Results with no scanId (old firmware) keep today's behavior unchanged. Other SGC implementations get ignoring stubs for the new parameter.
- **iOS chunk ordering made sound.** Accumulate-until-complete makes ordering load-bearing; the per-message unstructured `Task` hop could reorder chunks onto the MainActor. Replaced with `DispatchQueue.main.async` + `MainActor.assumeIsolated` (chunks originate on the serial bluetooth queue; serial→main dispatch is FIFO; same pattern as `BluetoothSdkModule.readOnMainActor`).

Stream side needs no phone-side changes: the existing exact-match routing plus the preemption rejection landed on #3474 already consume echoed ids. Behavior note: with new firmware, a preempted start fast-fails via its own id-echoed `stopped` (`stream_start_failed`) slightly before the newer start's `initializing` would have rejected it with `stream_preempted`; both are fast and accurate, the code just prefers the earlier signal.

Verification:
- `scripts/check-android-compile.sh asg` — BUILD SUCCESSFUL.
- `scripts/check-android-compile.sh bluetooth-sdk` — BUILD SUCCESSFUL.
- iOS: SPM export + `xcodebuild -scheme MentraBluetoothSDK` — BUILD SUCCEEDED.
- On-device verification of the streamId echo and a correlated scan on real glasses still pending (needs an asg_client build flashed).

Related: OS-1714, PR #3474 discussion threads r3605504628 / r3606439346.

---

> [!NOTE]
> **Medium Risk**
> Changes livestream BLE status semantics and WiFi scan correlation across glasses and mobile SDKs; fields are optional but mis-correlation could affect stream UI and provisioning flows.
> 
> **Overview**
> Adds **additive wire correlation** so the phone can attribute stream and WiFi events to the right request without changing behavior for old firmware or clients.
> 
> **Streaming (glasses):** Start rejections and **stopping** statuses now echo the command’s **`streamId`**. RTMP/SRT capture **`stoppedStreamId`** at stop entry so **`onStreamStopped`** and async stop failures still name the stream after timeout cancel or a newer start overwrites **`mCurrentStreamId`**; WHIP does the same when posting **`onStreamStopped`**. **`StreamingStatusCallback.onStreamError`** gains **`willRetry`** (default overload unchanged); RTMP/SRT pass **`true`** only where **`scheduleReconnect`** runs after start failures. **`MediaManager`** forwards **`willRetry`** on stream error BLE status.
> 
> **WiFi scan (glasses + SDK):** **`request_wifi_scan`** accepts optional **`scanId`**, echoed on every **`wifi_scan_result`** chunk; single-flight scan can **adopt** a newer id for concurrent requests. **`CommunicationManager`** centralizes scan result envelopes; unavailable network manager sends a terminal empty result instead of hanging.
> 
> **Bluetooth SDK (Android/iOS):** Generates **`scan-<uuid>`**, **`claimWifiScanResults`** before sending the command, accumulates/dedupes chunks per id until **`scanComplete`**, ignores stale ids; Bridge only mutates the device store for the active scan and forwards per-chunk payloads when correlated. iOS avoids reordering chunks via main-queue FIFO dispatch instead of per-message **`Task`** hops.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f4049194d8014a5521213204ed34b02bbc9d555. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

**Update (review round):**
- WiFi scan chunk forwarding is now scoped to the requesting scan: with a scanId, the Bridge forwards only the incoming chunk (the SDK owns accumulation/dedup) and the DeviceStore UI accumulator resets whenever the scanId changes, so a previous scan's delayed chunks can't leak into the current scan's payload (12f79eb9d7, both platforms).
- New additive protocol field `willRetry` on stream `error` statuses: emitted only at the call sites that schedule a reconnect (RTMP/SRT `start_error`/`start_exception`; all other `onStreamError` sites enumerated and classified terminal — WHIP's reconnect path never emits errors). The phone side consuming it lands in #3487; absent field = terminal, matching old-firmware behavior (1a01e414e5).

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
@@ -887,7 +887,7 @@ public class RtmpStreamingService extends Service {
                             mIsStreaming = false;
                             EventBus.getDefault().post(new StreamingEvent.Error(errorMsg));
                             if (sStatusCallback != null) {
-                                sStatusCallback.onStreamError(errorMsg, mCurrentStreamId);
+                                sStatusCallback.onStreamError(errorMsg, mCurrentStreamId, true);
                             }
 
                             // Report stream start failure
@@ -949,7 +949,7 @@ public class RtmpStreamingService extends Service {
             }
             EventBus.getDefault().post(new StreamingEvent.Error(errorMsg));
             if (sStatusCallback != null) {
-                sStatusCallback.onStreamError(errorMsg, mCurrentStreamId);
+                sStatusCallback.onStreamError(errorMsg, mCurrentStreamId, true);
             }
 
             // Report stream start failure
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
@@ -986,6 +986,14 @@ public class RtmpStreamingService extends Service {
     private void forceStopStreamingInternal(boolean preserveSession) {
         Log.d(TAG, "Force stopping stream and cleaning up resources (preserveSession=" + preserveSession + ")");
 
+        // Capture the id up front - cancelStreamTimeout() and the state reset below
+        // both clear it, and the stopped callback must identify the stream being
+        // stopped.
+        final String stoppedStreamId;
+        synchronized (mStateLock) {
+            stoppedStreamId = mCurrentStreamId;
+        }
+
         // Stop battery monitoring if not preserving session
         if (!preserveSession) {
             stopBatteryMonitoring();
@@ -1134,7 +1142,7 @@ public class RtmpStreamingService extends Service {
             Log.d(TAG, "Stream resources released for reconnection");
         } else {
             if (sStatusCallback != null) {
-                sStatusCallback.onStreamStopped(mCurrentStreamId);
+                sStatusCallback.onStreamStopped(stoppedStreamId);
             }
             EventBus.getDefault().post(new StreamingEvent.Stopped());
             Log.i(TAG, "Streaming stopped and cleaned up");
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
@@ -1038,9 +1038,13 @@ public class RtmpStreamingService extends Service {
                     StreamingReporting.reportStreamStopFailure(RtmpStreamingService.this,
                         "stream_stop_error", (Throwable) o);
 
-                    // Notify TPA developer of cleanup failure
+                    // Notify TPA developer of cleanup failure. Use the id captured
+                    // before cleanup: this continuation can resume after the state
+                    // reset cleared mCurrentStreamId or a replacement stream
+                    // overwrote it, and the failure belongs to the stream being
+                    // stopped.
                     if (sStatusCallback != null) {
-                        sStatusCallback.onStreamError("Failed to stop stream: " + ((Throwable) o).getMessage(), mCurrentStreamId);
+                        sStatusCallback.onStreamError("Failed to stop stream: " + ((Throwable) o).getMessage(), stoppedStreamId);
                     }
                 }
                 Log.d(TAG, "Stream stop completed");
--- a/asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/StreamCommandHandler.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/StreamCommandHandler.java
@@ -109,6 +109,7 @@ public class StreamCommandHandler implements ICommandHandler {
     private boolean handleStartCommand(JSONObject data) {
         boolean eisChanged = false;
         boolean streamStarted = false;
+        String streamId = data.optString("streamId", "");
         try {
             // Accept streamUrl first, then legacy rtmpUrl / srtUrl keys
             String streamUrl = data.optString("streamUrl", "");
@@ -118,18 +119,14 @@ public class StreamCommandHandler implements ICommandHandler {
 
             if (streamUrl.isEmpty()) {
                 Log.e(TAG, "Cannot start stream - missing stream URL");
-                streamingManager.sendStreamStatusResponse(
-                        false,
-                        ServiceConstants.STATUS_ERROR,
-                        ServiceConstants.ERROR_MISSING_STREAM_URL);
+                sendStreamErrorStatus(streamId, ServiceConstants.ERROR_MISSING_STREAM_URL);
                 return false;
             }
 
             Protocol protocol = detectProtocol(streamUrl);
             if (protocol == Protocol.UNKNOWN) {
                 Log.e(TAG, "Unknown stream URL protocol: " + streamUrl);
-                streamingManager.sendStreamStatusResponse(
-                        false, ServiceConstants.STATUS_ERROR, "Unknown stream URL protocol");
+                sendStreamErrorStatus(streamId, "Unknown stream URL protocol");
                 return false;
             }
 
@@ -139,9 +136,8 @@ public class StreamCommandHandler implements ICommandHandler {
                 if (batteryLevel >= 0 && batteryLevel < BatteryConstants.MIN_BATTERY_LEVEL) {
                     Log.w(TAG, "🚫 Stream rejected - battery too low (" + batteryLevel + "%)");
                     MediaCaptureService.playBatteryLowSound(context);
-                    streamingManager.sendStreamStatusResponse(
-                            false,
-                            ServiceConstants.STATUS_ERROR,
+                    sendStreamErrorStatus(
+                            streamId,
                             "Battery level too low ("
                                     + batteryLevel
                                     + "%) - minimum "
@@ -156,17 +152,13 @@ public class StreamCommandHandler implements ICommandHandler {
             // WiFi check (WHIP streams may work on mobile data; skip only for WHIP if needed)
             if (stateManager != null && !stateManager.isConnectedToWifi()) {
                 Log.e(TAG, "Cannot start stream - no WiFi connection");
-                streamingManager.sendStreamStatusResponse(
-                        false,
-                        ServiceConstants.STATUS_ERROR,
-                        ServiceConstants.ERROR_NO_WIFI_CONNECTION);
+                sendStreamErrorStatus(streamId, ServiceConstants.ERROR_NO_WIFI_CONNECTION);
                 return false;
             }
 
             // Stop any existing stream
             stopAllServices();
 
-            String streamId = data.optString("streamId", "");
             // Capture light is mandatory for privacy; ignore any caller-supplied flash value.
             boolean flash = true;
             boolean sound = data.optBoolean("sound", true);
@@ -181,7 +173,7 @@ public class StreamCommandHandler implements ICommandHandler {
                 case RTMP:
                     {
                         RtmpStreamConfig config = RtmpStreamConfig.fromJson(videoJson, audioJson);
-                        if (!preflightCameraCaptureForPackStreaming(config)) {
+                        if (!preflightCameraCaptureForPackStreaming(config, streamId)) {
                             return false;
                         }
                         // Toggle EIS for the duration of the stream (see EIS_IN_LIVESTREAMS).
@@ -198,7 +190,7 @@ public class StreamCommandHandler implements ICommandHandler {
                 case SRT:
                     {
                         RtmpStreamConfig config = RtmpStreamConfig.fromJson(videoJson, audioJson);
-                        if (!preflightCameraCaptureForPackStreaming(config)) {
+                        if (!preflightCameraCaptureForPackStreaming(config, streamId)) {
                             return false;
                         }
                         applyEisForStreaming(config.getVideoWidth(), config.getVideoHeight());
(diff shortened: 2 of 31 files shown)


mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/Bridge.kt
```diff
@@ -557,6 +561,9 @@ public class Bridge private constructor() {
             val body = HashMap<String, Any>()
             body["networks"] = updatedNetworks
             body["scanComplete"] = scanComplete
+            if (scanId != null) {
+                body["scanId"] = scanId
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Confirmed and fixed in 12f79eb9d7: when a chunk carries a scanId the Bridge now forwards only that chunk's networks (the SDK's handleCorrelatedWifiScanChunk owns accumulation — verified it dedupes by SSID on both platforms, so the double-accumulation goes away too), and the DeviceStore UI accumulator is scoped per scan: the stored list resets whenever the incoming scanId differs from the stored one, so a previous scan's delayed chunks can never leak into the current scan's forwarded payload. The scanId-less old-firmware path is byte-for-byte unchanged. Applied to both Bridge.kt and Bridge.swift.

asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
```diff
@@ -986,6 +986,14 @@ public void stopStreaming() {
     private void forceStopStreamingInternal(boolean preserveSession) {
         Log.d(TAG, "Force stopping stream and cleaning up resources (preserveSession=" + preserveSession + ")");
 
+        // Capture the id up front - cancelStreamTimeout() and the state reset below
+        // both clear it, and the stopped callback must identify the stream being
+        // stopped.
+        final String stoppedStreamId;
+        synchronized (mStateLock) {
+            stoppedStreamId = mCurrentStreamId;
```
aisraelov, line 994: **P2 — Use the captured ID for asynchronous stop errors**

The `stopContinuation` below still reports failures with `mCurrentStreamId`. That continuation can resume after cleanup has cleared the field or a replacement stream has overwritten it, so stream A’s stop failure can be emitted id-less or attributed to stream B and reject the wrong pending request. Please use `stoppedStreamId` for that callback as well; the SRT continuation has the same issue.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/Bridge.kt
```diff
@@ -542,10 +542,21 @@ public class Bridge private constructor() {
...
+        fun updateWifiScanResults(
+                networks: List<Map<String, Any>>,
+                scanComplete: Boolean,
+                scanId: String? = null
+        ) {
             var storedNetworks: List<Map<String, Any>> =
                     DeviceStore.get("bluetooth", "wifiScanResults") as? List<Map<String, Any>>
                             ?: emptyList()
+            val lastScanId = DeviceStore.get("bluetooth", "wifiScanResultsScanId")
+            if (scanId != null && scanId != lastScanId) {
```
aisraelov: **P2 — Reject stale scan IDs before mutating the observable store**

This makes any different incoming `scanId` the new store owner. If scan B has populated or completed and a delayed chunk/terminal response from scan A arrives afterward, A resets `wifiScanResults` (often to empty). The SDK promise ignores the mismatched ID later, but `DeviceStore` has already changed, so status/UI consumers see stale or cleared results. Please record the active request ID when the scan starts and ignore mismatched IDs for store updates; the iOS Bridge mirrors this issue.

aisraelov (commented):
Two correlation races remain in the current head; details inline.

asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/StreamCommandHandler.java
```diff
@@ -118,18 +119,14 @@ private boolean handleStartCommand(JSONObject data) {
 
             if (streamUrl.isEmpty()) {
                 Log.e(TAG, "Cannot start stream - missing stream URL");
-                streamingManager.sendStreamStatusResponse(
-                        false,
-                        ServiceConstants.STATUS_ERROR,
-                        ServiceConstants.ERROR_MISSING_STREAM_URL);
+                sendStreamErrorStatus(streamId, ServiceConstants.ERROR_MISSING_STREAM_URL);
```
aisraelov, line 122: **[P1] Do not treat an early rejection ID as proof that older starts were preempted**

These ID-bearing errors are emitted before `stopAllServices()` for missing/unknown URLs, low battery, and no WiFi. On the stacked Android base, every ID-bearing status calls `rejectPreemptedStreamStarts` (`MentraBluetoothSdk.kt:1555-1557`) under the assumption that the newer start has already stopped all older ones. If starts A and B are pending and B fails one of these preflight checks, this new error rejects A even though the glasses leave A running. B already stopped the SDK keep-alive monitor, and A can no longer resolve and start a replacement monitor, so the live stream can subsequently time out. Please exclude error statuses from the phone-side preemption inference (the old stream now has its own ID-bearing `stopped` verdict), or explicitly distinguish errors emitted after preemption. The same interaction will apply to iOS when #3487 is combined.

mobile/modules/bluetooth-sdk/ios/Source/Bridge.swift
```diff
@@ -432,10 +448,16 @@ class Bridge {
...
+                // Correlated scans: the SDK accumulates and dedupes chunks per scanId itself,
+                // so forward only this chunk; the merged store list is for UI consumers.
+                var body: [String: Any] = [
+                    "networks": scanId != nil ? networks : storedNetworks,
+                    "scanComplete": scanComplete,
+                ]
+                if let scanId {
+                    body["scanId"] = scanId
+                }
+                Bridge.sendTypedMessage("wifi_scan_result", body: body)
```
aisraelov, line 485: **[P2] Preserve scan chunk order through the event sink as well**

This makes the first hop to the main queue FIFO, but `sendTypedMessage` then invokes the SDK sink, and that sink wraps every event in a separate `Task { @MainActor ... }` (`MentraBluetoothSDK.swift:203-206`). Those unstructured tasks do not provide a FIFO ordering guarantee, so the terminal empty chunk can still reach `handleCorrelatedWifiScanChunk` before a preceding network chunk and resolve a partial result. Since `dispatchEvent` already guarantees that the sink runs on the main thread, please dispatch the SDK event synchronously on MainActor there (or use one serial consumer) rather than creating another task per message.

asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/WifiCommandHandler.java
```diff
@@ -277,30 +282,33 @@ public boolean handleRequestWifiScan() {
                             public void onNetworksFoundEnhanced(List<NetworkInfo> networks) {
                                 Log.d(TAG, "📡 Streaming " + networks.size() + " enhanced WiFi networks to phone");
                                 // Send each batch of networks immediately as they're found
-                                communicationManager.sendWifiScanResultsOverBleEnhanced(networks, false);
+                                communicationManager.sendWifiScanResultsOverBleEnhanced(networks, false, scanId);
```
aisraelov: **[P2] Serialize glasses-side scans before assigning correlation IDs**

Capturing `scanId` in each callback does not make overlapping Android scans attributable. A cancelled phone request can clear its pending entry while the glasses-side scan is still waiting up to 15 seconds, allowing a new request to start another thread here. `BaseNetworkManager.scanWifiNetworks` registers one receiver per call for the global `SCAN_RESULTS_AVAILABLE_ACTION`, so the first broadcast wakes both receivers and both callbacks stamp the same system result with their own IDs. The phone then accepts the result labeled with the newer ID even if it came from the older scan, and the newer receiver unregisters before its own scan broadcast. Please make this handler/network manager single-flight (or otherwise serialize/coalesce scans) before echoing per-request IDs.

asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/StreamCommandHandler.java
PhilippeFerreiraDeSousa: Sharp catch — with this branch's id-stamped preflight rejections the inference premise breaks exactly as you describe. Fixed in 8e59d2bfcc: rejectPreemptedStreamStarts is now gated on an explicit allow-list of states that prove the glasses ran the start path past stopAllServices (initializing, reconnecting, reconnected, streaming); error and stopped are excluded with the reason in a comment. The iOS mirror got the identical restriction on #3487 (85ac3f05b4), so the combined merge is covered in either order. The preempted older start's verdict comes from its own id-stamped stopped, as you note.

asg_client/app/src/main/java/com/mentra/asg_client/io/streaming/services/RtmpStreamingService.java
PhilippeFerreiraDeSousa: Fixed in 4d193d1d78 — the RTMP and SRT stop continuations now report failures with the captured stoppedStreamId. (The stopPreview/release error callbacks run synchronously before the state reset clears the field, so they keep mCurrentStreamId.)

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/Bridge.kt
PhilippeFerreiraDeSousa: Fixed in d00ae3a882: ownership is now claimed at request time — requestWifiScan records its generated scanId as the active store owner (both platforms) before sending the command, and updateWifiScanResults refuses store mutation for any chunk whose scanId doesn't match, so a delayed old-scan chunk can no longer reset or clobber the current scan's store (chunks still flow to the SDK sink, which drops stale ids). Old-firmware id-less path unchanged.

mobile/modules/bluetooth-sdk/ios/Source/Bridge.swift
PhilippeFerreiraDeSousa: Fixed in 255e50c442: the sink now consumes events synchronously under MainActor.assumeIsolated (dispatchEvent already guarantees main-thread delivery; the fallback hop is DispatchQueue.main.async, not a Task), so chunk order is preserved end to end. Audited the inline delivery for re-entrancy: the SDK sink is the only registered sink, its handlers only resolve PendingResponses (continuation resumes are scheduled, never run inline) and forward to Expo sendEvent — no handler emits further Bridge events. One residual out of scope: the DeviceStore change-listener still uses per-event Tasks; store updates aren't chunk-ordered.

asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/WifiCommandHandler.java
PhilippeFerreiraDeSousa: Fixed in ecea82fe30: the scan is now single-flight behind a lock — a request arriving mid-run doesn't spawn a second receiver, it adopts the running scan by swapping in its id for all subsequent emissions. Rationale in the code: the phone joins concurrent scan requests under one scanId, so a second distinct id can only mean the older request was abandoned; newest-id adoption serves the only live requester without double-sending chunks over BLE. Terminal emission reads the adopted id atomically and releases the slot.

asg_client/app/src/main/java/com/mentra/asg_client/service/core/handlers/StreamCommandHandler.java
PhilippeFerreiraDeSousa: Update: this restriction has been hoisted out of #3488 into the base PR #3474 (commit e4034912bb), since `rejectPreemptedStreamStarts` is defined there and the guard should be unconditional phone logic, not coupled to this PR's firmware id-echo. #3488 was rebased onto the new base and no longer carries the commit — the remaining commits' shas changed as a result. The behavior is unchanged; the guard just lives one PR down the stack now.

PR #3500, opened 2026-07-20
Improve Miniapp SDK audio performance

CaydenPierce:
## Summary

- Stabilize Mentra Live audio routing: prewarm cold A2DP routes so we don't miss the first bit of audio that is played back due to A2DP warming up, reuse playback safely, and prevent mic watchdog/audio-route races across Android and iOS.
- Keep LC3 capture alive during phone playback; suppress cloud STT only for spoken TTS as a stand-in for echo cancellation for right now - just a stand-in until we have it working better on the hardware - it breaks barge-in for agents but its better than echo - only for TTs so `speaker.play()` cues and PCM sound effects do not stop listening.
- Consolidate sentence TTS through `speaker.speak()`: local TTS generates/plays sentences sequentially, while cloud TTS receives the joined text.
- Make camera FOV configuration compatible with older glasses firmware by falling back to the legacy command when the newer override is unavailable... backwards compatability

## Why

This keeps voice interactions responsive and avoids TTS echo without sacrificing immediate sound cues, while protecting against route/mic recovery races and supporting older glasses firmware.

## Validation

- `bun test modules/engine/src/services/__tests__/AudioPlaybackService.test.ts`
- `bun compile`
- Android compile/build verification on a connected device.

---

> [!NOTE]
> **Medium Risk**
> Touches real-time audio (A2DP prewarm, LC3 suspend, cloud uplink gating) and camera FOV on device; regressions could affect barge-in, STT echo, or capture timing on older firmware.
> 
> **Overview**
> Improves **Mentra Live** voice and playback behavior while adding **legacy camera FOV** support for older ASG firmware.
> 
> **Audio routing and playback:** On Android, cold A2DP routes get a short silent PCM **prewarm** before URL playback or live PCM streams so the first audible frame is not clipped. Playback tracks a warm window, defers player pause until an A2DP tail drain (~700ms), and exposes optional **`suppressCloudUplink`** so only opted-in playback blocks cloud STT—not generic `play()` or PCM streams.
> 
> **Cloud STT vs TTS:** `AudioCloudUplink` reference-counts suppression sources and still forwards LC3 locally but drops frames to the cloud while any source is active. TTS paths (`LocalMiniappRuntime` speak, sentence offline playback) opt in; tail suppression can clear early when a non-TTS PCM cue starts. Native **`isMicSuspendedForAudio`** on `SGCManager`/`MentraLive` stops the glasses mic watchdog from fighting intentional LC3 suspend during phone audio.
> 
> **Speaker / TTS API:** `speaker.speak()` accepts a string or explicit sentence list; offline/cloud pipelines use that list (cloud gets joined text). Removed `segmentTextForOfflineTts` from the sentence pipeline.
> 
> **Camera FOV:** Bluetooth SDK adds **`setLegacyCameraFov`** / **`restoreLegacyCameraFov`** (Android/iOS/TS). `PhoneCameraFovCoordinator` falls back when lease overrides time out or are unavailable, waits ~7s after legacy HAL restarts, and clears legacy mode after a successful restore cycle so modern paths can be retried.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3191ebfdd9958b2066e8c6d9be74f023286d4da4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CaydenPierce:
Let's please test on iOS

PR #3502, opened 2026-07-20
Add AR99-family smart glasses support

2423043814:
add AR99 smart glasses

---

> [!NOTE]
> **High Risk**
> Large new BLE pairing, firmware OTA, and audio paths in the core glasses stack; incorrect gating or OTA bugs could brick pairing or devices for AR99 users.
> 
> **Overview**
> **Adds first-class support for Xingyi AR99** (`DeviceTypes.AR99`) in shared types and hardware capabilities (G1-like text display, no bitmaps, on-glasses mic, OTA), with compatibility docs that only allow BLE project id `AR99` and reject other prefixes.
> 
> **Native bluetooth-sdk (Android + iOS)** gets a new `Ar99` SGC: LE scan/connect with manufacturer-data parsing, `projectName` on discovery and default-device storage, MAC-based reconnect for AR99, minimal-text display, Opus→PCM mic via Concentus, and a dedicated GATT OTA stack (`Ar99OtaManager`, `ar99_ota_status` events). New bridge APIs: `startAr99OtaFromFile`, `cancelAr99Ota`, `sendAr99FactoryReset`, and `buildAr99OtaSignature` (MD5).
> 
> **Metro** watches `@mentra/crust` and aliases `@mentra/miniapp` / `@mentra/crust` to source for local dev bundling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b9bc525b726c5cf78dd8f8d804428a1ade292d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Adds first‑class AR99 smart glasses (Xingyi) support across cloud, native SDKs, and app: gated scan/pairing by projectName=AR99, Opus→PCM mic, OTA (check/download/transfer with status), and Factory Reset, with updated UI/assets and localized strings. Also fixes AR99 scan/reconnect issues, resolves Android BLE vs Wi‑Fi scan conflicts, normalizes miniapp schema ASCII, and makes postinstall reliable on Windows.

- **New Features**
  - Cloud/engine: added `DeviceTypes.AR99` with capabilities (text display only, mic, `hasOta`); compatibility docs restrict to `AR99` project identifier.
  - Native SDKs (Android/iOS): AR99 drivers with scan filter by project identifier; search results/defaults persist `projectName`; mic audio via Concentus; OTA with `ar99_ota_status` and APIs `startAr99OtaFromFile`, `cancelAr99Ota`, `sendAr99FactoryReset`, `buildAr99OtaSignature`.
  - App: AR99 pairing/settings with model assets and names; flows pass `ar99ProjectName`; `Ar99OtaModal` adds version check and firmware download service; Device Info exposes Factory Reset; localized onboarding/warnings; Metro watches/aliases `@mentra/crust` and `@mentra/miniapp`; tests cover advertisement gating.

- **Bug Fixes**
  - Tightened advertisement parsing and scan gating (accept only `AR99`); improved reconnect by MAC and fixed discovery timeouts.
  - OTA reliability (status, cancel, pause/resume on reconnect); prevented Android BLE scans from overriding Wi‑Fi scans; resolved mobile build config issues.
  - Normalized miniapp schema descriptions to ASCII; Windows‑aware postinstall via zx shell setup.

<sup>Written for commit 2b9bc525b726c5cf78dd8f8d804428a1ade292d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3502?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mobile/modules/bluetooth-sdk/ios/Source/DeviceManager.swift
```diff
@@ -557,7 +557,7 @@ struct ViewState {
     private func playStartupSequence() {
         Bridge.log("MAN: playStartupSequence()")
         // Arrow frames for the animation
-        let arrowFrames = ["↑", "↗", "↑", "↖"]
+        let arrowFrames = ["�?, "�?, "�?, "�?]
```
aisraelov: **Blocking: this source file is not valid UTF-8.** I reproduced this with `xcrun swiftc -frontend -parse`; it fails here and at numerous other locations. This executable string literal has also been corrupted, so this is not cosmetic comment damage. In total, 13 changed Kotlin/Swift/TypeScript source files fail a strict UTF-8 read. Please restore those files from current `dev` and reapply only the intended edits using a UTF-8-safe tool before asking CI or reviewers to evaluate the implementation.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/ar99/ota/Ar99OtaManager.java
```diff
@@ -0,0 +1,532 @@
...
+import android.os.Handler;
+import android.os.Looper;
+import android.util.Log;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.Arrays;
+
+public class Ar99OtaManager {
```
aisraelov, line 13: **Blocking organization/ownership concern:** this manager and the five accompanying `Ota*` classes are AR99 vendor protocol code, but they are being introduced in the generic `com.mentra.bluetoothsdk.ota` namespace with names such as `OtaProtocol` and `OtaCommandConstants`. That makes them look like shared Mentra OTA infrastructure and creates an obvious collision point for the next device. Please move the complete integration into a device-owned package such as `sgcs/ar99/{ble,audio,ota}` (and mirror that structure on iOS/JS/assets). Only an interface that is genuinely reusable across hardware should remain in a generic namespace.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/Ar99.java
```diff
@@ -0,0 +1,2654 @@
...
+  private static final long BATTERY_POLL_INTERVAL_MS = 60_000L;
+  private static final int BLE_WRITE_CHUNK_SIZE = 180;
+  private static final long BLE_WRITE_CHUNK_DELAY_MS = 10L;
+  private static final long READY_MINIMAL_MODE_DELAY_MS = 40L;
+  private static final long READY_LANGUAGE_DELAY_MS = 90L;
+  private static final long READY_DEVICE_INFO_DELAY_MS = 140L;
+  private static final long READY_BRIGHTNESS_DELAY_MS = 280L;
+  private static final long READY_BATTERY_DELAY_MS = 420L;
+  private static final long DISPLAY_SESSION_IDLE_RESTART_MS = 8_000L;
+  private static final String[] NAME_PREFIXES = new String[] {"AR99", "AF98", "AF99", "HVXM", "HVXF"};
```
aisraelov: What exact hardware is this PR claiming to support? The PR and final commit say AR99, this native filter silently accepts AR99/AF98/AF99/HVXM/HVXF, the UI metadata omits AF98, and model selection exposes only one Xingyi option while retaining Holovox assets and behavior. A prefix match is effectively a compatibility claim. Please define and document an explicit manufacturer/model/project matrix, remove out-of-scope prefixes, and add advertisement fixtures for every value that remains. We should not accidentally pair an unvalidated OEM variant under the AR99 label.

mobile/src/app/pairing/select-glasses-model.tsx
```diff
@@ -47,38 +74,43 @@ export default function SelectGlassesModelScreen() {
...
+  ]
 
-  const triggerGlassesPairingGuide = async (deviceModel: string) => {
-    push("/pairing/prep", {deviceModel: deviceModel})
+  const glassesOptions = Platform.OS === "ios" ? sharedOptions : sharedOptions
+
+  const triggerGlassesPairingGuide = async (option: GlassesOption) => {
+    push("/pairing/prep", {
+      deviceModel: option.deviceModel,
+      ar99ProjectName: option.deviceModel === DeviceTypes.AR99 ? undefined : option.projectName,
```
aisraelov: **Blocking pairing bug:** this condition is backwards. `projectName` exists on the AR99 options, but this sends `undefined` specifically when AR99 is selected. The scan screen treats an empty `ar99ProjectName` as a wildcard, so selecting the displayed `Xingyi AR99` option can list and pair AF99/HVXM/HVXF/AF98 devices as well. Pass the selected project through and make the scan filter fail closed; do not collapse distinct OEM products into a wildcard pairing flow.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/Bridge.kt
```diff
@@ -262,7 +263,7 @@ public class Bridge private constructor() {
                                         ?.toMap()
                             }
                             ?: emptyList()
-            val id = "$deviceModel:$deviceName"
+            val id = listOfNotNull(deviceModel, projectName?.takeIf { it.isNotBlank() }, deviceName).joinToString(":")
```
aisraelov, line 266: This still does not produce the stable device identity promised by the public `Device` type. Two devices with the same project and display name collapse into one search row, even when they have different addresses; subsequent RSSI updates can then overwrite the wrong device. `deviceAddress` is already available here and should be the primary key when present, with a namespaced fallback only when the platform provides no identifier. Please make the iOS key follow the same rule.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/Ar99.java
```diff
@@ -0,0 +1,2654 @@
...
+
+  private final BluetoothGattCallback gattCallback =
+      new BluetoothGattCallback() {
+        @Override
+        public void onConnectionStateChange(BluetoothGatt gatt, int status, int newState) {
+          if (newState == BluetoothProfile.STATE_CONNECTED) {
+            Bridge.log(TAG + ": connected to " + (gatt.getDevice() != null ? gatt.getDevice().getName() : ""));
+            controlGatt = gatt;
+            gatt.discoverServices();
+          } else if (newState == BluetoothProfile.STATE_DISCONNECTED) {
```
aisraelov, line 274: **Connection race:** this callback never verifies that `gatt` is still the active `controlGatt`. `connectGatt()` closes the previous object and assigns a new one, but the old object's asynchronous DISCONNECTED callback can arrive afterward; this branch then calls `cleanupGatt()` on the new global connection and tears it down. Guard every callback with connection identity/generation before mutating shared fields. The iOS `didDisconnectPeripheral` path has the same stale-peripheral problem.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/Ar99.java
```diff
@@ -0,0 +1,2654 @@
...
+
+          if (notifyQueue.isEmpty()) {
+            markReady();
+          }
+        }
+
+        @Override
+        public void onDescriptorWrite(BluetoothGatt gatt, BluetoothGattDescriptor descriptor, int status) {
+          isNotifyWriteInFlight = false;
+          if (status != BluetoothGatt.GATT_SUCCESS) {
```
aisraelov, line 317: A failed descriptor write is only logged, after which the queue advances and can call `markReady()`. For the control characteristic this marks the glasses connected/fully booted even though responses, battery, version info, and display acknowledgements cannot arrive. For the OTA characteristic it leaves `Ar99OtaManager` holding firmware in IDLE with no timeout because `onOtaNotifyEnabled()` is never called. Required notification failures must fail the connection or OTA attempt and surface an actionable error; optional Opus notification failure should at least disable the microphone capability rather than report a fully working device.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/Ar99.java
```diff
@@ -0,0 +1,2654 @@
...
+          processNextNotifyDescriptor();
+        }
+
+        @Override
+        public void onCharacteristicWrite(BluetoothGatt gatt, BluetoothGattCharacteristic characteristic, int status) {
+          if (characteristic != null && OTA_WRITE_UUID.equals(characteristic.getUuid())) {
+            synchronized (otaWriteQueue) {
+              isOtaWriteInFlight = false;
+            }
+            if (status != BluetoothGatt.GATT_SUCCESS) {
```
aisraelov, line 340: **Blocking OTA data-loss path:** the packet was removed from `otaWriteQueue` before this callback. On a failed GATT write we only log, then send the next packet, so the native manager continues as though a missing firmware frame reached the glasses. This must either retry with a bounded policy or transition the OTA state machine to failed and clear the queue. The equivalent iOS callback also logs and advances after `.withResponse` failures.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/Ar99.java
```diff
@@ -0,0 +1,2654 @@
...
+    ScanCallback cb = scanCallback;
+    if (scanner != null && cb != null) {
+      try {
+        scanner.stopScan(cb);
+      } catch (Throwable t) {
+        // ignore
+      }
+    }
+
+    handler.removeCallbacksAndMessages(null);
```
aisraelov: `handler` is not scan-owned: it also runs BLE write drains, ready/setup commands, display work, and OTA queue work. `removeCallbacksAndMessages(null)` therefore cancels unrelated device operations whenever a scan stops, which can strand a write with no future drain or prevent connection initialization. Keep a dedicated `scanTimeoutRunnable` (or a dedicated handler) and remove only that callback.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/ar99/ota/OtaProtocol.java
```diff
@@ -0,0 +1,89 @@
...
+    payload.write(0x00);
+    payload.write(0x01);
+    payload.write(0x0A);
+    payload.write(0x04);
+    payload.write(0x00);
+    payload.write(0x00);
+    payload.write(0x00);
+    payload.write(0x00);
+    payload.write(0x00);
+    return buildFrame(OtaCommandConstants.OTA.REQUEST_UPGRADE, (byte) 0x80, payload.toByteArray());
```
aisraelov, line 38: **Blocking protocol defect or undocumented contract:** `packageSize`, `blockSize`, and `crc32` are accepted by this method but none is serialized; every firmware file produces the exact same fixed request. `firmwareCrc32` is otherwise dead, and `buildConnectNegotiation` also ignores its block-size argument. Either the vendor protocol implementation is incomplete, or these signatures are misleading. Please implement the actual wire fields and add byte-for-byte fixtures from the vendor protocol. If the fixed bytes are genuinely correct, remove the dead parameters/calculation and document the frame layout so this can be reviewed safely.

mobile/modules/bluetooth-sdk/ios/Source/sgcs/ar99/ota/Ar99OtaManager.swift
```diff
@@ -0,0 +1,623 @@
...
+        let length = payload?.count ?? 0
+        frame.append(UInt8(length & 0xFF))
+        frame.append(UInt8((length >> 8) & 0xFF))
+        if let payload, !payload.isEmpty {
+            frame.append(payload)
+        }
+        return frame
+    }
+
+    static func buildRequestUpgrade(packageSize _: Int, blockSize _: Int, crc32 _: UInt32) -> Data {
```
aisraelov, line 109: The same OTA request problem exists independently on iOS: all three inputs are explicitly discarded and the request is hardcoded. Android and iOS must be validated against the same authoritative protocol vectors; copying the same unexplained constant onto both platforms does not establish parity or correctness.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/ar99/ota/Ar99OtaManager.java
```diff
@@ -0,0 +1,532 @@
...
+      handleOTAFailure(OtaCommandConstants.OTA.ErrorCode.UNKNOWN_ERROR, "MTU negotiation failed");
+      return true;
+    }
+    onMtuNegotiated();
+    return true;
+  }
+
+  public void handleOTAResponse(byte[] data) {
+    if (data == null || data.length < 5) return;
+    cancelTimeout();
```
aisraelov: The watchdog is cancelled before the frame is validated or even confirmed to be an OTA service/known command. A malformed, truncated-payload, wrong-service, or unknown notification can therefore disable the only timeout and leave the state machine hung forever. Parse and validate the complete frame and confirm that it is the response expected for the current state before cancelling/rearming the timeout. The iOS implementation has the same ordering.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/ar99/ota/Ar99OtaManager.java
```diff
@@ -0,0 +1,532 @@
...
+      handleOTAFailure(OtaCommandConstants.OTA.ErrorCode.PROTOCOL_ERROR, "Invalid firmware offset");
+      return;
+    }
+    if (fileOffset >= imageSize) return;
+
+    int remaining = imageSize - fileOffset;
+    int totalSendLength = fileLength > 0 ? Math.min(remaining, fileLength) : remaining;
+    int sentLength = 0;
+    int psn = 0;
+    while (sentLength < totalSendLength) {
```
aisraelov, line 383: This eagerly materializes and queues every requested firmware block in one pass while the complete firmware is already retained in memory. If the device requests the remaining image, a normal firmware file becomes tens of thousands of arrays/frames and roughly duplicates its memory footprint on a mobile device. It also sets `currentOffset` and reports progress for bytes merely enqueued, not bytes accepted by GATT/the glasses, so progress can jump to 100% while transmission is still running. Use bounded backpressure (one/few acknowledged frames at a time), stream from the file, and advance progress only from confirmed transport/protocol acknowledgements. iOS duplicates this behavior.

mobile/modules/bluetooth-sdk/ios/Source/sgcs/Ar99.swift
```diff
@@ -0,0 +1,2449 @@
...
+        var writeType: CBCharacteristicWriteType = .withResponse
+        if characteristic.properties.contains(.writeWithoutResponse) {
+            let maxWithoutResponse = peripheral.maximumWriteValueLength(for: .withoutResponse)
+            if otaWriteQueue[0].count <= maxWithoutResponse {
+                writeType = .withoutResponse
+            }
+        }
+
+        let maxLength = peripheral.maximumWriteValueLength(for: writeType)
+        guard otaWriteQueue[0].count <= maxLength else {
```
aisraelov, line 1454: This emits a UI `failed` event but never tells `Ar99OtaManager` to enter FAILED/cancel. The manager therefore remains in progress, retains the full firmware buffer, and may have no active timeout; the queue is left with the remaining frames. Please make transport errors flow through a single manager failure transition that cancels timers, clears queued data, releases firmware, and emits exactly one terminal event. Also negotiate/clamp block size before queuing so this path is not the normal result on peripherals with a smaller maximum write length.

mobile/src/services/ar99Ota.ts
```diff
@@ -0,0 +1,154 @@
...
+  const data = payload.data ?? {}
+  const latest = data.current_version?.trim() ?? ""
+  const url = data.url?.trim() ?? ""
+  return {
+    changeLog: data.change_log ?? "",
+    currentVersion: latest,
+    fileMd5: data.md5?.trim().toLowerCase() ?? "",
+    firmwareUrl: resolveFirmwareUrl(url),
+    forceUpdate: data.force_update === true,
+    hasUpdate: url.length > 0 && latest.length > 0 && latest !== version,
```
aisraelov, line 97: `latest !== version` treats every different string as an upgrade, including an older firmware version or a differently formatted equivalent. A stale/mis-scoped vendor response can therefore offer and flash a downgrade. Use an explicit server-provided update decision or a documented version comparator with downgrade policy; add tests for equal, newer, older, malformed, and vendor-prefixed versions.

mobile/src/services/ar99Ota.ts
```diff
@@ -0,0 +1,154 @@
...
+      }
+    },
+  })
+
+  const result = await promise
+  if (result.statusCode < 200 || result.statusCode >= 300) {
+    throw new Error(`Firmware download failed: ${result.statusCode}`)
+  }
+
+  if (expectedMd5) {
```
aisraelov, line 127: **Blocking OTA safety issue:** verification is optional. When `md5` is absent, the app proceeds directly to flashing, and `resolveFirmwareUrl` also accepts plain `http://`. At minimum this flow must fail closed on a missing/malformed digest and require HTTPS. More importantly, a transport hash such as MD5 is not firmware authenticity; before Mentra exposes this updater, require the vendor's signed-package verification (ideally against a pinned public key) and test tampered/truncated downloads.

mobile/modules/bluetooth-sdk/src/BluetoothSdk.types.ts
```diff
@@ -1036,6 +1049,10 @@ export interface BluetoothSdkPublicModule {
   checkForOtaUpdate(): Promise<boolean>
   /** Start the OTA flow with the same configured manifest URL used by checkForOtaUpdate(). */
   startOtaUpdate(): Promise<OtaStartAckEvent>
+  startAr99OtaFromFile(path: string): Promise<boolean>
+  cancelAr99Ota(): Promise<void>
+  sendAr99FactoryReset(): Promise<void>
+  buildAr99OtaSignature(secret: string, appName: string, currentVersion: string, serialNumber: string, nonce: string): string
```
aisraelov, line 1140: This HTTP-service signing primitive does not belong in the public Bluetooth SDK API. It couples every SDK consumer to one vendor's OTA backend, asks callers to pass a shared secret through the native bridge, and permanently expands Mentra's public surface for an app-internal concern. Keep the public Bluetooth boundary focused on device transport/commands; put vendor OTA service integration behind an AR99-specific internal adapter or trusted server API. The same applies to deciding which AR99-only methods truly need to be public SDK commitments.

mobile/src/app/pairing/prep.tsx
```diff
@@ -443,6 +445,31 @@ export default function PairingPrepScreen() {
...
+            <Image
+              source={require("../../../assets/guide/image_ar99_pair.png")}
+              resizeMode="contain"
+              className="w-full h-full"
+            />
+          </View>
+          <Text tx="pairing:instructions" className="text-2xl font-bold mb-4 text-secondary-foreground" />
+          <Text
+            className="text-lg text-secondary-foreground mb-2"
+            text="1. Confirm that the Bluetooth switch of the mobile phone is turned on."
```
aisraelov: Please move the new AR99 pairing instructions (and the many strings in `Ar99OtaModal`) into the translation catalog instead of hardcoding English. This repository's mobile guidelines require user-facing strings to use `tx`/`translate`, and adding a new hardware flow should not make localization coverage regress. Also have the device vendor confirm that these instructions apply to every project identifier this PR claims to accept.

aisraelov (changes requested):
Thanks for contributing AR99 support. We are open to third-party glasses integrating with the Mentra App, but this implementation is not ready to merge.

The main blockers are: the branch currently contains invalid UTF-8 that breaks compilation; AR99-specific code is spread through generic Mentra OTA/audio/service namespaces; the two OTA implementations have protocol and state-machine paths that can hang, silently lose writes, or report progress before bytes reach the glasses; the selected OEM/project is discarded by the pairing route; and the claimed device support matrix is inconsistent. The existing automated findings about the mic watchdog and incorrect inherited capabilities are also valid and need to be fixed.

This branch also has substantial content conflicts with current `dev`, including the `island` to `engine` migration and recent Bluetooth/pairing work. Please rebase before continuing so conflict resolution does not overwrite current Mentra behavior.

Because this adds roughly 8,100 lines across BLE, audio decoding, pairing, reconnect, factory reset, and firmware update paths, we need tests before accepting it: byte-for-byte OTA protocol fixtures on both platforms, OTA state-machine failure/reconnect/cancel coverage, advertisement fixtures for every supported project identifier, pairing isolation tests, audio/watchdog coverage, Android and iOS build evidence, and a documented real-hardware validation matrix. OTA code that can brick a third-party device cannot be merged based only on manual happy-path testing.

Please keep the ownership boundary explicit: AR99 is third-party hardware supported by the Mentra App, not a Mentra product. Device-specific protocol, transport, audio, OTA, services, and assets should live under an AR99-specific namespace/folder. Shared Mentra surfaces should contain only the narrow adapter hooks that are genuinely shared.

2423043814:
Updated the AR99 integration based on the review feedback:

- Removed unsupported `AF98`.
- Explicitly support only `AR99`, `AF99`, `HVXM`, and `HVXF`.
- Restored model selection to show the four AR99-family variants.
- Preserved the selected `projectName` through pairing.
- Made scan filtering fail closed by requiring the parsed advertisement `projectName` to match the selected model.
- Removed AR99-family fallback matching from BLE local name.
- Added compatibility matrix documentation and advertisement/pairing fixtures.
- Fixed AR99 audio activity reporting for the mic watchdog.
- Updated AR99 capabilities: `hasOta: true`, `canDisplayBitmap: false`.
- Kept AR99 out of the generic Mentra Live OTA/onboarding route after pairing success.
- Fixed invalid UTF-8/corrupted Swift literals in changed source files.
- Moved Android AR99 OTA implementation out of the generic `com.mentra.bluetoothsdk.ota` namespace into `com.mentra.bluetoothsdk.sgcs.ar99.ota`.
- Moved iOS `Ar99OtaManager.swift` from `Source/ota` to `Source/sgcs/ar99/ota`.
- Kept OTA protocol behavior unchanged during the namespace/path cleanup.

2423043814:
Added OTA state-machine fixes for the latest Cursor findings:

- Added an Android reconnect timeout while OTA is paused in `PAUSED_DISCONNECTED`.
- Re-armed the OTA watchdog after transfer progress responses on Android and iOS.
- Kept invalid/non-OTA notifications from clearing the active OTA timeout.
- Kept AR99 OTA on `WRITE_TYPE_NO_RESPONSE` for throughput.
- Removed duplicate Android OTA queue draining by ignoring OTA write callbacks while no-response mode owns the fast `handler.post` drain path.

mobile/modules/bluetooth-sdk/ios/Source/Bridge.swift
```diff
@@ -441,7 +444,7 @@ class Bridge {
         Bridge.sendTypedMessage("mtk_update_complete", body: eventBody)
     }
 
-    /// Send ota_start_ack — glasses confirmed receipt of ota_start command
+    /// Send ota_start_ack �?glasses confirmed receipt of ota_start command
```
aisraelov, line 501: **Blocking build regression remains:** this current-head file still contains invalid UTF-8 here and fails `xcrun swiftc -frontend -parse`. A strict UTF-8 scan finds 11 files changed by this PR that went from valid to invalid. Please restore those files from clean current `dev` versions and reapply only the intended edits; the latest partial character substitutions did not repair the source safely.

sdk/miniapp-cli/schema/miniapp.schema.json
```diff
@@ -64,7 +64,7 @@
         },
         "ui": {
           "type": "string",
-          "description": "Path to the UI bundle entry HTML, relative to the bundle root (e.g. ui/index.html). Optional — pure-background miniapps don't need a WebView."
+          "description": "Path to the UI bundle entry HTML, relative to the bundle root (e.g. ui/index.html). Optional �?pure-background miniapps don't need a WebView."
```
aisraelov: This is a new, unrelated regression in the follow-up. The committed schema contains mojibake here and again at line 172, while `generateSchemaString()` now emits different text, so the generated artifact is out of sync with its source. Please revert the Miniapp CLI changes from this hardware PR or regenerate the artifact exactly from the generator.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/ar99/ota/OtaProtocol.java
```diff
@@ -0,0 +1,89 @@
...
+    if (payload != null && payload.length > 0) {
+      try {
+        out.write(payload);
+      } catch (Exception ignored) {
+      }
+    }
+    return out.toByteArray();
+  }
+
+  public static byte[] buildRequestUpgrade(int packageSize, int blockSize, long crc32) {
```
aisraelov, line 25: To clarify the requested change: Mentra does not need to invent or redesign this vendor protocol. But all three inputs remain unused, so the implementation cannot be reviewed from its API. If this exact constant frame is vendor-defined, please remove the misleading arguments/dead firmware CRC, document the frame fields, and add an authoritative byte-for-byte fixture shared by the Android and iOS tests. Otherwise serialize the intended fields.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/ar99/ota/Ar99OtaManager.java
```diff
@@ -0,0 +1,566 @@
...
+      sendOtaData(OtaProtocol.buildSendImageData(psn, block));
+      sentLength += size;
+      psn++;
+    }
+
+    synchronized (this) {
+      currentOffset = fileOffset + sentLength;
+    }
+    notifyProgress(currentOffset, getProgress());
+    startTimeout();
```
aisraelov, line 396: Rearming the watchdog here fixes the previous no-timeout gap, but it does not make this transport progress. The loop above has already materialized and queued the entire requested range, and `currentOffset`/UI progress describe bytes queued in phone memory rather than bytes accepted by GATT or the glasses. Please keep the vendor wire protocol as required but feed the native transport through a bounded pump and advance progress from the relevant acknowledgement.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/Ar99.java
```diff
@@ -0,0 +1,2655 @@
...
+        @Override
+        public void onCharacteristicWrite(BluetoothGatt gatt, BluetoothGattCharacteristic characteristic, int status) {
+          if (characteristic != null && OTA_WRITE_UUID.equals(characteristic.getUuid())) {
+            synchronized (otaWriteQueue) {
+              if (otaWriteUsesNoResponseMode) {
+                return;
+              }
+              isOtaWriteInFlight = false;
+            }
+            if (status != BluetoothGatt.GATT_SUCCESS) {
```
aisraelov, line 340: This still skips a failed firmware frame. The packet was removed from `otaWriteQueue` before the callback; a non-success status is logged and the next packet is sent, while `Ar99OtaManager` remains active. This is host-side Mentra integration behavior, independent of the vendor wire protocol: retry with a bounded policy or invoke one terminal manager failure that also clears the transport queue. The equivalent iOS callback needs the same treatment.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/Ar99.java
```diff
@@ -0,0 +1,2655 @@
...
+    ScanCallback cb = scanCallback;
+    if (scanner != null && cb != null) {
+      try {
+        scanner.stopScan(cb);
+      } catch (Throwable t) {
+        // ignore
+      }
+    }
+
+    handler.removeCallbacksAndMessages(null);
```
aisraelov: This shared-handler cancellation is still present. `handler` also owns connection-ready commands, display/write drains, and OTA queue work, so stopping a scan can cancel unrelated operations and strand their state machines. Store the scan timeout runnable and remove only that runnable, or use a scan-owned handler.

mobile/src/services/ar99Ota.ts
```diff
@@ -0,0 +1,153 @@
...
```
aisraelov, line 127: The vendor credentials themselves can remain the vendor's responsibility, but flashing behavior inside the Mentra App is our boundary. A missing digest currently skips verification, and `resolveFirmwareUrl` accepts plain HTTP. Please require HTTPS and fail closed when integrity/authenticity metadata is absent or malformed; use a vendor-signed package/public-key check if their OTA format supports one.

mobile/src/components/settings/Ar99OtaModal.tsx
```diff
@@ -0,0 +1,325 @@
+import {useCallback, useEffect, useMemo, useRef, useState} from "react"
+import {ActivityIndicator, Modal, ScrollView, TouchableOpacity, useWindowDimensions, View} from "react-native"
+
+import BluetoothSdk, {Ar99OtaStatusEvent} from "@mentra/bluetooth-sdk"
+import {toolkit} from "@mentra/island"
+
+import {Text} from "@/components/ignite"
+import {Ar99VersionInfo, checkAr99OtaVersion, clearAr99OtaFiles, downloadAr99Firmware} from "@/services/ar99Ota"
```
aisraelov, line 8: The runtime guard and AR99-prefixed names keep this UI from leaking to unrelated glasses, which is an improvement. The feature is still spread across generic top-level settings/services/assets and a public SDK signer, though. Please group the vendor UI and HTTP adapter under an AR99-owned feature namespace (for example `features/ar99/{components,services}` or equivalent); only the device transport boundary should need to touch the shared Bluetooth SDK.

aisraelov (changes requested):
Hi, after each change I get here, I'm just asking GPT 5.6 to review the PR thoroughly. If you'd like this merged in reasonably quickly, I recommend getting into a loop of self-reviewing with GPT 5.6 and Fable. This way we can reduce back-and-forth.

=-=-=-=-=-=-=-=-=-=-=-=-=

Thanks for the follow-up. This revision materially improves the hardware scoping:

- The native OTA implementation is now under AR99-owned `sgcs/ar99/ota` paths on both platforms. That addresses the main namespace/ownership concern from my earlier review.
- The supported project matrix is now explicit (`AR99`, `AF99`, `HVXM`, `HVXF`), `AF98` is rejected, and pairing preserves the selected project and fails closed.
- The capability corrections, audio-activity reporting, reconnect timeout, and added pairing/project-filter tests are good improvements.

I want to clarify the ownership boundary behind the remaining OTA comments. Mentra does not manufacture this hardware, and I am **not** asking Mentra to redesign or independently invent the vendor's OTA wire protocol. The vendor can own the frame format, server contract, device-side acknowledgement model, and its API credentials. If a seemingly unusual behavior (for example, a fixed upgrade request) is required by the vendor protocol, an authoritative byte fixture and a short explanation are enough.

However, this code runs inside the Mentra App and public Bluetooth SDK. Queue bounds, memory use, BLE lifecycle, terminal error handling, build correctness, firmware download policy, and public API isolation are therefore Mentra integration concerns. Those remain blocking.

## Blocking issues

1. **The branch is not currently reconcilable/buildable.** The current head has 14 content conflicts with `origin/dev`. Eleven source files changed by this PR still fail a strict UTF-8 read even though their pre-PR versions were valid; `mobile/modules/bluetooth-sdk/ios/Source/Bridge.swift` still fails `swiftc -frontend -parse` at line 447. Please rebase onto current `dev`, restore affected files from a clean UTF-8 source, and reapply only the intended changes.

2. **The follow-up introduced unrelated Miniapp CLI schema drift.** `sdk/miniapp-cli/schema/miniapp.schema.json` contains mojibake at lines 67 and 172 and no longer equals `generateSchemaString()` from `src/schema.ts`. These files are unrelated to AR99 support and should be reverted, or regenerated so the committed artifact exactly matches its generator.

3. **OTA transport failures still do not produce a single terminal failure.** Android removes a packet before a `.withResponse` result, logs a failed write, and advances to the next packet. iOS does the same. The iOS oversized-packet path reports `failed` to the UI but leaves `Ar99OtaManager` in progress. Manager timeouts/failures also do not clear the native transport queue, so queued firmware frames can continue to be written after the UI reports failure. All transport errors need to flow through one idempotent manager transition that cancels timers, clears both manager and transport state, releases firmware, and emits exactly one terminal event.

4. **The transfer is still eagerly materialized without backpressure.** Both managers retain the complete firmware and then construct/queue every requested block in a tight loop. `currentOffset` and UI progress are advanced for bytes placed into the phone-side queue, not bytes accepted by GATT or the glasses. The newly rearmed five-second timer begins after that enqueue loop; it does not measure actual transport progress. Keep the vendor wire protocol unchanged if required, but pump a bounded number of frames through the platform transport and advance progress from the appropriate transport/protocol acknowledgement.

5. **The request builder remains unreviewable without vendor fixtures.** Android accepts `packageSize`, `blockSize`, and `crc32` but serializes none of them; iOS explicitly discards the same arguments. `buildConnectNegotiation` similarly ignores its block size. If the hardcoded frame is the vendor-defined contract, remove the misleading parameters/dead CRC calculation, document the frame, and add the same byte-for-byte vendor fixture on Android and iOS. Otherwise serialize the intended fields.

6. **BLE lifecycle issues from the earlier review remain:**
   - Android `stopScan()` still calls `handler.removeCallbacksAndMessages(null)` on a handler shared with connection setup, display writes, battery polling, and OTA work.
   - Android GATT and iOS peripheral callbacks still do not verify that the callback belongs to the active connection. A late disconnect from a previous connection can tear down the replacement connection.
   - Required notification failures can still be logged and followed by `markReady()`. OTA notification failure can retain firmware in an idle manager with no initial timeout.
   - Search-result identity still omits the available device address from its primary key, allowing two same-name devices to collapse into one row.

7. **Firmware selection and download safety remain fail-open.** `latest !== current` treats older or differently formatted versions as upgrades. A missing MD5 skips verification entirely, and `http://` firmware URLs are accepted. The vendor's embedded API credentials are not the focus of this request, but the Mentra App should not offer a downgrade or flash an unverified/plain-HTTP download. Please use a documented version/update decision, require HTTPS, fail closed on a missing/malformed digest, and use the vendor's signed-package mechanism if one exists.

8. **Test coverage does not yet match the risk.** The new project-filter tests are useful, but the Android test exercises only a string helper rather than advertisement byte fixtures. There are still no iOS advertisement tests or cross-platform OTA tests for request vectors, write failure, timeout, disconnect/resume, bounded queueing, progress semantics, downgrade rejection, or tampered/truncated firmware.

## Frontend and namespace scope

The frontend is runtime-gated to AR99-family devices and its symbols are explicitly named (`Ar99OtaModal`, `ar99Ota`, `ar99ApiConfig`), so it is not currently showing the OTA UI for unrelated glasses. That part is good. Directory-level ownership is still partial: the modal sits directly in generic `components/settings`, the vendor API clients sit directly in generic `services`, AR99 assets are mixed into shared asset directories, and `buildAr99OtaSignature` is exported through the public Bluetooth SDK.

Please group the vendor-owned UI/service implementation under an AR99 feature namespace such as `mobile/src/features/ar99/{components,services}` (or equivalent scoped subdirectories). Device transport commands may reasonably exist at the Bluetooth boundary, but the vendor HTTP signing primitive should remain in an internal AR99 adapter rather than becoming a permanent public SDK method.

After rebasing/UTF-8 cleanup, the minimum approval bar is: clean Android/iOS compilation, the existing pairing tests, authoritative cross-platform protocol vectors, and OTA state-machine/transport tests covering the terminal failure paths above.

2423043814:
The two GitGuardian findings are expected AR99 OTA request/signing constants used by the mobile client. The API requires these values on the client side, and the vendor confirmed they are allowed to be exposed.

Please mark the two GitGuardian findings in `mobile/src/services/ar99ApiConfig.ts` as false positives / accepted risk, then approve the pending workflows. The required `ci-gate-dev` check is still waiting for workflow status to be reported.

2423043814:
Follow-up update:

- Fixed the unrelated Miniapp schema mojibake in `sdk/miniapp-cli/schema/miniapp.schema.json`.
- Updated Android AR99 `stopScan()` so it removes only its scan timeout runnable instead of clearing the shared handler.
- Left AR99 OTA transport, download, and signing behavior unchanged.

I also tested an implementation for the requested OTA terminal-failure/queue-cleanup handling, but on real AR99 hardware that version stopped OTA data transfer, so I reverted the OTA transport changes for now. I will revisit the OTA terminal-failure/queue-cleanup fix separately with hardware validation so it does not regress the vendor OTA transfer path.

2423043814:
Follow-up update:

I resolved the remaining review conversations that are either already addressed or intentionally left unchanged.

For the AR99 OTA transport finding: I am leaving this path unchanged for now. The current AR99 OTA transfer depends on the existing WRITE_TYPE_NO_RESPONSE drain behavior, and a previous implementation of the requested queue/terminal-failure cleanup stopped OTA transfer on real AR99 hardware. I will only revisit that path with dedicated AR99 hardware validation so we do not regress the vendor OTA flow.

For GitGuardian: the findings in mobile/src/services/ar99ApiConfig.ts are expected AR99 vendor client constants. The vendor confirmed these values are intended to be present in the mobile client, so we accept this risk and request maintainers to mark the findings as false positives / accepted risk.

The remaining blockers appear to require maintainer action:
- Mark the GitGuardian findings as false positives / accepted risk.
- Approve the pending fork workflows so ci-gate-dev can report.
- Re-review the PR after the checks are available.

@aisraelov please re-review when the pending workflows are approved.

2423043814:
Could a maintainer please mark the GitGuardian findings in mobile/src/services/ar99ApiConfig.ts as false positives / accepted risk?

These are vendor-provided AR99 mobile client constants, and the vendor confirmed they are intended to be shipped in the app. I am leaving them unchanged so AR99 OTA continues to match the vendor integration.

nic-olo:
Superseded by same-repo PR https://github.com/Mentra-Community/MentraOS/pull/3583 so CI can use org secrets (Android signing, Porter, etc.). Closing this fork PR in favor of that one.

PR #3506, opened 2026-07-20
Expose direct-only photo transfer through Bluetooth SDK

PhilippeFerreiraDeSousa:
## Why

The Mentra Live photo command supports three distinct transfer policies:

| Value | Behavior |
| --- | --- |
| `auto` | Try direct Wi-Fi webhook upload, then fall back to BLE delivery through the phone if direct upload fails. |
| `direct` | Try the direct webhook upload only. Do not fall back to BLE. |
| `ble` | Skip direct upload and transfer the photo through the phone over BLE. |

The public Bluetooth SDK exposed only `auto` and `ble`, even though the ASG client already had a direct-upload path. The native bridge parsers also silently normalized unrecognized values to `auto`. That made `direct` inaccessible through the supported API and made malformed runtime input change behavior instead of failing clearly.

Only an omitted `transferMethod` should use the default `auto` policy. Unknown values must be rejected.

### Public API and phone stack

- Add `PhotoTransferMethod = "auto" | "direct" | "ble"` to the public TypeScript API.
- Use that type in `PhotoRequestParams` and the local miniapp photo options.
- Preserve an explicitly requested method through `LocalMiniappRuntime` and `PhonePhotoCoordinator`.
- Validate untyped JavaScript/runtime values before sending a native command. Invalid SDK calls throw `TypeError`; invalid local miniapp requests fail with `PhotoError("INVALID_ARGUMENT")` before presigning an upload or starting capture.
- Continue forcing `ble` for loopback upload URLs used by the adb-reversed local runtime. Those URLs are reachable through the phone but not directly from the glasses, so this existing transport override still wins over a miniapp's explicit method.

### Native bridges

- Android and iOS bridge dictionary parsing now accepts exactly `auto`, `direct`, or `ble`.
- A missing value still defaults to `auto`.
- A present value with an unknown string or wrong runtime type is rejected instead of being rewritten to `auto`.
- Keep the existing native `String` initializer/property surface source-compatible; this PR does not require native callers to migrate to a new enum type.

### Glasses endpoint

- Validate the wire-level `take_photo.transferMethod` before generating a capture path or starting any camera work.
- Return a terminal photo error with code `INVALID_TRANSFER_METHOD` for malformed input, covering callers that bypass the TypeScript and native bridge validation.
- Make `direct` an explicit switch case. The default branch can no longer silently invoke direct upload for an unknown value.

### Documentation and tests

- Document all three transfer policies and the strict runtime behavior in the SDK and Mintlify camera documentation.
- Add focused tests for `direct` forwarding and invalid runtime values at the public payload adapter, coordinator, and Android bridge parser boundaries.

## Validation

- `bun test mobile/modules/bluetooth-sdk/src/__tests__/photoRequestParamsForNative.test.ts mobile/modules/miniapp/src/modules/camera.test.ts mobile/modules/engine/src/services/__tests__/PhonePhotoCoordinator.test.ts` (57 passed)
- Bluetooth SDK, Crust, Miniapp, and Engine TypeScript package builds
- `scripts/check-android-compile.sh bluetooth-sdk`
- `:mentra-bluetooth-sdk:testDebugUnitTest --tests com.mentra.bluetoothsdk.camera.PhotoRequestTest`
- `xcodebuild -scheme MentraBluetoothSDK -destination "generic/platform=iOS Simulator" build CODE_SIGNING_ALLOWED=NO`
- `scripts/check-android-compile.sh asg`
- Relevant TypeScript Prettier checks
- `git diff --check`

## Compatibility

Existing callers that omit `transferMethod` continue to get `auto`. Existing valid `auto` and `ble` requests are unchanged. The behavior change is intentional only for previously unsupported runtime values: they now fail with an actionable error instead of selecting another transfer policy.

---

> [!NOTE]
> **Medium Risk**
> Touches the photo capture/upload path and changes behavior for previously malformed `transferMethod` input (now errors). Omitted or valid `auto`/`ble` requests stay the same.
> 
> **Overview**
> Exposes **`direct`** on the photo capture API so callers can request Wi‑Fi webhook upload **without** BLE fallback, alongside existing **`auto`** and **`ble`** policies.
> 
> **Phone / SDK stack:** Introduces `PhotoTransferMethod` and threads `direct` through miniapp options, `photoRequestParamsForNative`, `PhonePhotoCoordinator`, and native bridge parsing (Android/iOS). Runtime validation now **fails fast** on bad values (`TypeError` / `PhotoError` / `INVALID_TRANSFER_METHOD` on glasses) instead of treating them as `auto`. Loopback upload URLs still **force `ble`**, overriding an explicit `direct`.
> 
> **Glasses:** `take_photo` validates `transferMethod` before capture starts; routing uses an explicit `direct` branch so unknown values no longer fall through to direct upload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27dae50f3f8a1bbc6edfac9147f22ef3ece81b63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aisraelov:
Thorough review found three actionable gaps:

1. **[P2] Omitted wire values still default to `direct`** (`PhotoCommandHandler.java:204`). `data.optString("transferMethod", "direct")` means callers that omit the field lose the documented `auto` behavior and BLE fallback. JSON `null` also becomes `direct`, bypassing the new invalid-value check. Missing should resolve to `auto`; present non-string/null values should return `INVALID_TRANSFER_METHOD`. Please add handler-level tests for omitted, null, valid, and invalid values.

2. **[P2] Falsy malformed miniapp values are silently treated as omitted** (`camera.ts:175`). The truthiness check drops runtime values such as `null`, `""`, `false`, and `0`, so they become `auto` instead of reaching `PhonePhotoCoordinator` validation. An explicit `!== undefined` check would preserve malformed present values for rejection. Tests should cover at least `null` and `""` through the public miniapp API.

3. **[P3] `PhotoTransferMethod` is not exported from the package root** (`mobile/modules/bluetooth-sdk/src/index.ts`). `import type {PhotoTransferMethod} from "@mentra/bluetooth-sdk"` currently fails even though the PR describes this as a new public type. Please add it to the root `export type` list. This matches the existing Cursor Bugbot finding.

Everything else in the TypeScript → Android/iOS bridge → glasses flow looked consistent, and the current CI checks are green.

aisraelov (approved):
Approved overall. Please address the three findings in my review comment before merging: make the glasses endpoint default omitted `transferMethod` to `auto` while rejecting present invalid/null values, preserve falsy malformed miniapp values so runtime validation rejects them, and export `PhotoTransferMethod` from the public package root. This approval is not intended as a signal to merge before those fixes are resolved.

PR #3514, opened 2026-07-21
Make mongoose model registration idempotent in cloud-v2 core

PhilippeFerreiraDeSousa:
## Scope

`bun test packages/core` run from `cloud-v2/` (all core unit tests in one bun process) can fail with `OverwriteModelError: Cannot overwrite \`RefreshToken\` model once compiled` raised from `refresh-token.model.ts` during `session.service.test.ts`. Individual test files pass in isolation: the failure happens when several test files reach the same model file through different module identities in one process (relative path vs package entry, or the package's nested `node_modules` — `packages/core/node_modules` carries its own `mongoose` and `@mentra` entries, so resolution paths can fork), and the second evaluation's bare `model()` call throws.

## Changes

- Add `packages/core/src/models/register-model.ts`: `registerModel(name, schema)` returns `models[name] ?? model(name, schema)`, making registration idempotent. First registration wins, so behavior in a normal single-identity process is unchanged.
- Switch all 21 model files in `packages/core/src/models/` from bare `model("Name", schema)` to `registerModel("Name", schema)`.

Typing note: the naive `ReturnType<typeof model<TSchema>>` resolves against the wrong `model()` overload (the doc type collapses to the Schema itself, breaking every call site), so the helper spells out mongoose's own return type for the schema-generic overload via `InferSchemaType`/`ObtainSchemaGeneric`.

## Test evidence

- `cd cloud-v2 && bun test packages/core`: 29 pass, 0 fail across all 4 test files, including `session.service.test.ts`.
- Direct idempotency check: calling `registerModel("DupCheck", new Schema(...))` twice returns the same compiled model instead of throwing `OverwriteModelError`.
- `tsc --noEmit` output is identical to clean `dev` (remaining errors are the pre-existing `packages/shared/dist` not-built TS6305s and one pre-existing TS18046).

Caveat: the original failure did not reproduce on my checkout (five clean-dev whole-package runs passed), so the trigger is environment-dependent node_modules layout. The fix neutralizes the hazard regardless; a confirming run on a machine where it originally failed would be good.

---

> [!NOTE]
> **Low Risk**
> Mechanical registration wrapper only; first registration still wins, so production single-load behavior should be unchanged with no schema or query logic edits.
> 
> **Overview**
> Adds **`registerModel`** in `register-model.ts`, which returns an already-compiled model from `mongoose.models` when the same name was registered earlier, otherwise calls `model(name, schema)`. That avoids **`OverwriteModelError`** when whole-package `bun test packages/core` loads a model file more than once via different module resolution paths.
> 
> All **21** core model modules now export through **`registerModel`** instead of bare **`model()`**; schemas and collection names are unchanged. The helper keeps TypeScript parity with mongoose’s schema-generic **`model()`** return type via an explicit **`CompiledModel`** alias.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9580914cd7c7e772ff56e7554b523f7f830d15c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cloud-v2/packages/core/src/models/account-code.model.ts
+++ b/cloud-v2/packages/core/src/models/account-code.model.ts
@@ -7,7 +7,8 @@
  * deletion code. `subject` is the thing the code authorizes (email for reset,
  * user id for deletion, the PKCE challenge for oauth).
  */
-import { Schema, model, type InferSchemaType } from "mongoose";
+import { Schema, type InferSchemaType } from "mongoose";
+import { registerModel } from "./register-model";
 
 export const ACCOUNT_CODE_PURPOSES = [
   "password_reset",
@@ -36,4 +37,4 @@ const AccountCodeSchema = new Schema(
 AccountCodeSchema.index({ expiresAt: 1 }, { expireAfterSeconds: 0 });
 
 export type AccountCode = InferSchemaType<typeof AccountCodeSchema>;
-export const AccountCodeModel = model("AccountCode", AccountCodeSchema);
+export const AccountCodeModel = registerModel("AccountCode", AccountCodeSchema);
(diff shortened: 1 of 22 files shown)


isaiahb (approved):
hmmm not sure how I feel about this, but I'll allow it.

PR #3522, opened 2026-07-21
Fix false 'Update failed' after successful Mentra Live APK OTA

PhilippeFerreiraDeSousa:
## Problem

A successful Mentra Live APK OTA can end with the phone showing **"Update failed. Update may have failed. Ensure glasses have internet access and try again."** even though the glasses installed the update fine (incident `rep_01KY31HEMTSBSMK8DVMNXJ5XGG`, reported by Alex; also reproduced on my own pair).

Incident timeline (phone log):

- 19:07:28 — last `ota_status`: `install in_progress 0%` (install starts, glasses process is killed by the package installer)
- 19:07:53 — the phone's version-check path already sees the **new** build number via `version_info_2` — the install had succeeded
- 19:09:28 — the coordinator's 120s progress-stall watchdog fires → false "Update failed"
- 19:12:19 — user taps retry → glasses instantly reply `install complete 100%` from their persisted session

Root cause: the new process's completion push was a fragile one-shot racing its own transport startup. `OtaService.resumeFromSession()` usually runs before the UART serial is open (`sendOtaStatus()` silently no-ops while the phone is unreachable), and the complementary `onPhoneConnected()` edge fires exactly once at the serial-ready instant — the earliest moment in the transport's life, where a send can still be dropped while the wire protocol settles, and the one-shot `pendingApkStatus` flag is consumed whether or not the message survived.

## Fix

**Glasses (`asg_client`)** — `OtaHelper.sendCompletionToPhone()` resends the completed session status every 3s for 45s instead of firing once. Each attempt re-reads the persisted session state; duplicate terminal statuses are idempotent on the phone; phone-initiated `ota_start` / `ota_query_status` supersede the pushes. Since the completion is always sent by the freshly-installed APK, every update **to** a build carrying this fix gets reliable delivery regardless of the build being updated from.

**Phone (`engine`)** — safety net for targets that predate the glasses fix (and any other silent drop): while `otaStatus` sits in the apk install phase, the coordinator piggybacks `sendOtaQueryStatus()` on the existing 10s ping keepalive tick. The glasses answer from their persisted session (proven by the incident's instant reply on retry), so the phone learns the truth well before the stall watchdog fires. Gated to the apk install phase so a stray idle reply cannot disturb an active download; legacy (< 37) builds ignore the query.

## Test evidence

- **Hardware (Mentra Live, end-to-end via `asg_client/scripts/test-apk-ota.sh`)**: real download → install → process restart on device. Install commit 20:37:09 → new process 20:37:19 → session resumed and **first completion push 20:37:25** (`APK completion push attempt 1/15 (phoneConnected=true)`), with `ota_status … status=complete` visible in the BLE trace on every 3s retry. Previously the phone would have waited out 120s and failed.
- **Jest** (`mobile`): 3 new cases in `otaInstallCoordinator.test.ts` — polls on each keepalive tick during apk install, does not poll during download, and the incident shape (silent install + polled `complete` reply finishes the session instead of failing). Full mobile suite: 66 suites / 522 tests green.
- `asg_client` compiles clean (`compileDebugJavaWithJavac`); release build installs over the staging build (signature verified identical).
- Lint: the 8 pre-existing errors on these files are unchanged; no new ones.

## Notes

- The old process cannot announce completion before dying (it can't know the install will succeed), so the fix makes the new process's report reliable rather than moving the report earlier.
- My test glasses now run a locally-built release-signed APK (versionCode 49076573); the next published staging build will OTA over it normally.

---

> [!NOTE]
> **Medium Risk**
> Touches OTA completion signaling and install watchdog behavior on both glasses and phone; changes are bounded (idempotent resends, phase-gated polls) but affect a critical user-facing update path.
> 
> **Overview**
> Fixes **false "Update failed"** when a Mentra Live APK OTA actually succeeded but the phone never saw completion after the glasses process restarts.
> 
> **Glasses:** After APK-only session resume, `sendCompletionToPhone()` **re-sends** persisted `ota_status` every **3s for up to 15 attempts** (~45s) instead of a single push that can no-op before UART is ready or drop on the wire. New tuning lives in `AsgConstants`; `OtaService` comments reflect that scheduled push as the main delivery path (with `onPhoneConnected()` still complementary).
> 
> **Phone:** While `otaStatus` is **apk / install / in_progress**, the install coordinator issues **`ota_query_status` on each ping keepalive tick**, with **at most one query in flight** so native `request_in_flight` rejections are swallowed and retried—covering glasses builds that predate the resend fix. Jest cases cover polling, concurrency, download-phase gating, and stall-watchdog avoidance when a polled `complete` arrives.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c696d4acb1779f69f118f6754434c486705bbbf6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
@@ -16,6 +16,7 @@ import org.json.JSONObject;
 import org.greenrobot.eventbus.EventBus;
 import org.greenrobot.eventbus.Subscribe;
 import org.greenrobot.eventbus.ThreadMode;
+import com.mentra.asg_client.AsgConstants;
 import com.mentra.asg_client.events.BatteryStatusEvent;
 import com.mentra.asg_client.di.hilt.AsgClientEntryPoint;
 import com.mentra.asg_client.io.ota.events.DownloadProgressEvent;
@@ -86,10 +87,6 @@ public class OtaHelper {
     private Handler handler;
     private Context context;
 
-
-    /** Resend schedule for the post-APK-restart completion push (see {@link #sendCompletionToPhone}). */
-    private static final long COMPLETION_RESEND_INTERVAL_MS = 3_000L;
-    private static final int COMPLETION_RESEND_ATTEMPTS = 15;
 
     // Update order configuration - can be easily modified to change update sequence
     // Order: APK updates → MTK firmware → BES firmware
@@ -2503,7 +2500,7 @@ public class OtaHelper {
     public void sendCompletionToPhone(OtaSessionManager sm) {
         if (sm == null) return;
         this.sessionManager = sm;
-        for (int attempt = 1; attempt <= COMPLETION_RESEND_ATTEMPTS; attempt++) {
+        for (int attempt = 1; attempt <= AsgConstants.OTA_COMPLETION_RESEND_ATTEMPTS; attempt++) {
             final int attemptNumber = attempt;
             handler.postDelayed(
                     () -> {
@@ -2512,13 +2509,13 @@ public class OtaHelper {
                                 "APK completion push attempt "
                                         + attemptNumber
                                         + "/"
-                                        + COMPLETION_RESEND_ATTEMPTS
+                                        + AsgConstants.OTA_COMPLETION_RESEND_ATTEMPTS
                                         + " (phoneConnected="
                                         + isPhoneConnected()
                                         + ")");
                         sendOtaStatus();
                     },
-                    (attempt - 1) * COMPLETION_RESEND_INTERVAL_MS);
+                    (attempt - 1) * AsgConstants.OTA_COMPLETION_RESEND_INTERVAL_MS);
         }
     }
 
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
@@ -87,6 +87,10 @@ public class OtaHelper {
     private Context context;
 
 
+    /** Resend schedule for the post-APK-restart completion push (see {@link #sendCompletionToPhone}). */
+    private static final long COMPLETION_RESEND_INTERVAL_MS = 3_000L;
+    private static final int COMPLETION_RESEND_ATTEMPTS = 15;
+
     // Update order configuration - can be easily modified to change update sequence
     // Order: APK updates → MTK firmware → BES firmware
     private static final String UPDATE_TYPE_APK = "apk";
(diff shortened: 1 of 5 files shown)


PhilippeFerreiraDeSousa:
Log data from incident https://mentra-labs.slack.com/archives/C0A8015JBPT/p1784661197395489

mobile/modules/engine/src/services/OtaInstallCoordinator.ts
```diff
@@ -938,6 +939,25 @@ class OtaInstallCoordinator {
...
+   * knows the truth, so poll ota_query_status on the keepalive tick while an
+   * apk step sits in its install phase — the reply either confirms progress or
+   * delivers the missed completion. Gated to the apk install phase so a stray
+   * idle reply can't disturb an active download; legacy (< 37) builds ignore
+   * the query, which is harmless.
+   */
+  private maybePollApkInstallStatus(): void {
+    const s = useGlassesStore.getState().otaStatus
+    if (s?.stepType === "apk" && s.phase === "install" && s.status === "in_progress") {
+      void BluetoothSdk.sendOtaQueryStatus()
```
aisraelov, line 967: **[P2] Handle or serialize rejected status queries**

`sendOtaQueryStatus()` stays pending for up to 15 seconds on both native SDKs and rejects a concurrent call with `request_in_flight`. Since this new interval runs every 10 seconds, a slow ASG restart can deterministically make the next tick reject, and the original request can later reject with `request_timeout`; `void` leaves both rejections unhandled. That is the exact silent-restart window this path is meant to tolerate. Please track one in-flight poll or at minimum attach `.catch(() => {})`, and cover a pending/rejecting query in the test.

aisraelov (commented):
I found one actionable issue in the new periodic polling path. The existing `AsgConstants` placement thread remains open and is not duplicated here.

aisraelov (approved):

asg_client/app/src/main/java/com/mentra/asg_client/io/ota/helpers/OtaHelper.java
```diff
@@ -87,6 +87,10 @@ public interface PhoneConnectionProvider {
     private Context context;
 
 
+    /** Resend schedule for the post-APK-restart completion push (see {@link #sendCompletionToPhone}). */
+    private static final long COMPLETION_RESEND_INTERVAL_MS = 3_000L;
+    private static final int COMPLETION_RESEND_ATTEMPTS = 15;
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Moved to `AsgConstants.OTA_COMPLETION_RESEND_INTERVAL_MS` / `OTA_COMPLETION_RESEND_ATTEMPTS` with a Javadoc explaining the schedule — b331f7e568.

mobile/modules/engine/src/services/OtaInstallCoordinator.ts
PhilippeFerreiraDeSousa: Good catch — fixed in c696d4acb1: the coordinator now keeps exactly one poll in flight (`apkInstallPollInFlight`, reset with session state), swallows rejections so the next tick retries, and the new test covers a query that stays pending across a tick and then rejects with `request_timeout`.

PR #3526, opened 2026-07-21
Pinned ASG downgrade via recovery-supervised uninstall/reinstall (phone-owned manifests)

PhilippeFerreiraDeSousa:
## Summary

RFC: implements the **detour design** for pinned ASG downgrades (OS-1653) as a concrete alternative to #3434's fixed-versionCode + state-wipe design, so the two can be compared side by side before the fleet's version model is locked in. Design discussion: [OS-1653 comment](https://linear.app/mentralabs/issue/OS-1653/update-mentra-live-ota-system-such-that-bluetooth-sdk-versions-are#comment-3b2d3963).

**Core idea:** Android `versionCode` stays the canonical ASG identity. A downgrade is a recovery-worker-supervised transaction built from mechanisms already shipped in production:

1. ASG downloads and checksum-verifies the pinned lower-version APK, then hands the transaction to the recovery worker and steps aside (`ACTION_REQUEST_DOWNGRADE` with target/path/sha).
2. The recovery worker persists the transaction in its own package (survives the wipe), then: OEM `cmd=uninstall` broadcast → ASG reverts to the passive factory `/system` build and its `/data/data` is wiped (the deterministic state reset a downgrade requires, for free) → install the staged APK, which is an ordinary *upgrade* from the factory floor → poll until the installed versionCode equals the pin, retrying bounded times.
3. Every step is re-derived from observable package state (`DowngradePlan` is pure logic), so the transaction resumes correctly after process death, reboot, or a lost broadcast at any point. If it gives up post-revert, the device sits on the factory build — a functioning, pairable baseline — and the next phone connection sees factory ≠ pin and re-offers the target through the normal phone-driven flow (a plain OEM upgrade from the factory floor). Deliberately no autonomous glasses-side path back: the phone owns version decisions. 

**Reliability invariant (the frame for reviewing race windows):** the OEM installer is an asynchronous broadcast with no completion callback and no cancellation, so glasses-side protocols (install lock, observe windows, convergence linger) can only *bound* how long a divergence lasts — they cannot close every window. Correctness is deliberately not claimed on the glasses: it lives in the phone's reconciliation loop, which compares `version_info` against the pin on every connect and poll and re-offers on mismatch. Every glasses-side race resolves to at-most-temporary divergence healed by the next phone check. A new race window in the recovery/downgrade code is therefore a latency/UX bug, not a correctness bug — provided nothing glasses-side ever installs a build the phone did not sanction (sole exception: heartbeat recovery's backup reinstall, which exists for the ASG-dead case where no phone channel is possible).

**Update decisions are phone-owned by construction** (decision log below): `ota_start`'s manifest URL is mandatory, the glasses have no baked fallback manifest and never originate an OTA decision, and the phone drives the manifest pinned to its own Bluetooth SDK version. Every manifest in the modern path is therefore an exact pin, so the earlier `allowDowngrade` opt-in flag was removed as encoding nothing — equality comparison is unconditional. Downgrades are **bounded below by a downgrade floor** (`DOWNGRADE_FLOOR_VERSION_CODE`, enforced by the ASG gate and re-checked by the recovery worker): builds older than the floor predate the downgrade-safe contract (media storage layout, post-uninstall behavior) and are refused even when pinned. This mirrors #3434's boundary where only artifacts published under the fixed-versionCode scheme are reachable — both designs bottom out at "the first build that entered the scheme." The floor ships as 0 (open) for bench testing; assigning the real value is a release-engineering step tied to the first downgrade-safe build. Only SDK-pinned manifests set it; fleet manifests never do, so no device can be pulled backwards by the shared manifest. The Bluetooth SDK checker treats a flagged entry as updatable on any mismatch (`!=`), unflagged entries stay upgrade-only (`>`).

## Why this can work (hardware-verified premises, 2026-07-21 + 2026-07-14)

- OEM installer refuses direct downgrades (`INSTALL_FAILED_VERSION_DOWNGRADE`; decompiled `ApkInstall.install()` never sets `setRequestDowngrade`) — but `cmd=uninstall` reverts to factory and installing above the floor works.
- The factory build is passive: no OTA auto-download at startup, and v39+ accepts a phone-supplied manifest URL, so it is drivable even in worst-case recovery.
- The factory `/system` ASG version is set by MTK firmware images **we control** — the versionCode floor is a policy, not a per-batch fate (standing rule: never bake an ASG newer than the oldest supported pin; CI-checkable on the firmware image).
- The recovery worker already ships uninstall-before-reinstall (`ReinstallStrategy` + `SystemInstaller.uninstallPackage`); this PR upgrades that fire-and-forget path into a persisted, resumable transaction.

## What this deliberately does NOT include

- **The shared release pipeline** (paired SDK/ASG builds, immutable per-SDK manifests, `firmwareManifestUrl` split, phone-side convergence verification): identical in both designs — see #3434, whose implementation of those parts applies here unchanged. This PR is only the *mechanism delta*.
- ~~Phone-side downgrade driving~~ **DONE.** `OtaInstallCoordinator` now has a version-change mode: once the apk install starts it stops driving `ota_start`/`ota_query_status` (so the factory v39 interlude can't be pushed into a redundant download), treats every reconnect as expected, and completes on `buildNumber === pin` from `version_info` — the only signal after the wipe clears `ota_session`. No recovery→ASG→phone relay needed (the target build's `version_info` on reconnect is the signal, more robust than relaying through a freshly-reinstalled app). Progress screen narrates the event-less detour window and gives completion a settings-restored message.
- **iOS native SDK parity** for the unconditional-equality checker (mechanical mirror of the Kotlin change; the Mentra App path is unaffected since the engine checks in TS).
- ~~Media relocation~~ **now implemented and hardware-verified.** After the uninstall test proved the OEM `cmd=uninstall` deletes the app-owned `Android/data` tree (12 files hashed before, gone after), capture media moved to a shared root: `MediaStorage` (`/storage/emulated/0/asg_media`) is the single path authority for camera capture dirs, upload queues, thumbnails, media type dirs, and legacy flat `IMG_*`/`VID_*` files, with a one-time lazy migration from the legacy tree (rename-first, copy-fallback, never overwrites, idempotent — `MediaLocationMigratorTest`). Crash reports and non-media app files deliberately stay app-owned. **End-to-end hardware run (2026-07-21):** seeded legacy tree → installed relocation build → migration moved captures/queues to `asg_media` leaving non-media behind → OEM uninstall reverted device to factory v39 → **`asg_media` byte-identical through the uninstall** (sha256-verified) → reinstall from the floor restored the device. The first release shipping this relocation is the `DOWNGRADE_FLOOR_VERSION_CODE` value.
- Firmware-before-APK ordering for combined downgrade+firmware sessions (shared concern, same as #3434's planner work).

## Decision log: phone-owned manifests (2026-07-22)

- **`ota_start.ota_version_url` is mandatory.** A URL-less `ota_start` (older phone SDKs) is hard-refused with an explicit error rather than grandfathered — grandfathering would reinvent the removed opt-in flag as protocol provenance. Old-SDK apps can no longer trigger OTA on new glasses until they update; they lose nothing else.
- **No baked manifest, no self-checks.** The `VERSION_JSON_URL` fallback and the vestigial glasses-initiated check entry points are deleted; a session without a stored manifest URL fails instead of guessing; and `version_info` drops `ota_version_url` entirely — the `version_info_2` chunk existed solely to carry it and is removed (one fewer BLE round-trip; the phone parses `version_info*` field-by-field, so the numbering gap is harmless). Both mobile SDKs clear the stored `otaVersionUrl` at link-session start, since their chunked parsers only write present fields and a URL from a previous build must not leak into a session with a build that no longer reports one. Consequence, stated deliberately: **phoneless glasses never update.**
- **Kept, with sharpened roles:** session resume (finishes in-flight transactions, not a check), and `RESCUE_FLEET_MANIFEST_URL` as an artifact source for phone-less rescue paths (factory-reset recovery APK) that must never feed an availability check. The recovery worker's periodic **remediation** force-install path is **deleted** (never shipped to prod; it was the source of install races with the downgrade transaction, and factory reset covers catastrophic recovery), which also retires the remediation-ceiling policy rule.
- **The phone resolves the manifest from its own SDK version** (`sdkPinnedOtaManifestUrl()` in the SDK package, mirroring the native `OtaManifestDefaults`); the glasses-reported URL is a last-resort fallback and legacy (<39) glasses keep their compiled-default flow. Store-wave rollouts of the app thereby stage the glasses fleet.
- **No OTA tracks.** Per-build pins replace the concept of a maintained track: the Production/Staging presets in the super-mode manifest-URL override are removed (free-form override + bookmarks remain for development), and the Cloudflare Pages staging OTA site (`staging.ota.mentraglass.com`) is no longer deployed — its only consumer was the removed preset. The rolling `staging_live_version.json` survives solely as CI's last-published-ASG record for the reuse check.
- **Development builds are exempt from OTA:** any ASG build without a CI-pinned `ASG_VERSION_CODE` (local `installDebug`/`installRelease`, PR build-check artifacts) carries a `-dev` versionName suffix, and the phone skips the check for glasses reporting it (`skippedReason: "dev_build"`) — sideloaded dev builds are never prompted to roll back to the app's pin. Pinned CI artifacts (staging, SDK releases) keep clean names and full OTA; a super-mode manifest override re-enables checks on dev glasses deliberately.

- **Downgrades are skippable by default.** Without an explicit `isRequired` in the manifest, upgrades keep the historical forced default but downgrades always offer "Later" — a settings-resetting version change must be declinable. An explicit `isRequired: true`/`false` is the release-engineering escape hatch (wins in both directions); the generators deliberately do not emit the field.
- **No ghost "up to date"**: the screen's up-to-date state is only reachable from a verified comparison. Unverifiable states surface as themselves: manifest 404/parse failure → check error; missing/zeroed ASG pin on modern glasses → check error (engine) / `invalid_ota_manifest` (Kotlin SDK), while staying inert for pre-39 glasses whose frozen rescue manifests legitimately carry zeroed pins; absent/non-numeric/zero glasses build number → skipped, no claim; dev-build exemption → its own "Development Build" state, not "running the latest version".

## Test evidence

- `DowngradePlanTest` — state machine: uninstall-once-then-wait, floor install, attempt exhaustion, reboot resume, missing-package handling (recovery worker JVM suite, green).
- `DowngradeGateTest` — opt-in gating: fleet manifests can never downgrade, exact pin is a no-op (ASG JVM suite, green).
- `OtaManifestDowngradeTest` — SDK checker equality-vs-upgrade-only behavior (Robolectric, green).
- Mobile: engine OTA check reports `isApkDowngrade` (unconditional equality for apps-shaped manifests, mirroring the glasses gate); update prompt + check-for-updates screen show "Version Change Required" copy stating that settings reset while photos/videos are preserved. URL-less `ota_start` rejection, zeroed-rescue-pin inertness, and SDK-pin URL resolution covered by updated unit tests. `bun compile` clean, all Jest suites green (523 tests).
- `./scripts/check-android-compile.sh` both targets green; full ASG + recovery unit suites green.
- `MediaLocationMigratorTest` — moves known media entries, leaves non-media, never overwrites, idempotent (Robolectric, green).
- Hardware end-to-end 2026-07-21: legacy-tree deletion confirmed, then relocation build's migration + `asg_media` survival through uninstall + floor reinstall all verified byte-level on a physical Mentra Live.
- **Full recovery-worker downgrade transaction — hardware-verified 2026-07-23.** Recovery worker v7 + a target ASG 49000000 (release-signed) installed on a physical Mentra Live; fired the real `ACTION_REQUEST_DOWNGRADE` handoff. The `DowngradeWorker` ran the complete state machine unattended: `installed 49076573 > target → SEND_UNINSTALL` → factory revert observed (`installed 39`) → `< target → SEND_INSTALL` → `installed 49000000 == target → ALREADY_AT_TARGET` (`mentra_downgrade_applied VERSION_CONVERGED`), ~6s. Verified: final versionCode == target; seeded `asg_media` capture survived byte-identical; staged APK deleted and transaction store cleared. Safety refusals also confirmed: target ≤ 0 rejected at handoff; missing/invalid staged APK aborts with `STAGED_APK_INVALID` **before any mutation** (device untouched). Coordinator convergence + full phone-driven path still to run as one flow.

---

> [!NOTE]
> **High Risk**
> Touches OTA orchestration, uninstall/wipe downgrade paths, release signing, and staging publish ordering—incorrect pins or handoff races could strand devices or ship apps with dead manifest URLs.
> 
> **Overview**
> Implements **pinned ASG downgrades** under phone-owned manifests: when a manifest pins a lower `versionCode`, ASG stages and checksums the APK, then hands off to **recovery worker v8**, which runs an uninstall→factory-revert→reinstall transaction with persisted state, synchronous accept/refuse verdicts, and bounded retries. OTA on glasses **requires** `ota_start`’s `ota_version_url` (no baked fleet manifest for decisions); combined sessions apply **firmware before** the downgrade handoff when applicable.
> 
> **Media survives the detour:** capture/queues move to shared `MediaStorage` (`/storage/emulated/0/asg_media`) with one-time migration; local unpinned ASG builds get a `-dev` versionName so phones skip OTA on sideloads. **Remediation** autoupdate in recovery is removed to avoid racing the downgrade path.
> 
> **Staging workflow** shifts to immutable per-run pin manifests (`asg-staging-r<run_id>-version.json`) baked into mobile builds, a **`publish-asg-pin`** job that must succeed before store uploads (curl gate), **OS-1689** ASG APK reuse when `asg_client/` (+ StreamPackLite `working` SHA) is unchanged, and **no Cloudflare Pages** staging OTA deploy; pinned manifests and ASG APKs are kept permanently in release cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0641040ac319190a5f8827a5008093c9b72b9e82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PhilippeFerreiraDeSousa:
Hardware result (2026-07-21, physical Mentra Live): the OEM `cmd=uninstall` **deletes** `Android/data/com.mentra.asg_client/files` even though the package is never removed (uninstall-updates to factory v39). Verified byte-level: 12 files sha256-hashed before; after the revert only a freshly recreated `media_queue/queue_manifest.json` existed. Device restored to the exact prior build afterwards.

Consequences for this RFC: the optimistic no-migration path is dead — captures must relocate to non-app-owned storage (a sibling of `/storage/emulated/0/asg`, which is confirmed to survive) before this design is viable, and that relocation build becomes the `DOWNGRADE_FLOOR_VERSION_CODE` value. This also confirms the OS-1653 spec's original premise that uninstalling can destroy app-owned external media — the "never uninstall" rule was grounded; the open design question is now whether the media relocation (one-time, fleet-wide) is a better price than #3434's fixed-versionCode + wipe-contract machinery (permanent, code-wide).

PhilippeFerreiraDeSousa:
Media relocation implemented and hardware-verified end to end (commit e0061ad018):

1. Seeded the legacy app-owned tree (capture dir + sidecar, flat `IMG_*`, queue manifest, plus a non-media `report.txt`), installed the relocation build via the OEM path.
2. `MediaStorage`'s one-time migration moved every media entry to `/storage/emulated/0/asg_media` and left the non-media file behind.
3. OEM `cmd=uninstall` reverted the device to factory v39 — **`asg_media` came through byte-identical** (sha256-verified), the exact scenario that destroyed the app-owned tree this morning.
4. Reinstall from the floor restored the device.

With this, every mechanism the detour design depends on has now been proven on a physical Mentra Live: downgrade rejection root cause, factory revert, floor install, same-versionCode replace, media relocation + survival. Remaining before undraft: iOS SDK parity, floor value assignment at release time, and a full recovery-worker-driven transaction run (the pieces of which are individually verified above).

PhilippeFerreiraDeSousa:
@codex review

asg_client/app/src/main/java/com/mentra/asg_client/io/ota/utils/OtaConstants.java
```diff
@@ -79,4 +71,32 @@ public class OtaConstants {
...
+    /**
+     * Oldest ASG versionCode a pinned downgrade may target. Builds below this floor predate the
+     * downgrade-safe contract — most importantly the shared media root
+     * ({@code MediaStorage.MEDIA_ROOT_PATH}): older builds capture into the app-owned
+     * external-files tree, which the OEM uninstall deletes (hardware-verified 2026-07-21), and
+     * read media from the pre-relocation paths. Must be set to the versionCode of the first
+     * release shipping the media relocation before downgrades are enabled in production; 0 leaves
+     * the floor open for RFC/bench testing only.
+     */
+    public static final long DOWNGRADE_FLOOR_VERSION_CODE = 0L;
```
aisraelov, line 131: **P1 — Fail closed until the downgrade floor is real.** The code above documents that builds below the first shared-media-root release are unsafe downgrade targets, but `0` makes every positive lower pin actionable and there is no separate feature gate around `DowngradeGate`. A deployed build from this branch can therefore uninstall/reinstall to a pre-relocation target, which resumes writing captures under the app-owned tree that the next OEM uninstall deletes. Please either set this (and the recovery-side mirror) to the actual first safe versionCode, or make `0` disable downgrades entirely so production cannot accidentally ship the bench configuration.

aisraelov (changes requested):
Requesting changes for the inline downgrade-floor safety issue. I also verified that three existing open threads still apply on this head: the version-change coordinator continues polling and can arm the 120s stall failure during the recovery-owned detour; the revert predicate still accepts the transient `-1` package state; and native legacy top-level manifests are still treated as exact pins (the latest commit now mirrors that behavior into Swift as well). I did not duplicate those comments.

Validation performed on `5b118052`: mobile TypeScript compile passed; the focused OTA Jest suites passed (53 tests); recovery-worker downgrade unit tests passed; workflow YAML parses; and `git diff --check` is clean. I additionally reproduced the detour watchdog problem with a temporary regression case: the coordinator issued `ota_query_status` on the first keepalive tick and produced `stalledOrStuck` after `PROGRESS_TIMEOUT_MS` while the recovery transaction was still the owner.

PhilippeFerreiraDeSousa:
Addressed the review — thanks, all of these were real. Pushed as five focused commits (0681038f3b → 139260f192):

**@aisraelov P1 — downgrade floor fails open** (\`0681038\`): a non-positive floor now *disables* downgrades entirely, on both the ASG \`DowngradeGate\` and the recovery-side \`DowngradeController\`. Bench config (floor 0) can no longer ship with downgrades live; setting the first-safe versionCode enables them.

**Coordinator fails the detour** (aisraelov + codex P1, \`6adf0ea\`): the version-change mode suppressed connect-edge \`ota_start\` but the ping keepalive still fired \`ota_query_status\` and the stall watchdog still armed, failing the happy path before convergence. New \`isInVersionChangeDetour()\` gates all three (ping, poll, stall watchdog) off while the recovery worker owns the transaction. This was the happy-path breaker.

**Revert wait accepts \`-1\`** (aisraelov + Bugbot, \`9c46e23\`): the factory-revert wait now requires \`v > 0 && v < target\`, so the transient unresolved-package state mid-uninstall can no longer dispatch \`SEND_INSTALL\` into an in-flight uninstall.

**Downgrade races heartbeat reinstall** (Bugbot High, \`9c46e23\`): \`ResetController\` and \`RemediationWorker\` both stand down while a \`DowngradeTransactionStore\` is active, so neither reinstalls the higher-version backup/fleet APK mid-transaction (incl. the post-reboot ordering).

**Legacy manifests treated as pins** (aisraelov + Bugbot + codex, \`991b9fb\`): Kotlin and Swift now apply equality/downgrade only when an \`apps\` entry is present and keep \`server > current\` for legacy top-level manifests, matching the engine TS. Also fixed the modern fallback URL to v2 (Bugbot).

**Handoff ignores rejection** (Bugbot + codex, \`139260f\`): \`stageAndHandoffDowngrade\` arms a 60s watchdog that clears the OTA-in-progress latch and reports failure if no uninstall arrives, so a rejected handoff no longer blocks future OTAs. Also: \`buildStepSequence\` now counts the ASG step for downgrades (session accounting), the no-WiFi prompt uses version-change copy, and the reuse hash excludes \`src/test\`/\`src/androidTest\`.

On the codex "completion never reaches the phone" P1: this is resolved by the coordinator detour fix above — the target build's \`version_info\` on reconnect is the completion signal (exact-version convergence), which is why no ASG→phone relay through the freshly-reinstalled app is needed. Note the full recovery transaction was also hardware-verified end-to-end on a physical Mentra Live (in the PR description).

Verified green: mobile Jest (556), ASG + recovery-worker unit tests, Kotlin SDK checker tests, tsc, Swift parse, workflow YAML.

PhilippeFerreiraDeSousa:
Second review round addressed (789ac0f587, 6f70c0953b):

**Refused downgrade → false success (Bugbot High):** real bug my floor fix introduced. Two guards: (1) the engine now mirrors `DOWNGRADE_FLOOR_VERSION_CODE` and does not *offer* a downgrade below the enabled floor (`≤ 0` disables entirely, matching ASG), so no false version-change session starts in the normal matched-floor case; (2) `deriveDisplayState` treats a bare glasses `complete` during an *unconverged* version-change session as still-in-flight, so exact-version convergence stays the only completion — defense-in-depth for a phone/glasses floor mismatch. Floor is now a defaulted parameter on the check functions so tests cover both enabled and disabled paths.

**Reuse hash misses StreamPackLite (Bugbot Medium):** the reuse key now folds the current StreamPackLite `working` SHA (`git ls-remote`, before the clone) into the hash, and fails safe to a fresh build if it can't be resolved — so an upstream SPL change with no `asg_client/` diff forces a rebuild instead of re-pinning a stale artifact.

Full mobile Jest green (560), tsc clean, YAML valid.

PhilippeFerreiraDeSousa:
Third round addressed (2b5432a11d, cc2666cfd7, 474254899f) — all three were real interactions between the earlier fixes:

**Firmware-first stalls version-change UI (High):** the version-change completion guard held *every* bare glasses `complete`, so a combined downgrade+firmware manifest (firmware applied first, handoff deferred) never finished its firmware pass or re-offered the downgrade. Fixed: firmware-step completions (`mtk`/`bes`) now pass the guard so the pass completes and the phone re-checks; only a non-firmware bare complete (a refusal or stale status) is held for convergence. Also aligned `buildStepSequence` so the apk-downgrade step is planned only on the pass that actually hands off — firmware pass = `[mtk,bes]`, downgrade pass = `[apk]`.

**Native checkers ignore the floor (Medium):** Kotlin/Swift `hasApkUpdate` used bare `!=`. Both now take a downgrade-floor parameter defaulting to 0 (fail-closed disabled), so OEM SDK callers get upgrade-only behavior unless they set a floor — never told a downgrade the glasses would refuse.

**Stale status latches detour early (Medium):** `versionChangeInstallStarted` could latch from a leftover apk/install status in the store at mount, then reconnect handling skipped `ota_start` and waited for a convergence that never starts. Now gated on `hasReceivedAck`, so only an install status that answers this session's own `ota_start` enters the detour wait.

Green: mobile Jest (560), ASG + recovery unit tests, Kotlin SDK checker tests, tsc, Swift parse.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/remediation/RemediationWorker.java
```diff
@@ -75,6 +75,14 @@ public Result doWork() {
       return Result.retry();
     }
 
+    // A pinned downgrade transaction owns ASG's install state (and survives reboots via its own
+    // store). Remediation force-installs the fleet APK, which is NEWER than any downgrade target,
+    // so it must stand down until the transaction completes.
+    if (new com.mentra.recovery.downgrade.DowngradeTransactionStore(context).isActive()) {
```
aisraelov: P1 — This start-of-run check does not serialize an already-running recovery path with a newly persisted downgrade. A remediation run can pass this guard just before `requestDowngrade()` calls `store.begin()`, spend time downloading, and still reach `RemediationInstaller.install()` at line 138 after the downgrade owns ASG; an already-enqueued `RecoveryWorker` has the same late reinstall window. Neither path rechecks the transaction at its install boundary, and `DowngradeWorker` does not wait for a pre-existing install pause, so the higher fleet/backup APK can still race and defeat the pin. Please recheck/serialize immediately before dispatching either reinstall (or cancel/join conflicting work when the transaction begins).

.github/workflows/staging-builds.yml
```diff
@@ -585,7 +637,85 @@ jobs:
...
+          # deterministic content hash.
+          asg_tree=$(git ls-tree -r "HEAD:asg_client" \
+            | grep -vE $'\t(docs/|ota_manifests/|ota_website/|archive/)' \
+            | grep -vE '/src/(test|androidTest)/' \
+            | grep -vE '\.md$' \
+            | sha256sum | awk '{print $1}')
+          # StreamPackLite is cloned from its floating 'working' branch outside asg_client/, so an
+          # upstream change with no asg_client diff would otherwise reuse the stale APK. Fold the
+          # current 'working' SHA into the reuse key so such a change forces a fresh build.
+          spl_sha=$(git ls-remote https://github.com/Mentra-Community/StreamPackLite.git working \
```
aisraelov, line 668: P2 — Pin the build to the same StreamPackLite SHA that enters this reuse key. This samples `working` here, but the later clone still runs `git checkout working`; if the branch moves between these steps, the manifest records `hash(asg_tree, SHA_A)` for an APK built from `SHA_B`. The next run then mints a duplicate ASG version for identical bytes, and a move just after a matching reuse check can also leave the old artifact pinned for that run. Export `spl_sha` and check out that exact commit in the clone step so the recorded key and built input are atomic.

aisraelov (changes requested):
Re-review of `6f70c095`: the four blockers from my previous pass are materially addressed. The downgrade floor now fails closed when unset, the phone no longer polls/arms the progress-stall watchdog during a real recovery detour, the recovery revert predicate rejects the transient `-1` package state, and native legacy top-level manifests are upgrade-only.

I am still requesting changes for the two new inline findings below and three fresh open threads already present on this head: combined firmware+downgrade sessions never transition to the next pass; native Kotlin/Swift availability checks do not mirror the disabled/sub-floor downgrade gate; and stale `apk/install` state can latch a fresh session into detour mode before it sends `ota_start`. I did not duplicate those comments.

Validation on this exact head: mobile TypeScript compile passed; 117 focused OTA Jest tests passed; recovery downgrade tests passed; the repository Bluetooth SDK Android compile check and native `OtaManifestDowngradeTest` passed; workflow YAML and `git diff --check` passed. I also reproduced both coordinator issues with temporary regression tests (removed afterward): stale state entered `verifying` without querying/starting, and a completed firmware-first pass ended in the version-change global timeout. All GitHub CI checks are green.

mobile/modules/engine/src/services/otaDisplayState.ts
```diff
@@ -42,12 +42,42 @@ export function deriveDisplayState(args: {
...
+  // status still sitting in the store from before the handoff.
+  if (versionChangeConverged) return "complete"
+
+  // In a version-change session, convergence (above) is the ONLY completion for the ASG
+  // downgrade. A bare non-firmware "complete" before convergence — the glasses refusing
+  // the downgrade (floor/gate) with no install, or a stale pre-handoff status — must never
+  // present as success. Firmware-step completions (mtk/bes) DO pass through: a combined
+  // downgrade+firmware manifest applies firmware first on its own pass, which must finish
+  // so the phone re-checks and re-offers the now-firmware-free downgrade.
+  const isFirmwareComplete = otaStatus?.stepType === "mtk" || otaStatus?.stepType === "bes"
```
aisraelov, line 76: **P2 — Don't present a firmware-only pass as the completed version change.** This lets the firmware step reach `displayState === "complete"`, but the coordinator snapshot still exposes `isVersionChange: true`. `progress.tsx` therefore renders “Version Change Complete” and says the glasses are on the required version and their settings were reset—even though this first pass installed only firmware and the APK downgrade is still pending. Please distinguish this terminal firmware pass in the snapshot/UI (or otherwise clear the version-change presentation for this pass) before navigating back to re-check.

aisraelov (changes requested):
Re-review of `474254899f4fe788ec6662b25db6646a08637b95`.

The latest three commits address the previously open stale-status, firmware-first timeout, and native floor-checker findings; I reproduced the first two before the fixes and confirmed their regression cases now pass. I am still requesting changes for:

1. The recovery/remediation serialization race: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641547701
2. The StreamPack branch/SHA race: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641547703
3. The new firmware-only completion copy issue noted inline here.

Validation on this head: mobile TypeScript compile; 117 focused OTA tests; both temporary coordinator regressions (stale pre-session status and firmware-first completion); native Kotlin downgrade-manifest test. The review worktree is clean.

mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/OtaManifest.kt
```diff
@@ -95,15 +100,46 @@ internal object OtaManifestChecker {
...
+                )
+            }
+            return false
+        }
+        if (serverVersion > currentVersion) {
+            return true
+        }
+        // Downgrade: only actionable at/above the enabled floor; a non-positive floor disables
+        // downgrades (fail closed, matching ASG and the engine).
+        return downgradeFloorVersionCode > 0 && serverVersion >= downgradeFloorVersionCode
```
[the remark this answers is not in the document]
aisraelov: Confirmed in both Kotlin and Swift. The existing `exactPinIsNotAnUpdate` test passes only because it uses the default disabled floor (`0`); with any positive floor at or below the pin, equality reaches this return and becomes `true`. Please add the explicit `< currentVersion` guard and cover equality with an enabled floor.

PhilippeFerreiraDeSousa:
@aisraelov Fourth round (5d34678dc0 → 74e7e95a24):

**Remediation deleted (per @PhilippeFerreiraDeSousa).** The remediation force-install path was the source of the install-race complexity, and it never shipped to prod — factory reset already covers catastrophic recovery. Removed the whole package, scheduling, constants, tests, and the ASG-side deferral hook. That dissolves the P1 remediation race entirely. Heartbeat recovery stays and now rechecks the downgrade transaction at its install boundary (top of `RecoveryWorker` and again immediately before `ReinstallStrategy`), so an already-enqueued worker can't reinstall the higher backup over an active downgrade — the surviving half of your P1.

**Exact pin reported as update (confirmed):** the native floor gate dropped the `< currentVersion` guard. Added the equality short-circuit in Kotlin and Swift, plus a test at an enabled floor.

**Firmware-only pass shown as completed version change:** exposed `versionChangeConverged` in the coordinator snapshot; the progress screen now shows the "Version Change Complete / settings restored" copy only on real convergence — a firmware-only pass shows the normal "Update complete" before re-check.

**StreamPackLite SHA atomicity:** the reuse step now exports the exact `working` SHA it sampled and the clone step checks out that commit (not a later tip), so the recorded key and the built input are atomic.

Green: mobile Jest (560), ASG + recovery unit tests, Kotlin SDK checker tests, tsc, Swift parse, YAML.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/remediation/RemediationWorker.java
aisraelov: The remediation install path is gone, but the same race remains for `RecoveryWorker`: it checks `DowngradeTransactionStore.isActive()` at lines 115–119, then calls `ReinstallStrategy.execute()` at line 123. `requestDowngrade()` can persist a transaction between those operations, and neither side holds a shared lock/cancels-and-joins the other work; `ReinstallStrategy` can then uninstall/install the higher backup while `DowngradeWorker` starts. The second check narrows the window but does not serialize it, so this thread remains actionable.

mobile/modules/engine/src/services/otaDisplayState.ts
aisraelov: This removes the premature “Version Change Complete / settings reset” copy, but the fallback copy is still false for this state: the firmware-only pass now says “Your glasses are up to date” while the APK downgrade is still pending. Please render firmware-pass-specific completion copy (for example, firmware updated / continue to finish the version change) rather than the generic up-to-date message.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/service/RecoveryService.java
```diff
@@ -47,7 +47,7 @@ public void onCreate() {
     InstallPauseNotifier.setListener(paused -> healthMonitor.setPaused(paused));
     registerReceivers();
     healthMonitor.start();
-    RemediationController.schedule(this);
+    DowngradeController.resumeIfActive(this);
```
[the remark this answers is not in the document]
aisraelov: Confirmed. `giveUp()` clears the transaction after post-revert failures, and the resumed heartbeat monitor will accept a responsive factory ASG as healthy, so `RecoveryWorker` will not reinstall the fleet backup. With `RemediationController.schedule()` removed, there is no automatic path back to a supported fleet build; this also contradicts the still-current `DowngradeWorker` contract and PR description. Please add an explicit post-revert recovery action before deleting remediation.

.github/workflows/staging-builds.yml
```diff
@@ -585,7 +638,88 @@ jobs:
...
+            | awk '{print $1}')
+          if [ -z "$spl_sha" ]; then
+            echo "Could not resolve StreamPackLite working SHA; building fresh to be safe." >&2
+            echo "reuse=false" >> "$GITHUB_OUTPUT"
+            echo "tree_hash=" >> "$GITHUB_OUTPUT"
+            exit 0
```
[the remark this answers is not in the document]
aisraelov: Confirmed. The empty-ref branch exits successfully after only setting `reuse=false`/empty `tree_hash`; `steps.asg-reuse.outputs.spl_sha` is therefore empty and the unconditional fresh-build clone step executes `git checkout ""`. Either export a valid fallback revision for that branch or fail explicitly in the reuse step instead of claiming the fresh build can continue.

aisraelov (changes requested):
Re-review of `74e7e95a24a50c6f454d20252512f8af21739838`.

The native exact-pin equality bug is fixed and covered with an enabled-floor test. The normal StreamPackLite path now checks out the same SHA used by the reuse hash, so I resolved that original thread. I also resolved the older phone-completion and legacy-manifest threads after verifying their current implementations.

Four issues remain:

1. Recovery install is still not serialized with a newly begun downgrade; the added second check leaves a check-to-install race: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641547701
2. Removing remediation leaves a failed post-revert downgrade on the factory build with no automatic supported-fleet recovery: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641665386
3. Firmware-first completion no longer claims the version change completed, but still falsely says the glasses are up to date while the downgrade is pending: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641569221
4. The unresolved-StreamPack-ref fallback exports no `spl_sha`, so the promised fresh-build path runs `git checkout ""`: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641665396

Validation on this exact head: mobile TypeScript compile; 117 focused OTA Jest tests; full recovery-worker unit tests; native Kotlin `OtaManifestDowngradeTest`; `git diff --check`. All GitHub CI checks are green, and the review worktree is clean.

PhilippeFerreiraDeSousa:
@aisraelov Fifth round addressed (270cd49f92 → 7fb31d913c), decisions vetted with @PhilippeFerreiraDeSousa:

**1. Install race — truly serialized now.** You're right that the double-check only narrowed it. Both paths run in the recovery app's single process, so `DowngradeTransactionStore` now owns a process-wide `ReentrantLock`: `RecoveryWorker` holds it across its pre-reinstall check **and** the `ReinstallStrategy` dispatch; `DowngradeController` holds it around `begin()`. A handoff lands either before the check (recovery sees it, stands down) or after the reinstall completed (the downgrade supersedes the reinstalled backup) — never interleaved. The lock releases before the post-install pong waits, where the mutation is already done.

**2. Post-revert give-up — deliberate decision: phone re-offer, no autonomous path.** Philippe chose to keep `giveUp()` as-is and align the contract instead. Rationale: the factory build is a functioning, pairable baseline; the next phone connection sees factory ≠ pin and re-offers the target through the normal phone-driven flow — from the factory floor that's a plain OEM **upgrade** install, a simpler and independently exercised mechanism than the recovery-side install that just failed. Re-adding an autonomous glasses-side install path would reintroduce exactly the racing installer the remediation deletion removed. The `DowngradeWorker` contract javadoc, `InstallPauseNotifier` docs, and the PR description are all updated to state this model (the stale "remediation restores the fleet build" claims are gone).

**3. Firmware-only pass copy:** it now says **"Firmware updated"** / "Your glasses restarted with new firmware. One more step: they'll now continue to the required version." with a **Continue** button back to the check screen (which re-offers the now-firmware-free downgrade). No more false "up to date".

**4. Empty `spl_sha`:** the fallback branch now fails the step with an explicit `::error` instead of claiming a fresh build can proceed into `git checkout ""`. (Longer term #3581 vendors StreamPackLite as a submodule and deletes this whole mechanism.)

Green on this head: mobile Jest (560), tsc, ASG + recovery compile/unit tests, YAML.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/remediation/RemediationWorker.java
aisraelov: The shared lock closes the check-to-begin gap, but it still ends before the conflicting install has completed. `ReinstallStrategy.execute()` returns immediately after `SystemInstaller.installApk()` sends an asynchronous broadcast to SystemUI; there is no completion receiver or observed-version wait inside that critical section. `RecoveryWorker` then unlocks, so `requestDowngrade()` can persist/start the downgrade while the higher backup install is still in flight, and that install may finish after the pinned target. Please keep ownership until install completion is observable (or add a completion/epoch protocol that prevents stale install work from committing after downgrade begin).

aisraelov (changes requested):
Re-review of `7fb31d913c93c12feee1976eca03ab6f88df2b36`.

Philippe's latest commits materially address three of the four findings from my previous pass: the post-revert behavior is now explicitly documented as a deliberate phone-owned re-offer contract, the firmware-first screen has accurate intermediate-pass copy, and an unresolvable StreamPackLite SHA now fails explicitly instead of reaching `git checkout ""`. I resolved the firmware thread after verifying its implementation; the other two threads were already resolved.

One P1 remains in the recovery serialization thread: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641547701. The new lock is released when `ReinstallStrategy.execute()` returns, but that method returns after merely sending the asynchronous SystemUI install broadcast—not after the backup APK has finished installing. A downgrade can therefore begin while the higher backup install is still in flight, allowing it to finish over the pinned target. The critical ownership needs to extend to observable install completion, or use an equivalent completion/epoch protocol.

Validation on this exact head: mobile TypeScript compile passed; 117 focused OTA Jest tests passed; the full recovery-worker unit suite passed; workflow YAML parsing and `git diff --check` passed. ASG, recovery, mobile quality, OEM boundary, cloud, console, SDK, store, Bugbot, and security checks are green; the mobile Android/iOS builds and aggregate CI gate were still running when this review was submitted. Both review worktrees are clean.

PhilippeFerreiraDeSousa:
@aisraelov Addressed in 46626722e6 — ownership now extends to observable completion, and the serialization moved entirely into the workers:

- **`RecoveryWorker`** holds the install lock across its liveness check, the backup reinstall dispatch, **and a bounded wait (`REINSTALL_OBSERVE_TIMEOUT_MS`, 60s) for the installed versionCode to reach the backup's** (`BackupStore.getBackupVersionCode()`), releasing only after the asynchronous OEM install is observed (or the window expires).
- **`DowngradeWorker`** takes the same lock for its **entire state-machine run**, so it cannot dispatch its first uninstall while a recovery install is in flight — it blocks until the reinstall has observably finished.
- **`requestDowngrade` drops its lock use** — a deliberate correction of my round-5 placement: it runs on a broadcast receiver's main thread, where blocking behind a legitimate tens-of-seconds hold would ANR. Persisting the transaction is safe at any time; only the workers dispatch installs, and they are now the mutually-exclusive parties.

On the "prevents stale install work from committing" alternative: the OEM installer offers no cancellation, so a commit can't be *prevented* once broadcast — but it can't stick either. One property that was already present is worth naming: `WAIT_FOR_REVERT` **re-dispatches the uninstall on every pass whenever a higher-than-target build (re)appears** (asg_client/recovery_worker/.../DowngradeWorker.java — "Re-dispatching the uninstall is harmless while still above target"). So even a straggler install that outlives the 60s observe window commits only to be uninstalled by the still-active transaction, which then converges to the pin. The residual is an install so stale it outlives the observe window *plus* the entire transaction; the phone's next version check heals that by re-offering the pin.

Recovery + ASG unit tests and compiles green on this head.

PhilippeFerreiraDeSousa:
Both round-7 findings addressed in ba5d4cef56, plus the frame that should end this series:

**Handoff watchdog races the lock hold (High):** real — same-sized 60s windows. `DOWNGRADE_HANDOFF_TIMEOUT_MS` is now 3 minutes: worst-case legitimate queue time (a backup reinstall holding the install lock for its full 60s observe window) + dispatch/uninstall time + margin, so a handoff queued behind an in-flight reinstall is no longer misreported as failed.

**Straggler commits after convergence (Medium):** real — the observe-timeout release left a window where a very late backup install could land after the transaction cleared. `ALREADY_AT_TARGET` now **lingers** for a bounded window (`DOWNGRADE_CONVERGENCE_LINGER_MS`, 60s) re-checking the installed version before clearing; any change resumes the state machine, which re-derives and re-uninstalls the higher build.

**The reliability invariant, now codified** (DowngradeWorker contract javadoc + PR description): the OEM installer is async, uncancellable, and callback-free, so glasses-side protocols can only *bound* divergence duration — they cannot close every window, and we should stop treating each next window as a correctness defect. Correctness lives in the phone's reconciliation loop: `version_info` vs pin on every connect/poll, re-offer on mismatch. Under that frame, a residual race in this package is a latency/UX bug — provided nothing glasses-side installs a build the phone did not sanction (sole exception: heartbeat recovery's backup reinstall, for the ASG-dead case where no phone channel exists, serialized via the install lock). @aisraelov — I'd welcome a review of that invariant statement itself; if the frame holds, remaining straggler analyses can be judged against "how long can the divergence last" rather than "does a window exist."

Recovery + ASG compiles and unit tests green on this head.

mobile/modules/engine/src/services/OtaInstallCoordinator.ts
```diff
@@ -218,6 +242,12 @@ class OtaInstallCoordinator {
     if (this.attached) return
     this.attached = true
     this.resetSessionState()
+    const selected = useGlassesStore.getState().otaUpdateAvailable
+    if (selected?.isDowngrade && typeof selected.versionCode === "number" && selected.versionCode > 0) {
+      this.versionChangeSession = true
+      this.versionChangeTarget = selected.versionCode
+      console.log(`[OTA_PROGRESS] version-change (downgrade) session, target=${selected.versionCode}`)
+    }
```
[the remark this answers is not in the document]
aisraelov: Confirmed independently with a temporary regression: after a downgrade session received its ack and latched `apk/install` (`versionChangePhase === "verifying"`), calling `retry()` invoked `startOtaUpdate()` once despite recovery owning the detour. The overwrite is especially unsafe because the running `DowngradeWorker` caches `target` before taking the lock but reads APK/SHA/attempt state from `DowngradeTransactionStore` later; a second `begin().clear()` can make the old worker combine its old target with the new transaction's mutable fields and eventually clear the replacement transaction. Retry should remain in reconcile/wait mode while the detour latch is set, and recovery should reject or generation-protect a handoff that tries to replace an active worker.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/remediation/RemediationWorker.java
aisraelov: The new wait still does not prove completion of the dispatched install in the common same-version case. `ensureRecoveryBackupIfNeeded()` normally archives the currently installed ASG, and `ReinstallStrategy` does not uninstall when backupVersion == installedVersion. After `installApk()` sends the replacement broadcast, `waitForInstalledAsgVersion()` sees that already-installed version on its first poll and returns immediately, so the lock is again released at dispatch rather than after this replacement commits. The later 60s target linger only bounds the likely window; an uncancellable install can still land after it. Please observe the operation's generation (for example, capture pre-dispatch `lastUpdateTime` and require a newer matching package, or an equivalent completion signal), not just version equality.

aisraelov (changes requested):
Re-review of `ba5d4cef5690b61ad1775ef7a92516600bfcd346`.

The latest commits materially improve serialization: `DowngradeWorker` now waits behind heartbeat recovery, the handoff watchdog covers that queue, and convergence lingers for late commits. I am still requesting changes for two blockers:

1. The recovery completion check is satisfied by the package that was installed before the new replacement broadcast. The normal backup is the current ASG version, so `waitForInstalledAsgVersion()` can return on its first poll and release the lock at dispatch again: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3641547701
2. Retry bypasses the version-change ownership latch and can send a second `ota_start`/handoff while the first transaction is live. `begin().clear()` can then overwrite preferences under a worker that cached the old target but reads the remaining fields dynamically: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3647649126

I independently reproduced #2 with a temporary regression assertion: after ack + `apk/install` latched the detour, `retry()` called `startOtaUpdate()` once; the assertion expected zero calls and failed. The temporary test was removed afterward.

Validation on this exact head: mobile TypeScript compile passed; 117 focused OTA Jest tests passed; the full recovery-worker test suite passed; the repository ASG Android compile check passed; and `git diff --check` is clean. Completed code/security CI checks are green. The PR-agent planner's failure is a transient `api.github.com` connection timeout, not a source failure; the ASG/mobile builds and aggregate gate are still running.

PhilippeFerreiraDeSousa:
@aisraelov Both blockers addressed (b30678d832, 667c47d673) — and thank you for the temporary-regression reproduction of #2, that is exactly the level of verification this deserves.

**1. Same-version reinstall observed at dispatch (your generation suggestion, adopted as proposed):** the backup normally archives the currently installed build, so version equality was satisfied by the pre-existing package on the first poll. `RecoveryWorker` now captures the package's `lastUpdateTime` **before** dispatching and `waitForAsgReplacementCommit` requires BOTH the backup's versionCode AND a newer `lastUpdateTime` — the generation advances only when the replacement actually commits, so the wait observes the operation itself in both the same-version-replace and uninstall-then-install paths. The timeout fallback (release + WAIT_FOR_REVERT/linger self-heal) is unchanged.

**2. Retry bypasses the detour latch:** fixed on both sides.
- *Phone:* `retry()` during `isInVersionChangeDetour()` no longer sends `ota_start` — it clears the surfaced error and re-enters the reconcile wait; convergence or the session timeout decides. Two jest regressions added: retry during a latched detour sends nothing (your reproduced scenario), retry outside a detour still re-drives.
- *Recovery (generation-protecting the handoff, per your suggestion):* `requestDowngrade` now **refuses** rather than REPLACEs when a transaction is active or an install worker holds the lock — non-blocking `tryLock` (receiver main thread, no waiting) doubles as the worker-running probe, with `isActive` checked inside it. No `begin().clear()` can land under a running worker anymore; `begin()`'s "always supersedes" javadoc is corrected. A refused handoff surfaces through ASG's watchdog and the phone re-offers after the live transaction converges or gives up; stale transactions cannot wedge the path (resumeIfActive + stale-timeout give-up).

Green on this head: mobile Jest 562 (incl. the new regressions), tsc, recovery-worker compile + full unit tests.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/downgrade/DowngradeController.java
```diff
@@ -0,0 +1,96 @@
...
+    // DowngradeWorker caches its target once but reads apkPath/sha/attempts from the store
+    // dynamically — an overwrite mid-run would make it mix the old target with the new
+    // transaction's fields and later clear the replacement. A handoff that arrives while a
+    // transaction is active (or while a worker holds the install lock) is refused; ASG's
+    // handoff watchdog then reports failure to the phone, whose reconciliation loop re-offers
+    // the pin once the current transaction has converged or given up. tryLock (non-blocking,
+    // we are on a broadcast receiver's main thread — no waiting allowed) doubles as the
+    // is-a-worker-running probe; stale transactions cannot wedge this path forever because
+    // resumeIfActive re-arms them and DOWNGRADE_TRANSACTION_STALE_MS forces give-up.
+    if (!DowngradeTransactionStore.installLock().tryLock()) {
```
aisraelov, line 51: P1 — Refusing the handoff here can strand the phone in a detour that recovery never accepted. ASG emits `install/STARTED` before broadcasting `ACTION_REQUEST_DOWNGRADE`, so the coordinator latches `versionChangeInstallStarted`; if ordinary heartbeat recovery currently owns this lock, `tryLock()` returns false and no transaction is persisted. Three minutes later ASG reports `downgrade_handoff_failed`, but Retry enters the new detour branch, clears only `errorMsg` (leaving the failed `otaStatus`), and sends no new `ota_start`, so the UI stays failed with nothing that can converge. I reproduced that state with a temporary regression. Please queue acquisition off the receiver main thread (for example via `goAsync()`/work), add an acceptance signal before latching ownership, or explicitly allow this rejected-handoff error to reset and re-offer.

aisraelov (changes requested):
Re-review of `667c47d67346c16bb625377dd97380ae313d4b68`.

The two findings from `ba5d4cef` are materially addressed: recovery now observes a new package generation rather than pre-existing version equality, and the phone no longer sends a second `ota_start` while a live detour is assumed. I resolved the prior recovery serialization thread after verifying the `lastUpdateTime` implementation.

One P1 remains in the new inline finding. `requestDowngrade()` can refuse a handoff merely because heartbeat recovery holds the install lock, but ASG emitted `apk/install STARTED` before sending that handoff, so the phone has already latched the detour as recovery-owned. After the handoff watchdog reports `downgrade_handoff_failed`, Retry takes the new no-drive branch even though no transaction exists; it also leaves the failed `otaStatus` intact, so the button remains on the failure screen.

I reproduced that exact rejected-handoff state with a temporary regression: after latching `apk/install`, setting `otaStatus.status = "failed"` with `downgrade_handoff_failed`, and pressing Retry, `displayState` remained `failed`. The temporary test was removed afterward.

Validation on this exact head: mobile TypeScript compile passed; 119 focused OTA Jest tests passed; the full recovery-worker test suite passed; the repository ASG Android compile check passed; and `git diff --check` is clean. CI restarted for this head and was still running when this review was submitted.

mobile/modules/engine/src/services/OtaInstallCoordinator.ts
```diff
@@ -255,6 +285,18 @@ class OtaInstallCoordinator {
...
+    // converged), the phone must not drive: an ota_start now would push the
+    // factory/surviving build into a parallel download-install, and a second
+    // handoff would begin()/REPLACE the live transaction under the running
+    // worker. Retry just re-enters the reconcile wait — convergence or the
+    // session timeout decides the real outcome.
+    if (this.isInVersionChangeDetour()) {
+      console.log("[OTA_PROGRESS] retry during version-change detour — re-entering wait (no ota_start)")
+      this.setErrorMsg("")
+      this.emitInternalChange()
+      return
```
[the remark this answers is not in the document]
aisraelov: Confirmed independently. The cross-process trigger is `DowngradeController.requestDowngrade()` returning when `tryLock()` loses to ordinary heartbeat recovery: ASG has already emitted `apk/install STARTED`, but no downgrade transaction is persisted. I reproduced the resulting phone state with a temporary regression (`otaStatus.status = "failed"`, `error = "downgrade_handoff_failed"` after the detour latch): pressing Retry left `displayState === "failed"` and sent no replacement handoff. I posted the same finding moments after this thread was created; resolving my later thread as a duplicate.

PhilippeFerreiraDeSousa:
Follow-up hardening (b8e359504d), prompted by the retry finding: the detour ownership gate was enforced per entry point, which is exactly the topology that let the retry hole exist — a future entry point that forgets its gate would corrupt the transaction silently. `sendOtaStartWithWatchdogs` now carries an **invariant backstop**: it refuses to send while the detour is latched and logs `BACKSTOP` at error level. The site gates remain the behavior policy (each entry point still does the right thing *instead*); the backstop is pure invariant enforcement, so a forgotten gate degrades to a refused send + actionable log instead of a parallel install and a clobbered handoff. A firing backstop always means a missed site gate exists. Jest regression drives the sender directly (simulating an ungated future caller) and asserts the refusal and log. Full mobile suite green (563).

mobile/modules/engine/src/services/OtaInstallCoordinator.ts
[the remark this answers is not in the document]
aisraelov: Re-reviewed at `b8e35950`. The new `sendOtaStartWithWatchdogs()` backstop is good defense against future ungated callers, but it does not repair this failure path: when recovery's `tryLock()` rejects the broadcast, ASG has already emitted `install/STARTED` (latching the detour), no recovery transaction exists, and the watchdog later reports `downgrade_handoff_failed`. Retry returns at the detour gate and the new backstop would also refuse a send, so the failed state still has no convergence path. The new test covers only an unsafe direct send during a valid detour; it does not cover a rejected/failed handoff. Reopening this thread.

aisraelov (changes requested):
Re-reviewed through `b8e35950`. The invariant backstop is correct hardening and its focused test passes, but it does not resolve the remaining P1 in the reopened **Failed detour blocks retry forever** thread: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3647741886

If recovery rejects the handoff because the install lock is busy, ASG has already emitted `install/STARTED`, so the phone has latched detour ownership even though no recovery transaction exists. The watchdog then emits `downgrade_handoff_failed`; Retry returns at the detour gate, and the new backstop independently refuses `ota_start`. The user remains failed with nothing able to converge. Please add a real rejected-handoff recovery path (for example accepted ownership signaling, queued acquisition off the receiver thread, or a safe reset/re-offer for this explicit failure).

Validation on this head: mobile TypeScript compile passes, all 49 focused coordinator tests pass, and `git diff --check` is clean. The preceding `667c47d` recovery/ASG changes were also validated with recovery Gradle tests and the ASG compile check.

PhilippeFerreiraDeSousa:
@aisraelov The reopened P1 is addressed in 1c2b0d9852, via your third option — a safe reset on the explicit failure — chosen deliberately over the other two:

**The fix:** `downgrade_handoff_failed` is an explicit statement from the glasses that nothing owns the detour — the precise negation of what the latch asserts — so the coordinator now releases `versionChangeInstallStarted` when that error arrives during an unconverged detour. Retry then re-drives normally: ASG re-stages and re-hands-off; recovery accepts once the lock is free or the prior transaction has converged/given up. If recovery refuses again, the cycle repeats as failed → user retry while the live transaction finishes — churn, bounded, no deadlock, and convergence at the end. Jest regression covers the full sequence (latch → refusal → failed → retry re-sends `ota_start`); the backstop test still passes because the latch is genuinely released, not bypassed.

**Why not the other two options:** accepted-ownership signaling is the theoretically cleanest but adds a new cross-package protocol message (recovery→ASG→phone ack) whose absence must itself be timed out — a new state machine to review on a path that already has an authenticated failure signal. Queued acquisition off the receiver thread only improves first-attempt success for the lock-busy case; it does not resolve the transaction-already-active refusal, so the phone-side release is needed regardless — and once it exists, queued acquisition is an optimization, not a correctness requirement. Happy to add it as a follow-up if you think the first-attempt UX matters enough.

Green on this head: coordinator suite 50, full mobile Jest 564, tsc.

PhilippeFerreiraDeSousa:
Bugbot's unknown-BES finding addressed in febc4792e8 — with one wrinkle beyond the report: fixing only the ordering gate would have traded the eternal deferral for something worse. `checkBesUpdate`'s unknown-version→flash policy is shared by the gate AND phase 2's apply decision, so excluding unknown from the gate alone would let the same pass hand off the downgrade *and* start a blind BES flash — and the detour's uninstall kills ASG, which IS the UART flasher, mid-transfer.

The fix makes one session-wide decision instead: `besUpdateApplicableForSession` — with a pinned downgrade pending (`isPinnedDowngradePending`: apps entry + `DowngradeGate.shouldDowngrade`) and no cached BES version, **BES is out of scope for the whole session**, applied consistently at all three rootJson-scoped sites (ordering gate, `buildStepSequence` step accounting, phase-2 apply). No eternal deferral, no blind flash racing the uninstall, and step counts stay consistent with what the pass actually does. The detour reboots the glasses, BES reports its real version at boot (`hs_syvr`), and the phone's next check re-offers the firmware with a known version — the reconciliation loop doing its job. Plain upgrade sessions keep the flash-to-be-safe semantics untouched, including the updater-internal precheck (only reachable when phase 2 already decided to flash, i.e. non-downgrade sessions).

ASG compile + full unit tests green on this head.

mobile/modules/engine/src/services/OtaInstallCoordinator.ts
```diff
@@ -575,6 +646,67 @@ class OtaInstallCoordinator {
...
+    ) {
+      console.log("[OTA_PROGRESS] version-change: apk install started — entering detour wait")
+      this.versionChangeInstallStarted = true
+      this.emitInternalChange()
+    }
+
+    // Ownership refuted: ASG emits install/STARTED BEFORE handing off, so the latch above can
+    // be set even when recovery then REFUSES the handoff (install lock busy, transaction
+    // already active) and no transaction exists. ASG's handoff watchdog reports that exact
+    // condition as downgrade_handoff_failed — an explicit statement from the glasses that
```
aisraelov: P1 — This error does not actually prove that recovery refused ownership. `OtaHelper` emits `downgrade_handoff_failed` whenever ASG is still alive after 180s. An accepted transaction can still be active then: `DowngradeWorker` waits 90s for the factory revert, returns `Result.retry()`, and can be inside its next 90s wait at the watchdog boundary (the work is also non-expedited). Clearing the latch lets Retry enter `stageAndHandoffDowngrade()`, which deletes and rewrites the shared staged-APK path *before* `DowngradeController` can reject the second handoff on its lock/active-store checks, racing the live worker's artifact. Please distinguish an explicit recovery refusal from a slow accepted transaction—e.g. add accepted/refused signalling with a transaction generation, or only release after recovery confirms no active transaction.

aisraelov (changes requested):
Re-reviewed through `febc4792`. The original failed-Retry dead end is fixed for a **confirmed rejected handoff**, and the unknown-BES policy is now scoped consistently across step planning, ordering, and application. One blocking ownership issue remains in the new latch release (inline): `downgrade_handoff_failed` is currently a 180-second ASG-side timeout, not a rejection acknowledgement, so it can fire while an accepted recovery transaction is still active. Re-driving then races the live transaction the detour gate was added to protect.

Please make recovery acceptance/rejection explicit (or otherwise confirm no active transaction) before releasing ownership. The BES policy should also get a direct regression test for unknown BES + actionable downgrade; the current focused tests compile the new path but do not exercise that branch.

Validation on this head: mobile TypeScript compile passes; all 50 focused coordinator tests pass; the ASG compile check passes; focused `OtaHelperBesGuardTest` and `DowngradeGateTest` pass; `git diff --check` is clean.

PhilippeFerreiraDeSousa:
@aisraelov You were right, and the round-9 reasoning was unsound: `downgrade_handoff_failed` only ever proved "ASG alive at 180s", which an accepted transaction legitimately is mid-revert-wait. Addressed in e3256c39a4 with the accepted/refused signaling you proposed — plus artifact ownership so the re-stage race is closed structurally, not just probabilistically:

**1. Synchronous verdict on every handoff.** Recovery now answers from the decision point itself (`ACTION_DOWNGRADE_HANDOFF_RESULT`, guarded like the ping/pong channel): refused with a reason on every reject path (floor, worker busy, transaction active, artifact claim failed, invalid), accepted after `begin()`. ASG's watchdog is cancellable: refused fails fast with the distinct `downgrade_handoff_refused`; accepted swaps the 180s watchdog for a 40-minute long-stop (past recovery's 30-minute stale give-up) reporting `downgrade_transaction_stalled` only if the accepted transaction demonstrably wedged. The 180s timeout now fires only when recovery never answered — and `MIN_RECOVERY_VERSION_FOR_DOWNGRADE` is bumped to v8 (verdict-capable), so a silent-but-accepting older worker cannot exist behind the handoff gate: "no answer" really does mean "no transaction".

**2. The transaction claims its artifact.** On acceptance, recovery **renames** the staged APK to a `.txn`-suffixed path before `begin()` persists it (atomic within the filesystem). A later ASG re-stage writes the original filename — a retry can never touch the bytes a live worker validates and installs, independent of any signaling edge case.

**3. Phone releases the latch only on the three non-ownership codes** (refused / no-verdict / stalled) — never on a generic failure, so an accepted-but-slow transaction keeps ownership through arbitrary phone-side failures/retries. Coordinator tests cover all three release codes plus an accepted-but-slow hold case (generic `install_failed` does NOT release).

Green: ASG compile + unit tests, recovery compile + unit tests + assembleDebug (v8), full mobile Jest 567, tsc.

PhilippeFerreiraDeSousa:
Bugbot's stale-asset-constant catch fixed in 45e6fa2967 — `ASSETS_RECOVERY_VERSION` was left at 7 when recovery moved to v8, so on the parse-failure fallback path a device already on v7 would skip the bundled-v8 redeploy while the handoff gate requires v8, permanently refusing every pinned downgrade. Bumped to 8 with the comment binding it to `recovery_worker`'s versionCode. (The bundled asset itself is built from the recovery project in CI, so it carries v8 automatically — only this fallback constant lagged.)

Also merged `dev` into the branch (453e930bdf) to clear the conflict from #3591 landing: the only collision was `BluetoothSdk.types.ts`, where dev added `glasses_session_changed` beside the `ota_update_available` entry this PR deletes — kept both intents (new event in, dead event out).

Green post-merge: full mobile Jest 580, tsc, ASG + recovery unit tests.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/downgrade/DowngradeController.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private static void enqueue(Context context, ExistingWorkPolicy policy) {
+    try {
+      // No network constraint: the APK is already staged and checksummed on local storage.
+      OneTimeWorkRequest request = new OneTimeWorkRequest.Builder(DowngradeWorker.class).build();
+      WorkManager.getInstance(context)
+          .enqueueUniqueWork(RecoveryConstants.UNIQUE_DOWNGRADE_WORK, policy, request);
+    } catch (Exception e) {
+      Log.e(RecoveryConstants.TAG, "Failed to enqueue downgrade work", e);
+    }
```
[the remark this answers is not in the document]
aisraelov: Confirmed independently. `enqueue()` swallows synchronous exceptions and returns no success signal, yet `requestDowngrade()` unconditionally sends `accepted`; moreover `enqueueUniqueWork()` exposes asynchronous operation failure that this code does not observe. With the store already active, the 40-minute long-stop cannot recover it because stale cleanup only runs inside `DowngradeWorker`, and subsequent handoffs are refused as `transaction_active`. Acceptance needs to follow a confirmed enqueue, or enqueue failure must roll back the store/claimed artifact and send refusal; please cover that failure path in a controller test.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/downgrade/DowngradeWorker.java
```diff
@@ -0,0 +1,380 @@
...
+            store.clear();
+            telemetry.emit(
+                "mentra_downgrade_applied", String.valueOf(target), "VERSION_CONVERGED", attempt, true);
+            return Result.success();
+
+          case SEND_UNINSTALL:
+            // Validate the staged APK BEFORE the first mutation: if it is unusable, abort with
+            // the current build still installed and untouched.
+            if (!isStagedApkValid(context, store)) {
+              return giveUp(store, telemetry, target, "STAGED_APK_INVALID", attempt);
```
aisraelov, line 131: P1 — An `accepted` handoff can still lose ownership here without telling ASG. `DowngradeController` sends acceptance before this worker performs the full APK validation; if package/version/signature/testOnly validation fails, `giveUp()` clears the store and emits only recovery telemetry. The original ASG is still installed and alive, but it has replaced the short watchdog with the 40-minute supervision timer, so the phone remains latched and cannot retry for 40 minutes even though no transaction exists. This is additional validation beyond ASG's SHA check, so a mis-published or wrongly signed artifact reaches it. Please either complete this validation before sending `accepted`, or send an authenticated terminal/non-ownership result when the worker gives up before uninstall, with a regression test.

aisraelov (changes requested):
Re-reviewed through `45e6fa29`. The explicit accepted/refused verdict, transaction-owned `.txn` artifact, version-8 recovery gate, and corrected asset fallback together resolve the previous timeout/artifact P1; I resolved that superseded thread.

Two terminal-path blockers remain:

1. The open **Accepted handoff with no worker** thread: https://github.com/Mentra-Community/MentraOS/pull/3526#discussion_r3648760332
2. A distinct pre-uninstall failure after acceptance (inline): full archive/package/version/signature validation can reject the staged APK immediately, but recovery only emits telemetry while ASG has already swapped to the 40-minute supervision timer.

Please make every accepted transaction either start a worker or roll back with refusal, and return a terminal non-ownership signal when a worker gives up before uninstall. The new cross-package verdict/claim paths also need direct recovery/ASG tests; the current added tests cover only the phone's error-code gating.

Validation on this exact head: mobile TypeScript compile and all 56 focused coordinator tests pass; recovery-worker debug/release JVM suites pass; ASG compile check passes; `git diff --check` is clean. CI was green except the Android/iOS mobile builds still running.

PhilippeFerreiraDeSousa:
**Hardware verification of the verdict protocol (e3256c39a4 / 45e6fa2967)** — physical Mentra Live (adb root), branch ASG `49347055-dev` + recovery worker v8 built from this head:

**Refused path (pristine v8, floor 0):** `ACTION_REQUEST_DOWNGRADE` → recovery rejected on the fail-closed floor and sent the verdict; ASG's `ServiceHeartbeatReceiver` received `accepted=false, reason=floor_rejected` over the permission-guarded channel and `OtaHelper` correctly identified no pending handoff. Full round-trip in **7 ms**.

**Accepted path (bench-floored build, target = installed):** verdict `accepted=true` delivered; the staged APK was **claimed by rename** (`asg_client_update.apk` → `.txn`) before the transaction persisted; worker derived `ALREADY_AT_TARGET` and held the **convergence linger for its full 60 s** (23:45:57 → 23:46:57) before `VERSION_CONVERGED`; claimed artifact deleted, transaction store cleared, heartbeat monitor resumed.

**Concurrent handoff during the linger:** refused with `accepted=false, reason=worker_busy` — the install-lock `tryLock` probe held while the worker owned the run, verdict delivered to ASG in 6 ms.

Device restored: pristine floor-0 v8 reinstalled, no stray artifacts, ASG alive with heartbeat flowing. Together with the earlier full uninstall→revert→reinstall transaction verification, every recovery-side surface of this protocol is now hardware-proven; the remaining unexercised path is the ASG-initiated handoff wiring (needs the full phone-driven flow), covered by unit tests.

PhilippeFerreiraDeSousa:
@aisraelov Both terminal-path blockers addressed in 0641040ac3 — the principle applied is the one your two findings share: **"accepted" is a promise of ownership, so everything that can be known to fail must fail before the verdict, and everything that can still die after it must say so.**

**1. Unconfirmed enqueue (Controller):** `requestDowngrade` now blocks on the WorkManager `Operation` result (10s) before sending accepted; on sync exception, async failure, or timeout it **rolls back completely** — store cleared, claimed artifact renamed back — and refuses with `enqueue_failed`. The whole handoff moved off the receiver main thread (`goAsync` + dedicated executor) to make that blocking legal; `tryLock` stays non-blocking deliberately, since refusal-with-verdict is the protocol and queueing would just delay the verdict ASG is timing.

**2. Post-acceptance validation give-up (Worker, P1):** the full validation now runs **before** the verdict — `StagedApkValidator` (extracted from the worker, which delegates to it) validates the claimed artifact on the handoff path: bytes (presence/sha256) + archive (package, exact version, testOnly, signer match vs installed). A mis-published or wrongly-signed artifact is refused with the specific reason (`staged_apk_sha256_mismatch`, `staged_apk_signer_mismatch`, …) instead of accepted-then-abandoned.

**Defense in depth for the residue:** every `giveUp()` now broadcasts a terminal non-ownership verdict (`gave_up:<reason>`) through the same channel. ASG's armed supervision watchdog treats it exactly like a refusal — OTA latch and phone detour latch release immediately, not after 40 minutes. Post-revert give-ups send it too; the factory build has no receiver, so it drops harmlessly there.

**Tests:** this suite is plain JUnit (no Robolectric / work-testing), so controller-level coverage is via the extracted seams: `ArtifactClaimsTest` (claim/rollback semantics the enqueue-failure path depends on — rename, stale-claim replacement, exact-state restore) and `StagedApkValidatorTest` (the byte-stage mis-publish modes: missing/empty/sha-mismatch/case-insensitive match). If you want a true `requestDowngrade`-level test of the enqueue-failure path, I'd add `androidx.work:work-testing` + Robolectric to the recovery suite as a follow-up — happy to do it, flagging the toolchain addition rather than slipping it in.

Green: recovery compile + full unit tests (incl. the two new suites), ASG compile.

PhilippeFerreiraDeSousa:
**Hardware re-verification on the round-13 head (0641040ac3)** — the earlier hardware pass predated the pre-accept validation / confirmed-enqueue rework, so I re-ran it on the physical Mentra Live against this exact head (recovery rebuilt from it; the installed ASG already matches — r13 touched no ASG code):

- **Refused, floor path:** verdict `floor_rejected` — now demonstrably decided on the **handoff executor thread** (tid ≠ main), confirming the goAsync path.
- **Refused, validation path (new):** handoff with a wrong sha256 → recovery hashed the 105 MB artifact (~0.5 s), refused `staged_apk_sha256_mismatch`, and **rolled the claim back** — the file was restored to its original name, exactly the rollback the enqueue-failure path shares.
- **Accepted path:** handoff with the real sha (staged = the installed ASG APK, target = installed version) → full pre-accept validation passed (bytes + archive: package/version/testOnly/signer), transaction begun, **confirmed enqueue** succeeded, verdict `accepted`, artifact claimed to `.txn`, worker `ALREADY_AT_TARGET`, convergence linger held its full 60 s, `VERSION_CONVERGED`, artifact deleted, store cleared, heartbeat resumed.
- **Concurrent handoff during the linger:** refused `worker_busy` in 2 ms.

Device restored to the pristine floor-0 build, no test artifacts left. Every recovery-side surface of the current protocol is verified on hardware as shipped.

asg_client/recovery_worker/app/src/main/java/com/mentra/recovery/downgrade/DowngradeWorker.java
aisraelov: Re-reviewed at `0641040ac3`. This now closes the ownership gap in both directions: the controller claims and fully validates the archive before beginning the transaction, observes the WorkManager enqueue before sending `accepted`, and rolls the store/artifact back with an explicit refusal if enqueue cannot be confirmed. `RecoveryControlReceiver.goAsync()` keeps that blocking/hash work off the receiver main thread. The worker also emits a terminal non-ownership verdict from `giveUp()`, which the still-running original ASG consumes through its armed supervision watchdog; post-revert delivery is harmless because no handoff is pending. Focused recovery debug/release unit suites and the full ASG compile check pass locally. Resolved.

aisraelov (approved):
Re-reviewed the new recovery ownership patch at `0641040ac319190a5f8827a5008093c9b72b9e82`. The two outstanding blockers are addressed: acceptance now follows full staged-APK validation plus a confirmed WorkManager enqueue (with transaction/artifact rollback and refusal on failure), and terminal worker give-up releases ASG's long supervision wait. I also audited the enqueue-timeout race, artifact claim rollback, worker revalidation, post-revert verdict behavior, and receiver threading; I found no new blocking issue.

Local validation: recovery-worker `testDebugUnitTest` + `testReleaseUnitTest` passed, full ASG compile check passed, and `git diff --check` is clean. Exact-head CI is green for recovery, ASG, mobile quality, iOS, cloud/SDK/web, OEM boundary, security, and automated review checks; Android mobile build/the aggregate gate were still in progress when this review was submitted.

PR #3558, opened 2026-07-23
Persist local miniapp storage by user

aisraelov:
## Summary

- scope bundled miniapp SimpleStorage and blob storage to the stable Core-owned Mentra user ID instead of a rotating access token
- wait for authenticated Core identity before selecting storage namespaces, preserving blob request order during first-boot sign-in
- reject anonymous storage namespaces and redact token/email values from settings write logs
- add restart, account-isolation, and storage-operation coverage

## Root cause

Core access tokens are intentionally memory-only and are refreshed or re-minted after a process restart. Local miniapp storage used the entire access token as its user namespace, so force-stopping the Mentra App made the next launch unable to find settings such as Translation's saved target language.

## User impact

Local miniapp preferences now survive force-stops and token refreshes while remaining isolated by Mentra user and package.

## Validation

- `bun test src/services/__tests__/LocalMiniappStorage.test.ts src/services/__tests__/blobPaths.test.ts` — 20 passed
- `bun run build` in `mobile/modules/engine`
- `bun compile` in `mobile`
- changed-file ESLint — 0 errors (one pre-existing `BluetoothSdk` import warning)
- `git diff --check`

## Baseline issues observed

The repository-wide engine test command still has unrelated existing Bluetooth mock/module-resolution failures (280 tests pass; failures are outside this diff). The repository-wide engine lint also reports existing formatting errors outside the changed surface.

---

> [!NOTE]
> **Medium Risk**
> Changing storage and blob namespace keys can strand data written under the old token-based paths unless users migrate manually. Blob queuing and identity timing affect first-boot and account-switch behavior.
> 
> **Overview**
> **Local miniapp SimpleStorage and blob storage no longer key off the rotating Core access token** (or an `anonymous` fallback). They now use the stable Core `mentraUserId`, persisted in MMKV via new `LocalMiniappUserIdentity` and resolved through `cloudClientService.resolveMentraUserId()` / `getMentraUserId()`.
> 
> `LocalMiniappStorage` centralizes MMKV key prefixes per user + package; `LocalMiniappRuntime` routes all storage RPCs through it. **Blob paths, metadata keys, and usage cache** in `BlobStore` follow the same identity, with usage cached per user+package scope. **All `BLOB_*` handlers are queued per app** until identity resolves, with guards so stale app instances do not run after unregister; pre-sign-in failures return a clear not-connected error.
> 
> Identity is **remembered on Core token sync**, **cleared on logout** (`stop()`, `SIGNED_OUT`, and `SIGNED_IN` to avoid cross-account bleed). Settings logging **redacts token/email** values. Unit tests cover identity caching and storage isolation; bundled miniapp manifest list drops `com.mentra.enterprise`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36b9a77fc913db07b5d22141488cc3805efd44c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Persist miniapp SimpleStorage and blob storage by the stable Mentra user ID so preferences survive restarts, work offline, and stay isolated per user and app. Blob ops are queued until a stable user ID is available; legacy token‑scoped blobs are migrated to the user namespace, and ops are blocked if migration can’t safely finish.

- **Bug Fixes**
  - Add `LocalMiniappStorage` and `LocalMiniappUserIdentity`; cache `mentraUserId` in MMKV for offline use, forget on sign‑in/out and on `cloudClientService.stop()`; expose `cloudClientService.resolveMentraUserId()`/`getMentraUserId()`.
  - Scope SimpleStorage keys and blob paths/usage to `mentraUserId` + package; update BlobStore usage cache to be per user+package.
  - Queue/serialize all `BLOB_*` per app behind identity resolution; allow offline access with cached identity; return a clear pre‑sign‑in error if identity isn’t available.
  - One‑time blob migration from legacy token paths with safety (no overwrites, clean up old paths, block access on failed copy); redact `token`/`email` in settings logs; add unit tests for identity, restart persistence, and storage isolation.

<sup>Written for commit 36b9a77fc913db07b5d22141488cc3805efd44c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3558?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


isaiahb (commented):
Reviewed the rekey and both migration paths in detail. The architecture is sound: keying on the stable Core user id is right, the sanitizeSegment truncation trick that bridges full JWTs to the old 120-char blob namespaces checks out against blobPaths, and cross-account attribution via the sub claim embedded in legacy keys is safe. Tests are thorough and CI is green.

One finding I consider blocking, one worth fixing before merge, and three nits.

1. Blocking: offline cold start now breaks storage entirely. Every SimpleStorage and blob op awaits resolveMentraUserId, which awaits auth.getCoreToken, and access tokens are held in memory only. So the first storage op after every process start requires a network round trip. If the app cold-starts offline (or Core is unreachable), bundled miniapps lose their settings: the exact symptom this PR fixes, retriggered by "offline at launch". The old token-prefix scheme was wrong but worked offline. Suggestion: persist the mentraUserId (stable, non-secret) in MMKV once known, and fall back to it in resolveMentraUserId when the token exchange cannot complete.

2. SimpleStorage migration removes legacy keys without verifying the migrated writes landed. localMiniappStorageBackend.set ignores the mmkvStorage.save result, and migrateLegacyTokenScope deletes every legacy key right after copying. The blob path carefully checks storage.save and blocks on failure; the SimpleStorage path should match (or re-read the stable key before deleting), otherwise a failed write silently loses the user's newest settings.

3. Nit: enqueueBlobRequest serializes BLOB_SET_FROM_URL / BLOB_IMPORT / BLOB_SHARE, which previously ran concurrently. A slow URL download now head-of-line blocks that app's entire blob queue. If the strict ordering is intentional, a comment saying so would help.

4. Nit: a legacy blob scope that can never be attributed keeps complete=false forever, so every blob op for that app re-scans all MMKV keys, and its files are never reclaimed. Consider caching the unresolved verdict after a few attempts, or a cleanup path for unattributable scopes.

5. Nit: the getAccessToken hook can throw (getCoreAccessToken throws when resolvedCoreAccessToken is null, e.g. right after construct resets it). Inside the blob migration the optional call guards undefined but not a throw, so the whole blob op fails instead of deferring migration. Wrapping it and treating a throw as "no current token" degrades more gracefully.

Happy to run the force-stop repro on a real device against this branch if useful.

PR #3572, opened 2026-07-23
Fix Mentra AI requests in the Android background

aisraelov:
## Summary

- avoid Activity-bound Expo location permission prompts while the Mentra App is backgrounded
- add `mediaPlayback` to the existing Mentra Android foreground service and its manifest permissions
- skip the optional silent PCM route prewarm while backgrounded so it cannot block real URL/TTS playback
- reconcile a stale native “Mentra audio is playing” hint with Android's actual media state so the glasses mic watchdog can recover
- count miniapp liveness from ping rounds actually sent, preventing a paused React Native timer from unregistering Mentra AI immediately on resume
- add regression coverage for background playback, audio-state recovery, permission handling, and miniapp liveness

## Root cause

The incident logs identify three independent background/resume failure modes.

First, incident `rep_01KY6V1N08HW4PMVJJ19MQBWM0` showed that the Mentra foreground service was already running as `connectedDevice|microphone|location`, but `Location.requestForegroundPermissionsAsync()` still blocked. Expo's Android implementation delegates that call to `Activity.requestPermissions(...)` even when foreground location is already granted, so the callback waited for Activity resume. The host now reads `getForegroundPermissionsAsync()` first. Existing grants proceed immediately, while a missing grant only opens Expo's prompt path if the app is still active.

Second, `AudioPlaybackService` started its optional cold-A2DP silent PCM prewarm in the background; that prewarm completed only when the Activity became active, releasing queued sound at once. Background URL playback now bypasses that optional optimization and starts the real player directly. The persistent Mentra foreground service also declares and activates Android's `mediaPlayback` type.

Finally, incident `rep_01KY6WTT74Z49QY0G5JEWSV1CT` showed TTS completion without the debounced native `STOPPED` notification. The stale `PLAYING` hint then blocked the native glasses-mic recovery watchdog every ten seconds even though Android reported no active media. `PhoneAudioMonitor` now expires that hint after its short codec-transition holdoff when Android media is inactive, allowing mic recovery on the next watchdog pass.

The same incident also showed the React Native ping interval being paused for roughly two minutes in the background. On resume, elapsed wall time was incorrectly interpreted as six missed pings, so Mentra AI was unregistered and its transcription subscription—and home-screen mic icon—disappeared. Liveness now counts actual unanswered ping rounds instead of wall time, so resume sends a fresh ping rather than killing a context that was never probed.

## Validation

- `bun test src/services/__tests__/MiniappLiveness.test.ts src/services/__tests__/AudioPlaybackService.test.ts` (21 pass)
- `./gradlew :mentra-bluetooth-sdk:testDebugUnitTest --tests com.mentra.bluetoothsdk.utils.PhoneAudioMonitorTest --no-daemon`
- `bun compile`
- forced `:app:processDebugMainManifest` merge; packaged service includes `dataSync|connectedDevice|location|mediaPlayback|microphone`
- `git diff --check`

---

> [!NOTE]
> **Medium Risk**
> Touches foreground service types, background audio/location paths, and miniapp unregister logic—important for glasses connectivity and Mentra AI, but changes are targeted with unit tests.
> 
> **Overview**
> **Fixes Mentra AI and glasses-initiated flows when the Android app is backgrounded or resumes**, addressing several independent failure modes seen in production.
> 
> The **Mentra Bluetooth foreground service** now always includes **`mediaPlayback`** (manifest + Expo plugin + runtime type detection) so background TTS/prompts are allowed without spinning up a second service. **`AudioPlaybackService`** skips the optional silent PCM cold-route prewarm when **`AppState` is not active**, so glasses prompts are not held until the Activity resumes.
> 
> **`PhoneAudioMonitor`** reconciles the RN “own app playing” hint with **`isMusicActive`**: it still honors the short codec-transition holdoff, but clears a stuck **PLAYING** flag when Android reports no media so the glasses mic watchdog can recover.
> 
> **Miniapp location polls** use **`resolveForegroundLocationPermission`**: read grant first and only call Expo’s Activity-bound **`requestForegroundPermissionsAsync`** while foreground. **Miniapp liveness** counts **unanswered ping rounds** instead of wall-clock time since last pong, so a paused RN timer in background no longer unregisters Mentra AI immediately on resume.
> 
> Regression tests cover audio prewarm skip, location permission resolution, ping liveness, and **`shouldKeepOwnAppAudioSignal`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae39297d6de7e1c7012ef6c5356c49c781f0d543. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/modules/bluetooth-sdk/android/src/main/AndroidManifest.xml
+++ b/mobile/modules/bluetooth-sdk/android/src/main/AndroidManifest.xml
@@ -29,6 +29,7 @@
   <uses-permission android:name="android.permission.FOREGROUND_SERVICE_CONNECTED_DEVICE"/>
   <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC"/>
   <uses-permission android:name="android.permission.FOREGROUND_SERVICE_LOCATION"/>
+  <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK"/>
   <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE"/>
 
   <!-- Other permissions -->
@@ -36,6 +37,6 @@
   <uses-permission android:name="android.permission.VIBRATE"/>
 
   <application>
-    <service android:name="com.mentra.bluetoothsdk.services.ForegroundService" android:exported="false" android:foregroundServiceType="dataSync|connectedDevice|location|microphone"/>
+    <service android:name="com.mentra.bluetoothsdk.services.ForegroundService" android:exported="false" android:foregroundServiceType="dataSync|connectedDevice|location|mediaPlayback|microphone"/>
   </application>
 </manifest>
--- a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/services/Foreground.kt
+++ b/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/services/Foreground.kt
@@ -81,6 +81,14 @@ class ForegroundService : Service() {
 
         var serviceType = 0
 
+        // Audio prompts can be initiated by glasses while the host Activity is
+        // backgrounded. mediaPlayback has no runtime prerequisite, so keep it
+        // active on the existing Mentra service for the entire connected
+        // session rather than trying to launch a second service after a wake
+        // phrase arrives.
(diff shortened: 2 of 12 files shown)


aisraelov:
@codex review

aisraelov:
@codex review

PR #3581, opened 2026-07-24
Vendor StreamPackLite as a git submodule

PhilippeFerreiraDeSousa:
Stacked on #3526 (`pferreira/ota-downgrade-detour-rfc`). Retarget to `dev` once that lands, or merge in order.

## What

Vendor **StreamPackLite** (the RTMP streaming library) as a git submodule at `asg_client/StreamPackLite`, pinned to the current `working` tip (`e7a4517`), instead of cloning its floating `working` branch at build time.

## Why

The floating-branch clone was the root cause of a string of OTA review comments on #3526:

- the ASG reuse hash didn't cover StreamPackLite, so an upstream change with no `asg_client/` diff reused a stale APK;
- the fix sampled the branch tip via `ls-remote` and folded it into the reuse key…
- …which then needed the clone step to `git checkout` that exact SHA to stay atomic with the key;
- and builds were never reproducible — `working` moves, so an old commit can't be rebuilt against the StreamPackLite it shipped with.

Each of those was a workaround *around* the floating branch. A submodule fixes the root cause and lets us **delete** the workarounds.

## What this buys

- **Reproducible builds** — any checkout gets the exact StreamPackLite it was built against.
- **Reuse hash covers StreamPackLite for free** — the pinned commit is a gitlink inside `git ls-tree HEAD:asg_client`, so a bump changes the tree hash and forces a fresh build. The `spl_sha` `ls-remote` sampling, its fail-safe, and the atomic-checkout dance added in #3526 are all removed.
- **Explicit, reviewable updates** — moving StreamPackLite is a gitlink bump commit, not an invisible upstream branch move that silently changes every build.
- **One setup path** — three CI workflows now use `actions/checkout` with `submodules: recursive`; the local compile script and setup docs use `git submodule update --init asg_client/StreamPackLite`.

## Changes

- Add the submodule + `.gitmodules` entry (`branch = working`); drop `asg_client/StreamPackLite/` from `.gitignore`.
- `staging-builds.yml`: checkout `submodules: recursive`; revert the reuse hash to the plain `asg_client/` tree hash (gitlink carries StreamPackLite); delete the `spl_sha` sampling/export and the manual clone step.
- `mentra-asg-client-build.yml`, `bluetooth-sdk-release.yml`: checkout `submodules: recursive`; delete the manual clone steps.
- `scripts/check-android-compile.sh`, `asg_client/AGENTS.md`, `asg_client/README.md`: submodule init instructions.

## Verification

- `./scripts/check-android-compile.sh asg` → **BUILD SUCCESSFUL** against the submodule.
- Confirmed the gitlink appears in `git ls-tree -r HEAD:asg_client` (`160000 commit e7a4517…StreamPackLite`), so the reuse hash covers it.
- StreamPackLite is public (the old workflow cloned it unauthenticated), so `submodules: recursive` resolves with the default CI token.

---

> [!NOTE]
> **Medium Risk**
> Changes affect all ASG CI/release build paths and staging APK reuse semantics; misconfigured submodule init or a bad gitlink could break builds or skip needed rebuilds, but scope is limited to build/dependency wiring rather than runtime product logic.
> 
> **Overview**
> **StreamPackLite** is no longer cloned from the floating `working` branch at CI time; it is vendored at `asg_client/StreamPackLite` via `.gitmodules` (with `branch = working`), and the old `.gitignore` entry for that path is removed.
> 
> ASG workflows (`staging-builds`, `mentra-asg-client-build`, `bluetooth-sdk-release`) replace manual clone/checkout with **path-scoped** `git submodule update --init asg_client/StreamPackLite`, gated where builds can be skipped, and document why recursive init is avoided. **Persistent runners** also wipe `asg_client/StreamPackLite` in the pre-checkout clean step so stale Gradle outputs in the submodule worktree cannot leak across runs.
> 
> **Staging ASG reuse (OS-1689)** drops the `git ls-remote` / `spl_sha` composite hash and checkout-at-SHA logic; the pinned submodule **gitlink** in `git ls-tree HEAD:asg_client` is enough to invalidate reuse when StreamPackLite moves.
> 
> Docs and `scripts/check-android-compile.sh` now tell developers to init the submodule instead of cloning manually.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a6c6c5036683f57271c0df35bc339775ebe8590. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PhilippeFerreiraDeSousa:
Both review comments addressed (1e389cca3c), plus the branch is synced with the base's round-5 changes (merge ae2b851636 — the conflict was exactly the `spl_sha` block this PR deletes; kept the deletion).

**Recursive checkout inits broken submodule (Bugbot High):** verified precisely — the `mcu_client/peripheral_uart_next/nrf5340` entry is *dangling*: it exists in `.gitmodules` but has **no committed gitlink** (`git ls-files -s` is empty), and submodule init iterates index gitlinks, so `submodules: recursive` would not actually fail today. Path-scoped anyway: the three workflows now run an explicit `git submodule update --init asg_client/StreamPackLite` step instead, which removes the trap entirely if that gitlink ever lands and matches the docs. (StreamPackLite has no own `.gitmodules`, so non-recursive is complete.)

**README init fails from `asg_client/` (codex P2):** correct — the surrounding steps' implied cwd is `asg_client/`, where the repo-root-relative path errors with a pathspec failure. The README now uses the cwd-relative form (`git submodule update --init StreamPackLite`), and AGENTS.md + the compile script hint state "from the repository root" explicitly.

Validated by running both documented forms against an empty submodule dir (root form = CI's exact command; asg_client-relative form = README's) — both populate `e7a4517` — followed by `./scripts/check-android-compile.sh asg` → BUILD SUCCESSFUL. All three workflow YAMLs parse.

.github/workflows/bluetooth-sdk-release.yml
```diff
@@ -150,13 +150,19 @@ jobs:
 
       - name: Clean build artifacts (preserve caches)
         run: |
-          rm -rf asg_client/app/build asg_client/.gradle asg_client/StreamPackLite
+          rm -rf asg_client/app/build asg_client/.gradle
```
aisraelov: **[P2] Keep cleaning the submodule on the persistent release runner**

This job runs on `[self-hosted, macOS, ARM64]`. Once `StreamPackLite` is a tracked gitlink, the superproject clean performed by `actions/checkout` does not descend into the initialized submodule, and `git submodule update --init` does not remove ignored build outputs. I reproduced this after the compile check: `core/build/`, `extensions/rtmp/build/`, and `extensions/srt/build/` remained, while a root `git clean -ndx` did not list them. Removing `asg_client/StreamPackLite` from this cleanup therefore lets those products persist across SDK release runs (and through the retry cleanup), unlike the old clone flow. Please keep clearing the submodule/build directories here, or hard-reset and clean the submodule after initialization, so this release build remains clean and reproducible.

.github/workflows/staging-builds.yml
```diff
@@ -631,13 +630,19 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@v4
         with:
           ref: ${{ github.sha }}
 
+      - name: Init StreamPackLite submodule
+        # Path-scoped on purpose: .gitmodules also carries a dangling
+        # mcu_client/peripheral_uart_next/nrf5340 entry with an unfetchable relative
+        # self-URL; a blanket recursive init would break if its gitlink ever lands.
+        run: git submodule update --init asg_client/StreamPackLite
```
aisraelov: **[P2] Initialize the submodule only when a fresh APK is needed**

The reuse decision below now hashes the gitlink from `HEAD:asg_client`; it does not need the submodule working tree. Because initialization is unconditional and happens first, every app-only staging release still clones/fetches StreamPackLite and can fail on a transient submodule fetch even when the existing APK is fully reusable. Please move this step after `asg-reuse` and guard it with `if: steps.asg-reuse.outputs.reuse != 'true'`. The `sdk-ota` job in `bluetooth-sdk-release.yml` has the same ordering issue relative to `existing-sdk-ota`, so its init should likewise run only when that manifest is not being reused.

aisraelov (commented):
Reviewed the current head end to end. The pinned gitlink, path-scoped initialization, workflow syntax, local compile check, and CI build all validate; I found two remaining workflow reliability issues inline.

aisraelov (approved):

PhilippeFerreiraDeSousa:
@aisraelov Both P2s addressed (00c06834e7):

**Stale submodule outputs on the persistent runner:** the clean steps in all three workflows remove `asg_client/StreamPackLite` again — you're right that `actions/checkout` doesn't descend into submodule worktrees and init doesn't touch ignored outputs, so the old `rm -rf` was load-bearing on the self-hosted mac. Removal stays cheap under the submodule model: the git dir lives in `.git/modules`, so init re-checks-out locally rather than re-cloning.

**Unconditional init before the reuse decision:** init now runs *after* the decision and gated exactly like the old clone step — `asg-reuse != true` in staging-builds, `existing-sdk-ota != true` in bluetooth-sdk-release — so app-only releases neither fetch StreamPackLite nor can they fail on a transient submodule fetch. Verified the steps between checkout and each decision are pure git/curl (no Gradle in `asg_client`, which would need the worktree). `mentra-asg-client-build` always builds, so its init stays unconditional right after checkout.

All three workflow YAMLs parse.

PR #3589, opened 2026-07-24
Add opt-in muteMicWhileSpeaking to speaker.speak()

isaiahb:
**Draft on purpose. Do not merge without a decision on the blast radius below.**

Stacked on #3586 (base is that branch, so the diff here is only the new work; GitHub retargets to `dev` once #3586 merges).

## What

#3586 removed the platform's automatic mic mute during TTS. That left the mechanism plumbed but unreachable: a miniapp had no way to ask for it. This adds the way in.

```ts
await session.speaker.speak("Here is your answer.", {muteMicWhileSpeaking: true})
```

Off by default. `LocalMiniappRuntime` reads it as `payload.muteMicWhileSpeaking === true` and threads it into all three playback paths (cloud TTS, offline single-shot, offline sentence pipeline).

## Why this is draft and not just merged

The suppression is device-wide. There is one microphone uplink and one cloud STT stream shared by every running miniapp, so a miniapp setting this does not mute its own transcription. It mutes Captions, Live Translation, and everything else running, for the duration of the playback plus the 700ms A2DP tail, with no indication to the user or to the apps being silenced.

So this hands any miniapp, including third-party ones, a device-wide kill switch over everyone else's transcription behind a one-line option. That is a real product decision, not a mechanical follow-up, and it should be made deliberately.

Two things partially blunt it, neither of which is a fix:

- The option's doc comment states the blast radius in the API surface itself, where an author will actually read it.
- `LocalMiniappRuntime` logs a warning naming the requesting package on every arming, so a support case that starts "captions stopped for no reason" is traceable to a caller instead of being a mystery.

The actual fix is OS-1741's `mic.status` / `mic.onStatusChange`, which would let a muted app see that it has been muted and by whom. Until that exists, this is observable in logs only.

## Options if the blast radius is not acceptable

1. Merge as is, accept it, and prioritise `mic.status` right after.
2. Restrict the option to first-party packages (Mentra AI, Merge) until scoping exists.
3. Scope the suppression per-subscriber rather than per-uplink, so an app only mutes its own transcript delivery. Bigger change, and the right end state.
4. Leave this branch parked and do nothing, since AEC in firmware may remove the need entirely (OS-1799).

I lean 1 or 4 depending on how the firmware test goes.

## Test

- `mobile/modules/miniapp`: 253 pass, 0 fail. Added two cases pinning that the default is off across both `speak()` overloads and that an explicit opt-in is forwarded.
- `mobile/modules/engine`: 320 pass, 14 fail, identical to a clean `dev` checkout (pre-existing `PhoneCameraFovCoordinator` failures).
- `tsc --noEmit` clean for every changed file.
- Not exercised on device. #3586 has the device traces showing the underlying mechanism does what the counters say when it is armed.

---
## Summary by cubic
Adds an opt-in `muteMicWhileSpeaking` to `session.speaker.speak()` to suppress mic audio to cloud STT while TTS plays. Off by default; suppression is device-wide and will mute Captions/Live Translation for all apps during playback.

- New Features
  - Threads `muteMicWhileSpeaking` through `LocalMiniappRuntime` to `AudioPlaybackService`/`AudioCloudUplink` across all playback paths, and logs the requesting package when suppression is armed.
  - Tests cover default-off behavior and forwarding of an explicit opt-in; future `mic.status` (OS-1741) will let muted apps detect and attribute the suppression.

<sup>Written for commit 01d9da527fcef0e28f2a296f1f597e7b8ad91b82. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/mobile/modules/engine/src/services/AudioCloudUplink.ts
+++ b/mobile/modules/engine/src/services/AudioCloudUplink.ts
@@ -26,10 +26,13 @@ const suppressionSources = new Set<string>()
  * reopen the uplink until every active source has released.
  *
  * Suppression is global to the uplink: one source silences STT for every app.
- * That is why no platform path arms it any more (spoken TTS used to, which made
- * every assistant reply punch a hole in Live Translation and Captions). The
- * mechanism stays because it is the right primitive for an explicit,
- * caller-scoped mute; see OS-1741 for the per-app opt-in that will use it.
+ * That is why nothing arms it by default. Spoken TTS used to, unconditionally,
+ * which made every assistant reply punch a hole in Live Translation and
+ * Captions. The only way in now is a miniapp explicitly passing
+ * `speaker.speak(text, {muteMicWhileSpeaking: true})`, which LocalMiniappRuntime
+ * logs on every arming because the blast radius reaches apps that never asked
+ * for it. See OS-1741 for the mic.status work that would make it visible to
+ * the apps being muted rather than only in the logs.
  */
 export function setAudioCloudUplinkSuppressed(sourceId: string, suppressed: boolean): void {
   if (suppressed) suppressionSources.add(sourceId)
--- a/mobile/modules/engine/src/services/AudioPlaybackService.ts
+++ b/mobile/modules/engine/src/services/AudioPlaybackService.ts
@@ -16,7 +16,8 @@ interface AudioPlayRequest {
    * Suppress microphone audio sent to cloud STT while this audio is audible.
    * Opt-in and off by default: the suppression is global to the uplink, so a
    * caller that sets this silences STT for every running app, not just itself.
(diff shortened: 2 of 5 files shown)


aisraelov:
I don't think this is required now that BES has basic echo cancellation on it

PR #3591, opened 2026-07-24
Reify the ASG-to-BES UART link state into a LinkStateMachine

PhilippeFerreiraDeSousa:
## Why

"Connected" on the glasses side of the BES UART link was five different implicit facts, each a volatile boolean (or a boolean aspect of a timestamp) in `K900BluetoothManager`, each with its own ad-hoc reset sites:

1. `isSerialOpen` — serial port open (set in `onSerialReady`/`onSerialClose`/`onSerialOpen`)
2. the boolean aspect of `lastSrSyvrTime` — link proven alive at the current baud (any parsed `sr_syvr`)
3. the negotiated BES caps: `besWireCapsK900Le`, `besWireCapsBinary`, `besWireCapsFilePayloadV2`, `besSupportsBigPacks`, `besWireCapsProto` — reset only in `resetWireProtocolState()` on serial close
4. phone-facing wire state (`wireV2HandshakeSent`, BesWireFormat binary-active, reassembler) — reset on serial close and `onTransportReset`
5. the byte-stream discontinuity fact — implicit in `clearMessageParser()`, the shared prelude of all eight reopen/baud-switch call sites (`SerialPortBridge.reopen()` fires no serial callbacks)

Because these facts had no single owner and no observability, components that registered after the only edge of one of them simply never learned about it — the bug class behind the recent week of production link bugs.

## What

New `LinkStateMachine` (in `io/bluetooth/managers/mentralive/internal/`, next to `SerialPortBridge`/`BesWireFormat`) that owns the transport-side facts:

- **States:** `SERIAL_CLOSED → SERIAL_OPEN → LINK_PROVEN` (proven = `sr_syvr` parsed at the current baud).
- **Transitions** driven from the exact existing edges in `K900BluetoothManager`:
  - `serialReady()` / `serialClosed()` from the serial callbacks (`onSerialOpen(false)` maps to `serialClosed()`);
  - `streamDiscontinuity()` from `clearMessageParser()` — covers all eight reopen/baud-switch paths (boot probes, target-baud probe, revert-to-default, runtime recovery, rendezvous park, post-BES-OTA reconnects); drops `LINK_PROVEN → SERIAL_OPEN`;
  - `srSyvrParsed(caps)` from the `sr_syvr` handler (merges the wire_caps advertisement with the legacy accrual rules: flags only turn on, `proto` overwritten only when `binary` is advertised);
  - `binaryRelayObserved()` from valid inbound binary frames (binary=true, proto floored at v2, no state change — a relayed frame does not prove the sr_syvr round trip).
- **Observable with replay-on-subscribe:** `addListener()` synchronously replays the current state under the machine's monitor, so a late subscriber cannot miss an edge. All mutation + notification happen under one monitor → listeners see one global transition order.
- **Two caps views, on purpose:**
  - `getProvenCaps()` (also the listener payload) is non-null **exactly** in `LINK_PROVEN` — the snapshot new consumers may trust.
  - `getNegotiatedCaps()` is the sticky negotiation result, cleared only on serial close. This mirrors the legacy `besWireCaps*` reset rules bit-for-bit: a mid-session baud reopen never cleared them, and letting a reopen window flip `isBesBinaryRelaySupported()`/`file_payload_v2` to false could downgrade a whole phone session (e.g. a `glasses_ready` response built during the window would advertise no binary support). The legacy delegating getters read this view, which is what keeps this PR a strict-parity refactor.
- `K900BluetoothManager` keeps thin delegating getters (`isBesBinaryRelaySupported()`, `addPhoneWireCapsIfSupported()`, the file-transfer capability checks) so no external call site changes. The old booleans are deleted; `isConnected()` is still `serialOpen && super.isConnected()` — semantics unchanged.
- `lastSrSyvrTime` is kept as the baud-recovery **timing** input (last-proof wall clock); only its boolean "link proven" role moved into the machine.
- New consumers subscribe via `K900BluetoothManager.getLinkStateMachine()`.

## Deliberately out of scope

- **`PHONE_SESSION` state** — needs BES phone-presence reporting; separate PR.
- **Phone-facing wire state** (`wireV2HandshakeSent`, BesWireFormat binary-active, inbound reassembler) stays in `resetPhoneWireProtocolState()` — it belongs to the phone session, not the UART transport.
- **`OtaHelper.setPhoneConnectionProvider` replay workaround** stays: its "connected" is the phone BLE session (`onPhoneConnected` consumes `pendingApkStatus`), which this transport machine does not model. It becomes deletable once `PHONE_SESSION` lands.
- Non-transport consumers are untouched; only the internal owner changed.

## Behavior notes (parity)

- No timing changes, no new sends, no change to when any existing method returns true/false.
- One deliberate deviation from a spec-literal reading: `streamDiscontinuity()` hides the *proven* caps view but keeps the *negotiated* caps, because clearing them at every reopen would be a behavior change with session-long consequences (see above). The machine documents both views.
- Edge preserved: today `applyBesWireCaps` runs before the `B`-field parse that can throw in `handleSrSyvrResponse`, so caps could be recorded even when `lastSrSyvrTime` was not updated; the machine's `srSyvrParsed` sits at the same point, so caps accrual is unchanged (machine-proven can lead `lastSrSyvrTime` in that rare edge — new observable surface only, nothing existing reads it).

## Tests

New `LinkStateMachineTest` (22 tests) covering: replay-on-subscribe (late listener gets the current state synchronously), discontinuity dropping proven + hiding caps while negotiated caps survive, `srSyvrParsed` with/without caps, serial close resetting everything, caps accrual/merge rules (proto overwrite only with binary advert; binary-frame observation floors proto at v2), listener ordering, self-removal during notification, and a two-thread hammer verifying the (state, provenCaps) invariant never tears.

```
./gradlew :app:testDebugUnitTest compileDebugJavaWithJavac
# BUILD SUCCESSFUL — 583 tests, 0 failures (22 new)
```

---

> [!NOTE]
> **Medium Risk**
> Touches critical K900 UART recovery, baud switching, and wire-cap gating for file transfer and phone SDK behavior; intended as parity refactor but regressions could affect production link stability.
> 
> **Overview**
> Introduces **`LinkStateMachine`** as the single owner of ASG-to-BES UART transport facts (serial open, link proven via `sr_syvr`, negotiated BES wire caps), with **replay-on-subscribe** listeners so late subscribers cannot miss edges that previously lived in scattered volatile booleans.
> 
> **`K900BluetoothManager`** wires serial callbacks, `sr_syvr` / binary-frame handling, and every reopen/baud path into the machine; removes `isSerialOpen` and `besWireCaps*` fields in favor of thin getters and **`getLinkStateMachine()`**. Adds **`reopenSerial()`** (syncs closed port when `reopen` fails without callbacks), **`streamDiscontinuity()`** on parser clear and BES OTA reboot, and fixes runtime recovery **park** to avoid link updates under `baudSwitchLock` and to ignore stale generations.
> 
> **`SerialPortBridge.isOpen()`** exposes post-reopen port state. Unit tests cover the state machine and closed-port reopen behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bcde96c55314686a1aa9e495f4b6f45f1d011e2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
@@ -10,6 +10,7 @@ import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.BesMessag
 import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.BesWireFormat;
 import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.CsFltsAckPayload;
 import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.K900LengthCodec;
+import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.LinkStateMachine;
 import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.MessageChunker;
 import com.mentra.asg_client.io.bluetooth.managers.mentralive.internal.SerialPortBridge;
 import com.mentra.asg_client.io.bluetooth.utils.DebugNotificationManager;
@@ -45,7 +46,11 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
     private static final String TAG = "K900BluetoothManager";
 
     private final SerialPortBridge comManager;
-    private boolean isSerialOpen = false;
+
+    // Single owner of the transport-side link facts (serial open, link proven at current baud,
+    // negotiated BES wire caps). The legacy isSerialOpen/besWireCaps* booleans are derived from it
+    // through thin getters so external call sites keep their existing surface.
+    private final LinkStateMachine linkState = new LinkStateMachine();
     private final DebugNotificationManager notificationManager;
     // Keep normal K900 messages contiguous on the shared ASG-to-BES UART stream.
     private final Object outboundSendLock = new Object();
@@ -61,10 +66,6 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
     // interleaves messages. Bounded: abandoned reassemblies would leak entries, so the set
     // is cleared when it exceeds a size no legitimate interleave reaches.
     private final java.util.Set<Integer> pendingBinaryWakeMsgIds = new java.util.LinkedHashSet<>();
-    private volatile boolean besWireCapsK900Le = false;
-    private volatile boolean besWireCapsBinary = false;
-    private volatile boolean besWireCapsFilePayloadV2 = false;
-    private volatile int besWireCapsProto = BesWireFormat.PROTOCOL_VERSION_V1;
     private volatile int gattFilePackSize = BesWireFormat.FILE_PACK_SIZE_DEFAULT;
     private volatile int lastNegotiatedMtu = 0;
     private volatile boolean phoneSupportsFilePayloadV2 = false;
@@ -109,10 +110,9 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
     // first ACK repeatedly overran the BES parser on hardware and triggered 150 ms retries.
     private static final int FILE_PUSH_WINDOW_BIG_PACKS = 8;
     private volatile boolean besSupportsBatchedAcks = false;
-    private volatile boolean besSupportsBigPacks = false;
 
     private int effectivePushWindow() {
-        if (besSupportsBigPacks) {
+        if (linkState.getNegotiatedCaps().bigPacks) {
             return FILE_PUSH_WINDOW_BIG_PACKS;
         }
         return besSupportsBatchedAcks ? FILE_PUSH_WINDOW_BATCHED : FILE_PUSH_WINDOW;
@@ -458,7 +458,7 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
             return false;
         }
 
-        if (!isSerialOpen) {
+        if (!linkState.isSerialOpen()) {
             Log.w(TAG, "📡 ❌ Cannot send data - serial port not open");
             notificationManager.showDebugNotification(
                     "Bluetooth Error", "Cannot send data - serial port not open");
@@ -564,23 +564,22 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
         wireV2HandshakeSent = false;
     }
 
-    /** Reset all negotiated wire protocol state when the BES UART link is closed. */
+    /**
+     * Reset the remaining wire protocol state when the BES UART link is closed. The negotiated BES
+     * caps are cleared by {@code linkState.serialClosed()}; this covers the phone-facing state and
+     * the UART length endianness, which live outside the link state machine.
+     */
     public void resetWireProtocolState() {
         resetPhoneWireProtocolState();
-        besWireCapsK900Le = false;
-        besWireCapsBinary = false;
-        besWireCapsFilePayloadV2 = false;
-        besSupportsBigPacks = false;
-        besWireCapsProto = BesWireFormat.PROTOCOL_VERSION_V1;
         uartToBesEndian = K900LengthCodec.Endian.BE;
     }
 
     /** Send BLE Wire v2 handshake frame (payload "v2"). */
     public boolean sendWireV2Handshake() {
-        if (!isSerialOpen) {
+        if (!linkState.isSerialOpen()) {
             return false;
         }
-        if (!besWireCapsBinary) {
+        if (!linkState.getNegotiatedCaps().binary) {
             Log.i(TAG, "📡 Skipping BLE wire v2 handshake; BES has not advertised binary relay");
             return false;
         }
@@ -684,10 +683,8 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
         if (!header.valid) {
             return false;
         }
-        besWireCapsBinary = true;
-        if (besWireCapsProto < BesWireFormat.PROTOCOL_VERSION_V2) {
-            besWireCapsProto = BesWireFormat.PROTOCOL_VERSION_V2;
-        }
+        // A valid binary frame proves binary relay support even without a wire_caps advert.
+        linkState.binaryRelayObserved();
 
         if ((header.flags & BesWireFormat.FLAG_HANDSHAKE) != 0) {
             if (BesWireFormat.isV2HandshakePayload(header.payload)) {
@@ -762,20 +759,20 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
      * reachable.
      */
     public void addPhoneWireCapsIfSupported(JSONObject response) {
-        if (response == null
-                || (!besWireCapsK900Le && !besWireCapsBinary && !besWireCapsFilePayloadV2)) {
+        LinkStateMachine.BesCaps besCaps = linkState.getNegotiatedCaps();
+        if (response == null || (!besCaps.k900Le && !besCaps.binary && !besCaps.filePayloadV2)) {
             return;
         }
         try {
             JSONObject caps = new JSONObject();
-            if (besWireCapsK900Le) {
+            if (besCaps.k900Le) {
                 caps.put("k900_le", true);
             }
-            if (besWireCapsBinary) {
+            if (besCaps.binary) {
                 caps.put("binary", true);
-                caps.put("proto", Math.max(besWireCapsProto, BesWireFormat.PROTOCOL_VERSION_V2));
+                caps.put("proto", Math.max(besCaps.proto, BesWireFormat.PROTOCOL_VERSION_V2));
             }
-            if (besWireCapsFilePayloadV2) {
+            if (besCaps.filePayloadV2) {
                 caps.put("file_payload_v2", true);
                 caps.put("file_payload_gatt_max", BesWireFormat.FILE_PACK_SIZE_GATT_MAX);
                 caps.put("file_payload_coc_max", BesWireFormat.FILE_PACK_SIZE_COC_MAX);
@@ -786,8 +783,18 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
         }
     }
 
+    /** Whether the BES has proven binary relay support (wire_caps advert or an observed frame). */
     public boolean isBesBinaryRelaySupported() {
-        return besWireCapsBinary;
+        return linkState.getNegotiatedCaps().binary;
+    }
+
+    /**
+     * Observable transport link state (serial open, link proven, negotiated BES caps). New
+     * consumers should subscribe here: {@code addListener} replays the current state
+     * synchronously, so subscribing late cannot miss the only edge.
+     */
+    public LinkStateMachine getLinkStateMachine() {
+        return linkState;
     }
 
     /** Queues an internal command without broadcasting it as a phone-facing command response. */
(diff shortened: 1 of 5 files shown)


asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/LinkStateMachine.java
```diff
@@ -0,0 +1,327 @@
...
+    /** Must hold the monitor. */
+    private BesCaps provenCapsLocked() {
+        return state == LinkState.LINK_PROVEN ? negotiatedCaps : null;
+    }
+
+    /** Must hold the monitor. */
+    private void notifyListenersLocked() {
+        BesCaps provenCaps = provenCapsLocked();
+        for (Listener listener : listeners) {
+            listener.onLinkStateChanged(state, provenCaps);
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Fixed in d83501fdde: notifyListenersLocked now snapshots both state and provenCaps into locals before the loop, so every listener in one notification pass receives the consistent pair of the transition that triggered it; a reentrant transition delivers its own pass. Added `reentrantTransitionInListener_neverTearsThePairForLaterListeners`: listener A calls serialClosed() from inside its LINK_PROVEN callback, and listener B is asserted to receive the reentrant (SERIAL_CLOSED, null) pass plus the original consistent (LINK_PROVEN, caps) outer pass — never a torn pair. Full suite green (584 tests).

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/LinkStateMachine.java
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Fixed in d83501fdde (same fix as the duplicate Codex thread): notifyListenersLocked snapshots both state and provenCaps before the loop, so a reentrant transition from inside a callback can no longer hand later listeners in the same pass a torn (SERIAL_CLOSED, non-null caps) pair. Covered by the new `reentrantTransitionInListener_neverTearsThePairForLaterListeners` test; full suite green (584 tests).

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
```diff
@@ -2390,7 +2409,7 @@ public void onSerialReady(String serialPath) {
         Log.d(TAG, "🔌 =========================================");
         Log.d(TAG, "🔌 Serial path: " + serialPath);
 
-        isSerialOpen = true;
+        linkState.serialReady();
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Fixed in 229a8e49e2: onBesOtaApplied now calls linkState.streamDiscontinuity() synchronously, right where lastSrSyvrTime is zeroed (and outside baudSwitchLock so listener callbacks never run under it), so subscribers stop observing LINK_PROVEN the moment the BES reboot invalidates the proof — instead of waiting for the async rendezvous path's clearMessageParser, which a generation-mismatch early return in prepareRendezvousAfterBesOta can skip entirely. The call is idempotent if that path does run it (streamDiscontinuity is a no-op unless LINK_PROVEN — covered by streamDiscontinuity_whenNotProven_isANoOp). No manager-level unit test: constructing K900BluetoothManager pulls in SerialPortBridge/SerialManager native serial I/O, which is not JVM-testable; the machine-level idempotency plus the one-line synchronous call site is the verifiable surface. Full suite green (584 tests).

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/LinkStateMachine.java
```diff
@@ -0,0 +1,335 @@
...
+        return state == LinkState.LINK_PROVEN ? negotiatedCaps : null;
+    }
+
+    /**
+     * Must hold the monitor. Snapshots BOTH state and caps before the loop: the monitor is
+     * reentrant, so a listener may perform a transition from inside its callback, and reading the
+     * {@code state} field live per iteration would then hand later listeners in this pass a torn
+     * pair (e.g. {@code SERIAL_CLOSED} with non-null caps). With the snapshot, every listener in
+     * one pass sees the consistent pair of the transition that triggered it; the reentrant
+     * transition delivers its own pass.
```
aisraelov: The snapshot fixes the torn pair, but it still violates the advertised global transition order. With the reentrant close in the new test, listener B receives `SERIAL_CLOSED` from the nested pass and then receives the outer, now-stale `LINK_PROVEN` event; the machine ends closed while B's last observation says proven. A real consumer can therefore re-enable work after the link has closed. Please serialize/queue reentrant notifications (or otherwise prevent reentrant transitions during dispatch) so every listener sees `LINK_PROVEN` before `SERIAL_CLOSED`; the test at lines 357-363 currently codifies the unsafe reverse order.

aisraelov (commented):
I’m not approving the current head because the new observable state contract still has correctness gaps; I left three inline findings. I also confirmed the unresolved Cursor finding on this same head: a runtime-recovery task can pass its generation guard, then a valid frame can cancel recovery/re-prove the link, and the stale task can still call `clearMessageParser()` and demote that fresh proof; `parkRuntimeRecoveryAtRendezvous` additionally calls it before its generation guard. The build and tests are green, but these races affect the production semantics this PR is introducing.

Validation: `./gradlew :app:testDebugUnitTest compileDebugJavaWithJavac` — BUILD SUCCESSFUL.

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/LinkStateMachine.java
```diff
@@ -0,0 +1,335 @@
...
+     */
+    private void notifyListenersLocked() {
+        LinkState snapshotState = state;
```
aisraelov: One throwing listener aborts this loop, prevents all later listeners from receiving the transition, and propagates into the UART caller. In the `srSyvrParsed` path, `handleSrSyvrResponse` then reports a parse failure and falls through after the machine has already mutated, which makes the replay/ordered-delivery contract unreliable. Please catch/log per-listener failures and continue delivery (and handle a failed initial replay so a permanently throwing listener does not poison every future transition), consistent with the existing transport/peripheral listener buses.

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
```diff
@@ -2520,7 +2545,14 @@ private void reconnectAfterBesOta(int generation, int attempt, String previousRe
...
     }
 
+    /**
+     * Shared prelude of every reopen/baud-switch path: the byte stream is about to be
+     * discontinuous, so drop any partially parsed frame and tell the link state machine the link
+     * is no longer proven at the current baud ({@code SerialPortBridge.reopen()} fires no serial
+     * close/ready callbacks, so this is the only signal those paths emit).
+     */
     private void clearMessageParser() {
+        linkState.streamDiscontinuity();
```
aisraelov, line 2585: `streamDiscontinuity()` can only demote to `SERIAL_OPEN`, but `SerialPortBridge.reopen()` can fail both the requested baud and its default-baud fallback, leaving `mbStart=false` with no serial callback. Every caller then leaves this machine reporting an open port indefinitely (`isSerialOpen()`/`isConnected()` and replayed state are wrong, while actual sends fail). Since this class is now the source of truth for the serial-open fact, please propagate the bridge's actual post-reopen open/closed result—likely via an `isOpen`/richer reopen result—and call `serialClosed()` on a true total failure.

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
```diff
@@ -2390,7 +2409,7 @@ public void onSerialReady(String serialPath) {
         Log.d(TAG, "🔌 =========================================");
         Log.d(TAG, "🔌 Serial path: " + serialPath);
 
-        isSerialOpen = true;
```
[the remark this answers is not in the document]
PhilippeFerreiraDeSousa: Fixed in bcde96c553: the generation guard now runs FIRST, so a stale park returns without touching anything — no clearMessageParser, no streamDiscontinuity, no reopen. The discontinuity + rendezvous reopen then run outside baudSwitchLock (link listeners must never be notified under the baud lock; all reopen paths are serialized on the single-threaded baud executor so nothing interleaves), and the retry is rescheduled only if the generation is still current. Full suite green (587 tests).

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/LinkStateMachine.java
PhilippeFerreiraDeSousa: Fixed properly in 992d27d800: notifications now go through a FIFO dispatch queue. A transition performed inside a listener callback is enqueued behind the pass in flight and drained after it completes, so every listener observes transitions in one global order and its LAST observation always matches the machine's final state — no stale proven pair can arrive after a close. Each queued pass snapshots its audience at transition time, so a listener added mid-dispatch (whose replay-on-subscribe already delivers the final current state) never also receives older queued passes. The reentrancy test now asserts B sees (SERIAL_OPEN, LINK_PROVEN, SERIAL_CLOSED) in global order and that B's last observation equals the machine's final state and provenCaps. Full suite green (587 tests).

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/mentralive/internal/LinkStateMachine.java
PhilippeFerreiraDeSousa: Fixed in 992d27d800 (same commit as the ordering fix — the guard is part of the new dispatch path): every delivery, including the synchronous replay in addListener, goes through deliverSafely, which catches and logs per-listener RuntimeExceptions and continues. A throwing listener no longer starves later listeners in the pass, and nothing propagates into the transport caller, so handleSrSyvrResponse can no longer log a bogus parse failure after the machine mutated. Covered by two new tests: throwingListener_doesNotStarveLaterListenersOrPropagateToTheCaller and throwingListener_duringReplayOnSubscribe_doesNotPropagate. Full suite green (587 tests).

asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
PhilippeFerreiraDeSousa: Fixed in cbece42025: SerialPortBridge now exposes the actual port state via isOpen() (mbStart), and every reopen call site routes through a new reopenSerial() helper that drives linkState.serialClosed() when the port genuinely ended up closed after the attempt (try/finally, so a throwing reopen is covered too). It checks the bridge rather than the boolean return, because reopen() returning false after a successful default-baud fallback still leaves the port open and must NOT close the machine. Once fully closed the bridge refuses further reopens, so the only path back to an open link is a fresh start(), which fires the normal onSerialOpen/onSerialReady callbacks the machine already consumes. Added a bridge-level test (reopenOnClosedPort_failsAndReportsClosed); the reopenSerial path itself is not JVM-testable since a real reopen needs the native serial stack. Full suite green (587 tests).

aisraelov (approved):

PR #3597, opened 2026-07-25
Confine camera teardown to the camera thread (OS-1816 follow-up)

PhilippeFerreiraDeSousa:
## Scope

Second PR of the OS-1816 series, stacked on #3596. The point-fix stops the crash; this PR removes the structural cause: camera teardown ran on six different thread classes (main, BLE RecvThread, NanoHTTPD workers, streaming-service threads, the StillCaptureCb handler thread, raw Timer threads), racing the open/session work that Camera2 delivers on the CameraNeoBackground thread — coordinated only by a best-effort semaphore that tears down unguarded after a 5s timeout, and bypassed entirely by two recovery paths.

## Changes

**Commit 1 — thread confinement.** `CameraCoordinator` gains `runOnCameraThread` (post; inline when already on the camera thread or when it's gone, so a close is never dropped) and `awaitOnCameraThread` (bounded wait for callers that need the eviction finished before proceeding). All fire-and-forget teardowns now route through it:

- `closeKeptAliveCamera` (video/stream starts evicting the kept-alive camera) — waits for the close, snapshots `sInstance` against a racing `onDestroy`, and re-checks shot state under `SERVICE_LOCK` on the camera thread so a photo that raced in keeps its camera.
- video `onSessionTerminated`, warm-up teardown (`stopCameraWarmUp`, `expireWarmLeases`), `onDestroy`, both `CameraRecoveryHelper` paths (which previously closed device/session with **no** lock at all), and `StillCaptureCallback`'s close hook.
- keep-alive expiry no longer runs teardown inline on the raw Timer thread when the camera thread is gone.
- `CameraCoordinator`'s device/session/handler/timer fields become volatile for the remaining cross-thread advisory reads.

`onDestroy` still guarantees the close completes before returning: `stopBackgroundThread`'s `quitSafely` + `join` drains the posted close first.

**Commit 2 — video open-failure routing.** `openCameraInternal`'s `InterruptedException`/generic handlers notified the *photo* error path even for video opens; and a MediaRecorder setup failure continued into the open only to fail again later with a second confusing error.

**Commit 3 — dead code.** Remove the caller-less `setupCameraForQueuedRequest` wrapper.

Deliberately *not* in this PR: the synchronous close+reopen inside photo dispatch (`PhotoSession.setupCameraForQueuedRequest`/`setupWarmUp`) stays on the dispatch thread — it is serialized by the open/close semaphore held through session creation since #3596, and confining it changes dispatch semantics (follow-up PR).

## Test evidence

- `./scripts/check-android-compile.sh asg` green after each commit
- `./gradlew :app:testDebugUnitTest`: 634 tests, 0 failures
- New `CameraCoordinatorTest` coverage: `runOnCameraThread` on/off-thread/no-thread, `awaitOnCameraThread` completion + timeout, keep-alive expiry now runs on the camera thread, and the raw-Timer-thread fallback skip (the old test encoding the removed fallback was updated to start the camera thread).

---

> [!NOTE]
> **Medium Risk**
> Touches core Camera2 lifecycle and multi-threaded teardown ordering; behavior changes are intentional hardening but affect video eviction, warm-up, and service destroy paths where races were previously possible.
> 
> **Overview**
> Follow-up to OS-1816: **camera teardown is serialized on the CameraNeo background thread** instead of running from main, timer, capture, and recovery threads that could race Camera2 open/session setup.
> 
> `CameraCoordinator` adds **`runOnCameraThread`** and **`awaitOnCameraThread`** (with **`CAMERA_TEARDOWN_AWAIT_MS`** for eviction). **`closeKeptAliveCamera`** now waits for that close and re-checks shot state under `SERVICE_LOCK` so a racing photo is not torn down mid-capture. Warm-up expiry, video session end, `onDestroy`, recovery helpers, and still-capture close hooks route through the same path; **`closeCameraFromWarmTeardown`** re-validates warm leases and shot state before closing. Keep-alive expiry no longer runs teardown on a raw timer thread when the camera handler is gone; coordinator device/session fields are **volatile**.
> 
> **Video open failures** now report via the video error path (including MediaRecorder setup failure, which stops instead of continuing into a doomed open). Removes an unused `setupCameraForQueuedRequest` wrapper. Unit tests cover thread posting, await timeout, and keep-alive behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c25a4ca35baee4fa3150748c866242721ca8558. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/AsgConstants.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/AsgConstants.java
@@ -47,6 +47,13 @@ public class AsgConstants {
     public static final long CAMERA_FOV_OVERRIDE_DEFAULT_TTL_MS = 300_000L;
     public static final long CAMERA_FOV_OVERRIDE_MAX_TTL_MS = 600_000L;
 
+    /**
+     * How long a caller evicting the kept-alive camera waits for the camera-thread close
+     * to finish before proceeding. Slightly above closeCamera's own 5s open/close-lock
+     * timeout so the wait can only expire after the close itself gave up.
+     */
+    public static final long CAMERA_TEARDOWN_AWAIT_MS = 6_000L;
+
     public static String appName = "AugmentOS ASG Client";
     public static int augmentOsSdkVerion = 1;
     public static int asgServiceNotificationId = 3540;
--- a/asg_client/app/src/main/java/com/mentra/asg_client/camera/CameraNeoService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/camera/CameraNeoService.java
@@ -1492,6 +1492,10 @@ public class CameraNeoService extends LifecycleService {
                     notifyVideoError(
                             videoSession.currentVideoId(),
                             "Failed to set up video recorder: " + ioe.getMessage());
+                    // Without the recorder surface the open can only fail again later
+                    // with a second, more confusing error — stop here.
+                    conditionalStopSelf();
+                    return;
                 }
             } else {
                 // For photos, find the closest available JPEG size to our target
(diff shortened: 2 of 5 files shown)


PhilippeFerreiraDeSousa:
Shelved pending bench verification on hardware — architectural hardening, not required to close OS-1816 (#3596 stops the crash on its own). Known open finding to address before un-drafting: the warm-teardown re-check does not cover statically pending warm callbacks (sPendingWarmCallbacks).

PR #3598, opened 2026-07-25
Add camera lifecycle generation guard for stale Camera2 callbacks (OS-1816 follow-up)

PhilippeFerreiraDeSousa:
## Scope

Third PR of the OS-1816 series, stacked on #3597. The teardown-detection in the camera's async callbacks was a set of null checks — and a null check cannot tell a reopened camera from its own. After a close+reopen, `device`/`session` are non-null again, so a late callback for the OLD camera would mutate the successor's state: `onConfigured` installing a dead session, `onDisconnected` wiping the live device, a stale keep-alive expiry closing a camera it never owned (`Timer.cancel()` cannot dequeue an expiry already posted to the camera thread).

## Changes

`CameraCoordinator` now tracks:

- a coarse **lifecycle state** (`CLOSED → OPENING → OPENED → CONFIGURED`), for logging/diagnostics and as scaffolding for further confinement work;
- an authoritative **generation counter**: `beginOpen()` bumps it for every open attempt, `closeDeviceAndSession()` for every teardown.

Async work captures the generation it was issued under and drops itself when superseded:

- `onOpened` closes the surplus device instead of installing it — this also recovers cleanly when a 5s proceed-anyway teardown retired the open while the HAL was still working on it (previously that path re-leaked into exactly the abandoned-surface crash).
- `onDisconnected` / `onError` skip `clearDevice`/notify/`stopSelf` when stale.
- `onConfigured` / `onConfigureFailed` close the session and return when stale.
- keep-alive expiry skips when the camera it was armed for is gone — this closes the audit finding that an expiry posted before `Timer.cancel()` could tear down the camera of a capture that had just cancelled it.

The pattern follows the existing `captureMetadataGeneration` precedent in `PhotoSession` (`recordStillCaptureMetadata_staleGenerationIgnored`).

## Test evidence

- `./scripts/check-android-compile.sh asg` green
- `./gradlew :app:testDebugUnitTest`: 638 tests, 0 failures
- New `CameraCoordinatorTest` coverage: `beginOpen` bumps generation + `OPENING`; `closeDeviceAndSession` retires the open generation; full state walk `CLOSED→OPENING→OPENED→CONFIGURED→CLOSED`; stale-generation keep-alive expiry is skipped after a reopen.

---

> [!NOTE]
> **High Risk**
> Touches core Camera2 open/close/teardown and async callback ordering; mistakes could regress first-shot failures, leaks, or races already tracked under OS-1816.
> 
> **Overview**
[...]

PR #3599, opened 2026-07-25
Run the camera open path on the camera thread (OS-1816 follow-up)

PhilippeFerreiraDeSousa:
## Scope

Final PR of the OS-1816 series, stacked on #3598. It completes the single-owner-thread model: after #3597 confined teardown and #3598 guarded stale callbacks, the camera **open** machinery (capability queries, ImageReader creation, `openCamera`) was the last lifecycle work still running on arbitrary threads — the BLE command thread, NanoHTTPD web-server workers, or main, depending on who dispatched the photo/video/warm-up.

## Changes

- `openCameraInternal` self-posts to the camera thread. The posted body re-takes `SERVICE_LOCK`, so the open runs under the same lock its dispatch decision was made under — and the video open path, which previously took **no** lock, now behaves like the photo path.
- If an open is already in flight (`LifecycleState.OPENING`), fail fast with a camera-busy error instead of blocking the camera thread on the open/close permit: the in-flight open's callbacks are delivered on this same thread, so waiting could only delay them and then time out anyway.
- `CameraCoordinator.isOnCameraThread()` + test coverage.

With this, every mutation of camera lifecycle state (open, session setup, capture dispatch targets, teardown) happens on the CameraNeoBackground thread; cross-thread entry points only enqueue, evict-and-wait, or read advisory state.

## Scope note

The originally sketched follow-up — extracting the camera controller from the Android `Service` — is deliberately **not** included: its correctness motivation (main-thread teardown racing background setup) is gone after this series, leaving it a large refactor-only diff. Worth scheduling separately if wanted; not needed to close OS-1816.

## Test evidence

- `./scripts/check-android-compile.sh asg` green
- `./gradlew :app:testDebugUnitTest`: 639 tests, 0 failures

---

> [!NOTE]
> **Medium Risk**
> Touches core camera open/teardown and video stop threading where OS-1816 races were fixed; behavior changes for overlapping opens (busy error) and off-thread entry points now async-post to the camera thread.
> 
> **Overview**
> Finishes the OS-1816 **single camera-thread owner** model by moving the last lifecycle work that still ran on BLE, HTTP workers, or main onto `CameraNeoBackground`.
> 
> **`openCameraInternal`** now posts to the camera thread and re-takes `SERVICE_LOCK` there so open/session mutations stay under the same lock as dispatch; the real HAL work lives in **`openCameraOnCameraThread`**. If state is already **`OPENING`**, it **fails fast** with a camera-busy error instead of blocking on the open/close permit (and it does **not** stop the service while the in-flight open still owns it). The video open path now follows the same locking pattern as photo.
> 
> **Video stop** (`ACTION_STOP_VIDEO_RECORDING`) and **`onDestroy`** when recording both call **`videoSession.stopRecording`** on the camera thread so MediaRecorder callbacks cannot race the lock-free **`VideoRecordingSession`**; destroy still posts **stop then close** on that thread before the background thread is drained.
> 
> Adds a unit test that **`CameraCoordinator.isOnCameraThread()`** reflects the calling thread.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 197a79429e5ccbfb7069065dd5b53fd1a8b5d3f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/camera/CameraNeoService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/camera/CameraNeoService.java
@@ -1176,7 +1176,11 @@ public class CameraNeoService extends LifecycleService {
                     }
                 case ACTION_STOP_VIDEO_RECORDING:
                     String videoIdToStop = intent.getStringExtra(EXTRA_VIDEO_ID);
-                    videoSession.stopRecording(videoIdToStop);
+                    // The recorder is created on the camera thread (its onInfo/onError
+                    // listeners fire there); stopping from main would race them inside
+                    // the lock-free VideoRecordingSession. Stop on the camera thread.
+                    cameraCoordinator.runOnCameraThread(
+                            () -> videoSession.stopRecording(videoIdToStop));
                     SystemControllerFactory.get(this).setEisEnabled(false);
                     break;
                 case ACTION_WARM_UP_CAMERA:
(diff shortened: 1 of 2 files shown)


PhilippeFerreiraDeSousa:
Shelved pending bench verification on hardware. Known open findings to address before un-drafting: the OPENING fail-fast can route its error to the in-flight request's callback rather than the rejected newcomer, and the async video stop can reject an immediate re-start with 'Already recording'.

PR #3604, opened 2026-07-26
Deliver already-saved photos across camera device loss (OS-1816 follow-up)

PhilippeFerreiraDeSousa:
## Scope

Fifth PR of the OS-1816 series, stacked on #3599. Driven by the reporter's adb logcat from the #3596 build: the process crash is gone, but the first capture of a session still fails. The log shows why — **the photo is captured and saved to disk**, then the MediaTek camera service disconnects the client (`checkPidStatus:2264` eviction) 4ms later, while the capture is only waiting for HAL metadata:

```
19:44:18.827 CameraNeo: Photo saved successfully: .../base.jpg
19:44:18.831 CameraCaptureSession: CAMERA_DISCONNECTED (2): checkPidStatus:2264
19:44:18.835 CameraNeo: CameraNeoService service destroying
19:44:19.383 MediaCaptureService: Failed to capture photo: Camera service terminated unexpectedly
```

Our `onDisconnected` handler failed the request and stopped the service; `onDestroy`'s `failAllPending` error is delivered inline and can outrun the success callback queued on the executor — so a photo sitting on disk was reported as `CAMERA_CAPTURE_FAILED` and its files deleted.

## Changes

- `PhotoSession.flushPendingCapturedPhotoNow(reason)`: emits the pending saved photo immediately without metadata, reusing the exact emit path the existing metadata-wait timeout uses. Idempotent via `photoCapturedCallbackSent`.
- `onDisconnected`/`onError` (current-generation, photo path) now salvage instead of fail: close local camera state first (so a queued request dispatched by the completion cold-opens cleanly), flush the saved photo, and route `stopSelf` through the callback executor so the delivery drains before `onDestroy` can run — skipping the stop entirely if a new request took over.
- A disconnect that lands **before** the image reaches the reader still fails the request exactly as before.

This does not address *why* the MTK stack evicts the client after first-in-session captures (still under investigation with the reporter — full unfiltered logcat with `cameraserver` lines requested); it makes the eviction harmless to the user.

## Test evidence

- `./scripts/check-android-compile.sh asg` green
- `./gradlew :app:testDebugUnitTest`: 641 tests, 0 failures
- New `PhotoSessionTest` coverage: flush emits the saved photo without metadata exactly once; flush without a pending photo returns false.

---

> [!NOTE]
> **Medium Risk**
> Touches camera disconnect/error teardown and callback ordering in `CameraNeoService`; behavior change is scoped to photo salvage when a file is already saved, with explicit queue-dispatch handling to avoid stuck requests.
> 
> **Overview**
> When the camera HAL disconnects or errors **after** the JPEG is on disk but **before** HAL metadata arrives (seen on MediaTek after first-in-session capture), photo `onDisconnected` / `onError` no longer fail the shot and tear down the service immediately.
> 
> **`PhotoSession.flushPendingCapturedPhotoNow`** emits the pending saved file through the normal capture callback with `null` metadata, guarded by `photoCapturedCallbackSent` so it only fires once.
> 
> **`CameraNeoService.salvageCapturedPhotoAfterDeviceLoss`** runs on those photo paths: closes local camera state, flushes if a pending save exists, then on the single-thread `executor` either **`stopSelf`** (so success can run before `onDestroy`’s `failAllPending`) or **`dispatchNextPhotoRequest`** if more work is queued. Disconnects with no saved image still error as before; video is unchanged.
> 
> Unit tests cover flush success, idempotency, and the no-pending-photo case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1b708da2f84643d2849cc36ddb6c902890824b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/camera/CameraNeoService.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/camera/CameraNeoService.java
@@ -1718,6 +1718,8 @@ public class CameraNeoService extends LifecycleService {
                 cameraCoordinator.clearDevice();
                 if (forVideo) {
                     notifyVideoError(videoSession.currentVideoId(), "Camera disconnected");
+                } else if (salvageCapturedPhotoAfterDeviceLoss("camera disconnected")) {
+                    return;
                 } else {
                     photoSession.notifyHostPhotoError("Camera disconnected");
                 }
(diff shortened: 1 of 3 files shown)


PhilippeFerreiraDeSousa:
Shelved. The salvage concept stands, but the tip contains a confirmed double-dispatch bug (the deferred executor kick re-dispatches a request the flush's own completion path already dispatched — see Bugbot finding above). The intended rework is a simplification: salvage should be closeCamera() + flushPendingCapturedPhotoNow() only, deleting the custom stop/dispatch block entirely and letting the normal post-capture lifecycle (dispatch-or-keep-alive → expiry stop) handle the rest. Do not merge the current tip.

PR #3620, opened 2026-07-28
Orchestrate hotspot-served OTA sessions on the phone

PhilippeFerreiraDeSousa:
## Scope

Fourth PR for OS-1676, stacked on #3618. Plan: `agents/os-1676-hotspot-ota-plan.md`.

`HotspotOtaTransport` ties the stack together: `engine.ota.hotspotSession.begin(checkResult)` downloads the artifacts (#3618), asks the glasses for their hotspot, joins it via the scoped local-network transport, verifies the glasses answer over the link (`/api/health` on the camera web server), and serves the rewritten manifest via the LocalOtaServer (#3617).

**Coordinator integration is one seam**: ota_start URL resolution in `sendOtaStartWithWatchdogs`. While a hotspot session is active, every send site (initial mount, retry, query fallback, post-APK re-arbitration) resolves through `ensureSession()`, which transparently re-establishes the whole link when it is down — which it always is after the APK step: the install kills the asg process owning the `LocalOnlyHotspot` reservation and credentials rotate on restart (hardware-verified in the Phase 0 spike). Watchdogs arm only after the link is up so bring-up time doesn't eat into them; bring-up failures land in the coordinator's existing retry path (#3614's glasses-side resume probe covers the same window from the other end).

Details:
- **Keep-alive**: health polling every 30s holds the hotspot open — the glasses' idle monitor stops it after 120s without inbound HTTP, and during phone-served OTA the glasses are the HTTP *client*. Works on every field build, no glasses change needed.
- **Single-tenant hotspot**: gallery sync refuses to start while an OTA session owns the link.
- **Lifecycle**: `coordinator.finish()` ends the session and deletes the staged artifacts; `detach()` (screen abandoned mid-session) keeps them so a retry skips the re-download.
- **iOS**: joins verify by SSID polling (NEHotspotConfiguration resolves on acceptance, not association) before the glasses health probe.
- `MentraOtaServer` exported through `@mentra/bluetooth-sdk/internal` alongside `MentraLocalNetwork`.

## Test evidence

- 6 unit tests: full bring-up (incl. the two-step server start that feeds the base URL into the manifest rewrite), join-failure artifact retention, concurrent-session refusal, healthy-link short-circuit, **post-APK re-establish with rotated credentials**, and teardown.
- Engine `tsc --noEmit` clean.
[...]

PR #3621, opened 2026-07-28
Offer the hotspot-served install when glasses have no WiFi

PhilippeFerreiraDeSousa:
## Scope

Fifth and final PR of the OS-1676 stack (on #3620), completing the user flow from the ticket: *"There is an update ready for your Mentra Live. Please press Join."*

When the glasses are not on WiFi, **Update Now no longer dead-ends into WiFi setup**. The check screen starts a hotspot-served session instead: it downloads the update to the phone over the phone's own connection (the tap is the consent — no background pre-download, per the plan decision), then brings up the glasses hotspot (`engine.ota.hotspotSession.begin`, #3620) and hands off to the existing `/ota/progress` screen, whose install coordinator drives `ota_start` through the session automatically — including the post-APK-restart re-join.

- `check-for-updates`: new `preparing` state — phone-download percent, then the hotspot-join phase with a heads-up that the OS may show a join/connect dialog. Failure returns to the update prompt (staged artifacts kept, so retrying skips the re-download) and offers WiFi setup as the alternative.
- Update-available copy explains the phone-transfer path; **Setup WiFi remains only as the fallback** when there is no manifest to serve.
- `OtaUpdateChecker`: both background prompts (return-home and main-check) now offer Install without the glasses-WiFi gate; the WiFi-setup-only alert is gone. The glasses-WiFi path itself is untouched and remains primary.

## Test evidence

- Mobile `tsc --noEmit` clean; full mobile jest suite green (including the 25 `OtaUpdateChecker` tests).
- End-to-end on-device verification of the full stack (apk+mtk+bes over hotspot, two hotspot windows) is the next step now that the flow is drivable from the UI — tracked as the stack's hardware gate before any of it merges.

---

> [!NOTE]
> **Medium Risk**
> Changes core OTA user flows and transport selection when glasses lack WiFi; regressions could block updates or mis-route users, though WiFi path is preserved and manifest gating limits the new path.
> 
> **Overview**
> **When glasses are not on WiFi, OTA no longer forces WiFi setup as the only path.** Users with a servable manifest can install via the phone hotspot flow (OS-1676).
> 
> On **`check-for-updates`**, **Update Now** starts `engine.ota.hotspotSession.begin` when there is no glasses WiFi but `manifestBody` exists: a new **`preparing`** state shows phone download progress, then hotspot join messaging, then navigation to `/ota/progress`. Failures return to the update prompt with an alert and optional WiFi setup. Copy and the primary button switch to **Update Now** / hotspot explanation when a manifest is available; **Setup WiFi** remains only without a manifest.
> 
> **`OtaUpdateChecker`** removes the glasses-WiFi requirement for background and return-home update prompts—both now show **Install** (same as on WiFi) instead of a WiFi-setup-only alert. WiFi-connected behavior is unchanged.
> 
> New **`ota:*`** strings cover hotspot transfer, preparing phases, and prepare-failure messaging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac50ca1840a03f8606db48fa233810fd94576c1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/src/app/ota/check-for-updates.tsx
+++ b/mobile/src/app/ota/check-for-updates.tsx
@@ -10,9 +10,12 @@ import {useAppTheme} from "@/contexts/ThemeContext"
 import {useEngineSnapshot} from "@/hooks/useEngineSnapshot"
 import {translate} from "@/i18n/translate"
 import {useNavigationStore} from "@/stores/navigation"
+import {showAlert} from "@/utils/AlertUtils"
 import {getNextOnboardingRoute} from "@/utils/onboarding/getNextOnboardingRoute"
 
-type CheckState = "checking" | "update_available" | "no_update" | "dev_build" | "error"
+type CheckState = "checking" | "update_available" | "preparing" | "no_update" | "dev_build" | "error"
+
+type CheckResult = Awaited<ReturnType<typeof engine.ota.checkForUpdates>>
 
 export default function OtaCheckForUpdatesScreen() {
   const {theme} = useAppTheme()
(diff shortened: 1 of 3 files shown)


PhilippeFerreiraDeSousa:
Superseded by #3727, which implements and hardware-verifies the approved simplified hotspot OTA design. Closing this original stacked PR to avoid maintaining two competing implementations.

PR #3643, opened 2026-07-30
Read Mentra Live serial from current BES version reply

PhilippeFerreiraDeSousa:
## Summary

- Cache `serial_number` from the current BES `sr_syvr` response path.
- Reuse the existing serial validation, storage, and `version_info_3` forwarding behavior.
- Correct the Mentra Live protocol documentation to distinguish current `sr_syvr` from legacy factory `hs_syvr`.

## Why

Current ASG clients consume `sr_syvr` directly in `K900BluetoothManager`, before `McuEventParser`. The previous implementation only parsed `hs_syvr`, so the current response path discarded the glasses identity even when BES supplied it.

## Analytics behavior

This PR does not add or rename analytics events or properties. The existing `bluetooth_sdk_glasses_identified` event still fires once per connection only after a valid manufacturing serial reaches the Bluetooth SDK. Empty and all-zero values remain omitted.

## Validation

- On physical hardware, confirmed the current protocol is `cs_syvr` -> `sr_syvr`.
- Installed the patched ASG and successfully OTA-updated the paired BES firmware to 26.7.30.5.
- Captured `sr_syvr` carrying `serial_number: "0000000000000000"`; the ASG correctly rejected it and omitted it from `version_info_3`. This unit therefore cannot emit `bluetooth_sdk_glasses_identified` until its BES manufacturing serial is provisioned.

## Dependency

- BES firmware: https://github.com/fengyue120/mentra-live-bes/pull/27

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/bluetooth/managers/K900BluetoothManager.java
@@ -1203,7 +1203,7 @@ public class K900BluetoothManager extends BaseBluetoothManager implements Serial
             }
 
             cacheBesBaudSwitchVersion(bData);
-            cacheBesVersionFromSyvrBField(bData);
+            cacheBesMetadataFromSyvrBField(bData);
 
             if (result == BesUartTransportCoordinator.SystemVersionResult.READY) {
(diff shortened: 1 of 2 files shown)


PhilippeFerreiraDeSousa:
Closing because Mentra Live manufacturing serial provisioning is being addressed separately. We should not ship this telemetry path based on the current all-zero device value.

PR #3668, opened 2026-08-03
Harden BES OTA rescue and recovery UI

PhilippeFerreiraDeSousa:
## Summary

Release A is an ASG + Mentra App safety release. It does not change OTA pinning, staging workflow, or firmware manifests.

- keep every legacy v1 K900 chunk within the fixed 240-byte packed envelope while preserving the existing pacing
- send terminal OTA success/failure as one compact reliable notification and retain verbose diagnostics on the glasses
- fail closed before `mh_ota` unless the BES artifact passes immutable identity, compressed/container/decompressed, product, flash-limit, and embedded-version checks
- atomically persist one authorization reservation, bound to a stable Linux boot id and exact target version, before the first `mh_ota` byte
- make authorization recovery monotonic: framed authorization -> raw read-only recovery -> transfer or durable UART quarantine
- restore quarantine after serial close/reopen or ASG process restart, and release it only after a full glasses reboot changes the boot id
- report BES success only after rebooted BES returns the exact target version in a fresh `sr_syvr`
- reuse the compact `install_failed` phone event; the Mentra App shows restart instructions and does not offer an immediate Retry after BES install failure

## Authorization ambiguity

`mh_ota` is a mode transition without a transaction id. A false write result or missing framed `hm_ota` does not prove whether BES stayed in normal K900 mode or entered raw OTA mode. Sending another framed command from that state is unsafe.

The flow is deliberately one-way:

1. Reserve the current glasses boot atomically and send framed `mh_ota` exactly once.
2. If `hm_ota` arrives, promote to raw OTA and begin the protocol.
3. If `hm_ota` is missing, promote once to raw routing and send the read-only, idempotent `0x99` protocol query up to three times.
4. A raw `0x9A` proves BES is in OTA mode and continues with `SetUser`, transfer, CRC verification, and apply.
5. If all raw probes are silent, do not switch back to framed routing, send `cs_syvr`, or resend `mh_ota`. Reject both parsers and all UART writes until a full glasses reboot.

The five raw probe bytes contain no K900 `##` marker, so the deployed normal parser discards them as junk. The durable reservation is also the startup quarantine record, so an app/serial restart cannot accidentally resume framed traffic into a potentially raw BES parser.

## Completion semantics

The `0x93` apply acknowledgement means BES accepted apply; it is not terminal success.

Before releasing raw OTA ownership, ASG persists a post-apply verification phase and the exact target from the admitted artifact. It then waits for BES reboot, reopens at the rendezvous baud, and permits only system-version discovery. `FINISHED` is emitted only when a current-session `sr_syvr` exactly matches the target. A mismatch, 60-second timeout, or durable-state write failure emits `install_failed`. Pending verification survives an ASG process restart, including the listener-registration race.

## Root cause and release behavior

`wire_caps.notify_cap` can reflect a higher-capacity EATT bearer without guaranteeing that every outbound notification uses that bearer. ASG raised its v1 chunk budget to that advertised size, so terminal OTA frames could be truncated on a lower-capacity bearer and the phone reported failure even when installation succeeded.

The existing staging pinning and manifest generation are unchanged. Current legacy BES manifest metadata is intentionally not admissible by this ASG, so it cannot authorize a BES update until Release B supplies the complete artifact metadata.

## Validation

Passed locally on the pushed head:

- `./gradlew :app:testDebugUnitTest`
- focused authorization, durable quarantine, post-apply verification, UART coordinator, watchdog, terminal-message, and artifact-validator tests
- `git diff --check`
- existing focused mobile recovery-policy tests from this PR

Known baseline limitation:

- `:app:spotlessJavaCheck` reports pre-existing whole-file formatting drift in `OtaHelper.java`; applying it would add roughly 1,300 unrelated changed lines, so this PR keeps only the required target-version call-site edit

---

> [!NOTE]
> **High Risk**
> Changes BES firmware install path, bootloader size limits, and UART quarantine behavior where mis-handling can brick devices or strand users until reboot; extensive new durable-state logic increases race and persistence failure surface.
> 
> **Overview**
> **Release A hardens Mentra Live BES OTA on the ASG** so authorization, transfer, and completion are fail-closed and survive process/serial restarts, without changing staging pinning or manifest generation in this PR.
> 
> **Pre-flight and wire safety:** Downloads must pass a new **LZMA chunk container** validator (format, product, size limits, embedded version) before `mh_ota`. Legacy v1 K900 string chunks stay capped at **240 bytes**; dynamic sizing from `wire_caps.notify_cap` is removed so terminal OTA notifications are not truncated on lower-capacity BLE bearers. Terminal phone status uses a **compact reliable payload** with full diagnostics logged on-glasses.
> 
> **Authorization and UART:** Exactly **one** `mh_ota` per glasses boot is reserved atomically (Linux boot id + exact target version + OTA session id). Missing `hm_ota` triggers read-only **0x99** recovery probes instead of resending `mh_ota`; silent failure **quarantines** UART until reboot. Apply success waits for reboot and an exact **`sr_syvr`** match—not the apply ACK alone—with durable handoff replayed to `OtaService`/phone on reconnect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5234209aec4e3ec9cc3d7ca8c6c95d4acf6123d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PhilippeFerreiraDeSousa:
@codex review

aisraelov (changes requested):
Blocking: the new post-apply verifier does not actually gate the phone-visible BES success state. mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/MentraLive.kt still maps sr_adota progress >= 100 or type=success to complete/step_complete before ASG reads the rebooted sr_syvr. Meanwhile OtaService.onBesOtaProgress only logs FINISHED and never sends the verified terminal ota_status.

The failure correction path is not reliable either. On a version mismatch, BesOtaAuthorizationGate is cleared and BesUartTransportCoordinator.onSystemVersion immediately starts the fast-baud transition before the FAILED EventBus subscriber queues its terminal message, so runNormalWrite rejects it while state is SWITCH_REQUESTED/WAITING_FAST_REOPEN. On verification timeout or persistence failure, the coordinator is quarantined before FAILED is posted, so every UART send is rejected. ReliableMessageManager removes a message after a failed physical send and does not retry when the UART later becomes ready.

That leaves a concrete false-success case: sr_adota makes the phone session terminal, the BLE reconnect edge advances the UI to complete, then exact target verification fails or times out but its corrective failure never reaches the phone. The phone watchdog has already stopped because sr_adota was terminal.

Please make phone-visible BES success originate only from the verified post-reboot callback, and persist/defer the verified terminal result until the UART is safe and ready to deliver it. Alternatively, treat sr_adota 100 as nonterminal until a separate verified terminal status arrives. Add an integration test covering sr_adota success followed by version mismatch and verification timeout.

aisraelov:
Follow-up/correction after confirming the hardware behavior: BES OTA success hard-power-cycles the whole glasses, so the primary blocker is even earlier than the delivery race described in my review.

BesOtaAuthorizationGate.isPostApplyVerificationPendingForCurrentBoot() requires currentBootId == attemptedBootId (lines 114-120). The apply-pending record is written before the OTA-triggered hard power cycle, so after boot the Linux boot ID is necessarily different. BesOtaManager construction therefore does not restore awaitingPostApplyVerification, and verifyPostApplyVersion returns NOT_PENDING for the first fresh sr_syvr. The documented exact post-reboot target verification never runs on the real success path.

At the same time, the phone still maps the pre-power-cycle sr_adota success/100% event to complete or step_complete and completes it on the reconnect edge. So this detail does not remove the blocker; it makes the false-success path deterministic with the current boot-ID predicate.

The pending apply phase needs semantics distinct from the same-boot authorization quarantine: authorization ambiguity is scoped to the attempted boot, while apply verification must be accepted exactly once on the immediately following boot and must bind to the persisted target. Tests should simulate boot-a reservation/apply-pending, boot-b startup, fresh sr_syvr match/mismatch, and terminal phone behavior.

aisraelov (changes requested):
Blocking: the verified boot-B result can still be lost before `OtaService` subscribes.

On a cold glasses boot, `ServiceLifecycleManager.initialize()` fully initializes the Bluetooth/K900/BES managers first, then only schedules `OtaService` to start 5 seconds later (`ServiceLifecycleManager.java:56-60,124-135`). The post-apply UART path starts probing after 2.5 seconds. If the fresh `sr_syvr` arrives in that normal window, `K900BluetoothManager` durably consumes/clears the verification record, `BesOtaManager.onBesPostApplyVerification()` posts a non-sticky EventBus `FINISHED`/`FAILED`, but `OtaService` has not yet registered (`OtaService.java:62-65`). `pendingBesOtaVerification` only bridges the earlier K900-listener race; that listener is registered during Bluetooth initialization and immediately turns the result into the lossy EventBus post.

The result is that a correctly updated device can clear its one verification record while the persisted phone OTA session remains `in_progress` with no later source for the terminal snapshot/resends. This is especially likely after the confirmed whole-glasses hard power cycle.

Please make the verified terminal outcome durable independently of `OtaService` registration (or establish ordering plus reliable replay), and add a startup-order test where boot-B `sr_syvr` is handled before the OTA consumer exists, then the consumer starts and the phone receives the terminal status.

The prior boot-ID predicate and direct `sr_adota` false-success issues are fixed in `986771c`; this remaining delivery race is why I cannot approve the updated head yet.

PhilippeFerreiraDeSousa:
Closed, superseded by #3675

PR #3685, opened 2026-08-05
Recover completed BES updates after reboot

PhilippeFerreiraDeSousa:
## Summary

- Keep the OTA progress UI as one React screen while selecting a sticky `legacy` or `unified` coordinator policy at flow attach.
- Preserve the old legacy route's BES reboot invariant: once BES install reaches 100%, an observed disconnect/reconnect cannot be overwritten by stale pre-reboot progress.
- Carry the manifest's expected BES version into the install session and complete only after post-reboot `version_info` matches it. Frozen legacy flows without target metadata retain the historical disconnect/reconnect fallback.
- Keep early BES-install link drops on the normal query/resume path, and share all behavior between iOS and Android through the engine coordinator.

## Root cause

Before the progress routes were merged, `check-for-updates.tsx` selected `progress-legacy.tsx` once for builds below 37. That component remained mounted after an APK step upgraded ASG and completed BES locally after an observed restart disconnect/reconnect.

The merged coordinator instead allowed the latest OTA event shape to change the compatibility policy. After BES reboot, the store still contained the pre-reboot `in_progress` state at 100%, so reconnect derived `updating` again and eventually fired the progress timeout despite the firmware having installed successfully.

## Device evidence

Captured from `iphone-syslog-ota-verification-20260805-123944.txt` during the staging iPhone + Mentra Live validation:

```text
1497760: Aug  5 12:47:03.910102 Mentra[17008] <Info>: BLE_TRACE direction=glasses_to_phone layer=sdk_ble_event source=processJsonObject(_:)(MentraBluetoothSDK/MentraLive.swift:2416) type=version_info_3 payload={"type":"version_info_3","bt_mac_address":"","bes_fw_version":"17.26.7.9","mtk_fw_version":"MentraLive_20260709"}
1500500: Aug  5 12:47:04.548495 Mentra(React)[17008] <Info>: '[OTA_PROGRESS] displayState restarting -> updating', '{"connected":true,"sawReconnectEdge":false,"errorMsg":null,"otaStatus":{"stepType":"bes","phase":"install","status":"in_progress","step":"2/2","pct":100},"otaProgress":{"currentUpdate":"bes","stage":"install","status":"PROGRESS","progress":100}}'
1500510: Aug  5 12:47:04.549036 Mentra(React)[17008] <Info>: [OTA_PROGRESS] connect-edge: false->true, flipping sawReconnectEdge=true
1500536: Aug  5 12:47:04.550195 Mentra(React)[17008] <Info>: [OTA_PROGRESS] connect-edge: reconnected, sending ota_query_status
1501155: Aug  5 12:47:04.605477 Mentra(React)[17008] <Info>: 'CORE:', 'LIVE: Sending data to glasses: {"mId":74,"type":"ota_query_status","timestamp":1785959224553}'
1505133: Aug  5 12:47:08.108642 Mentra[17008] <Info>: BLE_TRACE direction=glasses_to_phone layer=sdk_ble_event source=processJsonObject(_:)(MentraBluetoothSDK/MentraLive.swift:2416) type=msg_ack payload={"type":"msg_ack","mId":74,"timestamp":1785959224952}
```

The glasses reported the installed target BES version before the mobile UI moved from `restarting` back to stale `updating`. The query received a delivery ACK for message 74 but no fresh `ota_status`, leaving the stale 100% state in control.

## Validation

- `bun run test -- --runInBand --silent src/__tests__/otaInstallCoordinator.test.ts src/app/ota/__tests__/deriveOtaDisplayState.test.ts src/__tests__/otaUpdateCheckService.test.ts src/app/ota/__tests__/progress.test.tsx` — 134 tests passed.
- `bun run compile` — passed.
- `git diff --check` — passed.

Regression coverage includes:

- day-one build 27 upgrading to ASG 39 before the BES reboot;
- stale unified BES `in_progress` at 100% after reconnect;
- BES target version arriving before and after the reconnect edge;
- legacy fallback without target metadata;
- an early BES disconnect at 50% staying on the normal query/resume path;
- later link noise not reopening a completed recovery;
- zero-padded manifest BES versions matching unpadded `version_info`;
- Continue remaining disabled while BES target verification is pending;
- Retry clearing stale BES reboot-attempt latches.

---

> [!NOTE]
> **Medium Risk**
> Changes core OTA install completion and reconnect arbitration in `OtaInstallCoordinator`; behavior is heavily tested but mistakes could mark success too early or block legitimate retries.
> 
> **Overview**
> **Fixes BES firmware updates that actually succeeded but the progress UI treated as stuck or failed** because stale `in_progress` at 100% came back after reconnect and `ota_query_status` never refreshed session state.
> 
> The install coordinator now **locks `legacy` vs `unified` OTA behavior at attach** (matching the old split routes when an APK step upgrades ASG mid-flow), and adds a **phone-observed BES reboot path**: after BES install hits 100%, disconnect/reconnect is tracked separately from store OTA status. On reconnect it **skips `ota_query_status`**, polls **`version_info`**, and marks complete only when **`besFirmwareVersion` matches the manifest `besVersion`** (with semantic version compare); legacy sessions without target metadata still complete on the historical reboot edge. **Continue stays disabled** until verification finishes or times out (~1 minute).
> 
> `OtaUpdateInfo` and the update check now **surface `besVersion` from the manifest** when a BES step is pending. `deriveDisplayState` **prioritizes `besRestartRecovery`** so stale progress cannot flip the UI from `restarting` back to `updating`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7044fe3ddc4c17c0ad3558020f2d4e94c58429a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/modules/bluetooth-sdk/src/BluetoothSdk.types.ts
+++ b/mobile/modules/bluetooth-sdk/src/BluetoothSdk.types.ts
@@ -1201,6 +1201,8 @@ export interface OtaUpdateInfo {
   updates: string[] // ["apk", "mtk", "bes"]
   totalSize: number
   cacheReady?: boolean
+  /** Exact BES target selected from the OTA manifest, when a BES step is pending. */
+  besVersion?: string
   /** True when the APK step installs an older build than the glasses currently run (exact-pin manifests only). */
   isDowngrade?: boolean
 }
--- a/mobile/modules/engine/src/services/OtaInstallCoordinator.ts
+++ b/mobile/modules/engine/src/services/OtaInstallCoordinator.ts
@@ -25,6 +25,7 @@ import type {OtaProgress, OtaStatus} from "@mentra/bluetooth-sdk/internal"
 import GlobalEventEmitter from "../utils/GlobalEventEmitter"
 import {isGlassesConnected, useGlassesStore} from "../stores/glasses"
 import {resolveOtaManifestUrl} from "./otaManifestUrl"
+import {compareVersions} from "./OtaUpdateCheckService"
 import {deriveDisplayState, type DisplayState} from "./otaDisplayState"
 import {
   BES_CONTINUE_LOCKOUT_MS,
@@ -326,11 +327,14 @@ class OtaInstallCoordinator {
       this.clearPerStepTimers()
       this.clearLegacyApkSettleTimer()
       this.clearMtkSimulationTimers()
+      this.clearContinueLockoutTimer()
       this.retryCount = 0
       this.hasFirstActivity = false
       this.hasReceivedAck = false
+      this.resetBesRestartAttempt()
       this.setSawReconnectEdge(false)
       this.setErrorMsg("")
+      this.setContinueButtonDisabled(false)
       this.setApkCompletedViaBuildIncrease(false)
       this.setLegacyApkSettleHold(false)
       this.setMtkSimulatedPercent(null)
@@ -392,7 +396,10 @@ class OtaInstallCoordinator {
         versionChangeSession: this.versionChangeSession,
       }),
       errorMsg: this.errorMsg,
-      continueButtonDisabled: this.continueButtonDisabled,
+      // A timed post-restart lockout may expire while version_info is still
+      // missing or mismatched. Verification is a stronger gate: the user must
+      // not clear the session until the installed BES target is confirmed.
+      continueButtonDisabled: this.continueButtonDisabled || this.besRestartRecovery === "awaiting",
       connected,
       // Copies: the snapshot must not hand callers mutable references into the store.
       otaStatus: otaStatus ? {...otaStatus} : null,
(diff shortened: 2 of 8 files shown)


aisraelov (changes requested):
Two blocking recovery issues remain:

1. BES verification has no bounded failure path. Once the disconnect sets `besRestartRecovery` to `awaiting`, the derived `restarting` state clears the global timeout, the stall watchdog ignores the session, and Continue stays disabled. If `version_info` is missing/mismatched or the one-shot request rejects, the user is stranded indefinitely. I reproduced this by advancing twice `GLOBAL_OTA_TIMEOUT_MS`; the snapshot still had an empty error and remained `restarting`.

2. `compareVersions` is not safe for exact BES identity verification. Its `parseInt` component handling accepts numeric prefixes, so a reported `17.26.7.9-not-the-target` is treated as equal to manifest target `17.26.7.9` and the coordinator completes. BES OTA identity elsewhere requires exactly four numeric byte components; this gate should canonicalize/validate both inputs with those rules before comparing.

The existing focused suite (134 tests) and `bun run compile` pass, but focused regression tests for both cases fail on `ae57ae6`.

PR #3688, opened 2026-08-06
Integrate G2 native notifications with Notify on Android and iOS

aheschl1:
Use G2's native notification centre as Notify's presentation surface on Android and iOS. Native mode replaces the Mentra card, so each captured event still reaches authorized miniapps without a second overlay. Starting/stopping Notify owns presentation on both platforms; native mode remains off by default.

The existing `engine.phoneNotifications` policy now applies the shared controls and Android blocklist. Notify settings expose history/popups, duration, do-not-disturb, permission guidance, and transfer failures. The SDK exposes native configuration/status while keeping iOS full-content relay, iOS app-filter editing, and phone-dismissal synchronization explicitly unsupported. G2 receives ANCS directly from iOS; the SDK reapplies controls after connection/authorization changes; unsupported uploads fail instead of succeeding as no-ops.

Android file transfers use a separate, tested service on the existing BLE write executor. The bounded queue reports delivery/drop/cancellation outcomes, updates retain their IDs for the driver lifetime (including reconnects), and disable/filter changes/disconnect invalidate pending work. The 8,000-ID range is never recycled: once exhausted, new IDs fail explicitly while existing IDs remain usable. Disabled configuration stops presentation without changing the firmware whitelist. ACKs are CRC-checked and tied to the current right GATT connection. An ambiguous timeout or cancelled transfer requires reconnect because the verified wire ACK has no transaction id.

Validation:
- Android public SDK compile check and 153 unit tests passed, including file phase ordering, CRC rejection/corruption, cancellation, reconnect, and id handling.
- Swift package: 23 tests passed; unsigned iPhone-device SDK build passed, including ANCS code.
- Mobile Jest: 828 passed, 2 skipped (98 passing suites); host/UI cases cover native routing without duplicate cards, iOS Notify registration, permission guidance, and shared settings.
- Follow-up regressions: 20 focused mobile tests passed, including stale Android permission refreshes and full-content iOS presentation; macOS reports native notifications as unsupported.
- 10 engine policy tests passed; SDK, engine, and mobile TypeScript checks passed.

Before the latest Android review fixes, remote CI passed on `cb0573c217af2399ac5c5d9b787e80a66975eb9f`: Android and iOS app builds, Swift, OEM boundary, and mobile quality (856 tests passed, 2 skipped).

The latest review fixes pass the Android compile check and 153 unit tests, four settings tests, and mobile TypeScript. Updated remote checks are pending.

Physical G2 qualification and UI screenshots are pending. Local builds and synthetic protocol tests do not prove firmware behavior. The end-to-end device checklist and protocol limits are recorded in `notes/g2-notification-service.md`. Native mode remains disabled by default while physical qualification is pending.


PhilippeFerreiraDeSousa:
> My initial assessment of #3688 is positive on the feature, but not on the separate forwarding path. The G2’s native notification centre adds something the existing display cards cannot provide. However, the proposed bridge bypasses Notify’s presentation control and can show the same notification twice.

It would be best to reuse the existing notification flow and support iOS through ANCS.

PhilippeFerreiraDeSousa:
# PR #3688: G2 native notification centre

Recommendation: keep the feature, but integrate its presentation and settings into the existing notification system before merging. It is a feature addition: populating the firmware-owned notification history and overlay is different from drawing a Mentra notification card.

The feature branch now includes current dev (333e6c36f3), merged as d07af1cf89. The only textual conflict was a camera-comment encoding change. The Android SDK compile check and all six notification-bridge tests pass on the combined source. This does not establish physical G2 behaviour.

## Existing systems to extend

- `mobile/modules/engine/src/services/PhoneNotificationsSync.ts` already owns Android notification capture configuration and the per-app blocklist for every host.
- `mobile/modules/engine/src/facades/phoneNotifications.ts` already exposes notification permissions and filtering to hosts. Extend this facade with presentation settings and platform capability/authorization state.
- `mobile/src/services/MantleManager.ts` already accepts both Android Crust events and iOS Bluetooth relay events, forwards them to subscribed miniapps, and controls wearer-facing cards/speech through Notify. Move only the shared routing/presentation policy into an engine notification service as needed; do not create a second independent subscription that bypasses it.
- The G2 driver already owns native dashboard/menu/calendar/notification protocol primitives. The new Even File Service is a legitimate hardware transport addition. Extract its framing/CRC/ack state from the already-large driver and use the driver's existing BLE scheduling infrastructure.

## Why the current integration needs revision

The new `G2NotificationBridge` forwards every allowed Android notification when its Super Settings flag is enabled, independently of Notify. If Notify is also running, both the existing card and native overlay can appear. Stopping Notify therefore no longer stops every wearer-facing notification. This is a product-policy change, not just a transport addition.

Use one presentation policy with an explicit native-centre output choice for supported glasses. Preserve notification forwarding to miniapps independently of presentation, as the current system does. Decide whether native history should remain populated when popups are disabled; model history delivery and automatic display separately rather than hardcoding both on.

The driver currently hardcodes automatic display on, five seconds, DND avoidance off, and disables the firmware whitelist. Turning the JS flag off only stops future enqueueing; it neither disables the firmware control plane nor invalidates already queued transfers. Dismissal events are not forwarded, although posted/updated notifications reuse a stable numeric ID. These lifecycle behaviours need explicit settings and tests before release.

## Android/iOS parity

| Concern | Android | iOS |
|---|---|---|
| Source | Existing Crust NotificationListenerService | G2 firmware receives notifications from iOS through ANCS |
| Deliver content to native centre | G2 Even File Service JSON upload | Firmware handles ANCS content directly; do not resend the metadata relay as a new notification |
| Shared controls | Enable/history, automatic display, duration, DND, supported app filtering | Same controls through G2 notification control messages, subject to verified firmware support |
| Permission UI | Existing notification-listener permission | ANCS authorization and system notification sharing; expose denied/unavailable states |
| Updates/removals | Stable IDs plus verified removal action; subscribe to existing dismissal events | ANCS has update/removal events; verify the G2 firmware handles them correctly |
| Full content for miniapps | Existing captured title/body | Current G2 relay supplies only app identity, with empty title/body; richer parity requires additional verified firmware/protocol support |

Apple's ANCS model gives notification access to the Bluetooth accessory. It is not an Android-style cross-app notification listener inside the iOS app:
- [Apple ANCS specification](https://developer.apple.com/library/archive/documentation/CoreBluetooth/Reference/AppleNotificationCenterServiceSpecification/Introduction/Introduction.html)
- [CoreBluetooth ANCS connection option](https://developer.apple.com/documentation/corebluetooth/cbconnectperipheraloptionrequiresancs)
- [CoreBluetooth ANCS authorization state](https://developer.apple.com/documentation/corebluetooth/cbperipheral/ancsauthorized)

The current iOS `sendPhoneNotification` no-op only provides a matching method signature. It does not provide functional parity. Existing `G2.swift` already has notification-control builders, a native-panel command, and an inbound NOTIFICATION_IOS handler. Reuse those; the inbound handler explicitly has no title/body. Mentra Live's separate iOS ANCS authorization/relay implementation is a useful lifecycle reference, not a protocol implementation to copy into G2.

## Suggested implementation order

1. Define the shared presentation policy and native-notification capabilities/status on the existing engine facade. Decide how Notify's start/stop maps to native history and popups, including the current absence of Notify from the iOS built-in catalog.
2. Keep the Android file service, but add session-scoped queue cancellation, verified acknowledgement/corruption tests, update/removal handling, and observable delivery failures. Reapply settings after reconnect; disabling must invalidate queued sends and update firmware state.
3. Add iOS G2 ANCS authorization/status and apply the shared firmware controls. Keep Android JSON upload separate from iOS firmware-managed receipt.
4. Add parity UI in notification settings. Do not show unsupported operations as successful no-ops. Map filtering to Android package names versus iOS bundle IDs only where firmware support is verified.
5. Validate physical G2 behaviour on both phones: permission denial/revocation, reconnect, burst traffic, updates/removals, disabled presentation, DND, filtering, and no duplicate overlays. Keep full-content miniapp relay parity a separately scoped capability until demonstrated.

No feature redesign or merge of #3688 has been performed. The conflict-resolution update is pushed for continued review.

PR #3694, opened 2026-08-07
Use xy hotspot for Mentra Live gallery sync

aisraelov:
## Summary

- restore the K900 firmware's SmartXY `ap_start` intent as the sole Mentra Live hotspot path
- wait for the AP gateway and firmware SSID/password before publishing hotspot readiness to the phone
- cancel pending readiness checks and disable the firmware AP on stop or startup failure

## Why

Current Mentra Live firmware rejects or cannot complete Android `LocalOnlyHotspot`, which prevents the phone from joining the glasses for gallery photo and video sync. The existing SmartXY hotspot intent is the device-supported path and avoids the incompatible Android API.

This replaces the closed hybrid approach in #3691 with a staging-based, K900-only change.

## Validation

- `./gradlew :app:testDebugUnitTest --tests 'com.mentra.asg_client.io.network.managers.K900NetworkManagerGatewayTest' :app:assembleDebug`
- installed the debug APK on a Mentra Live over ADB
- verified start reaches `K900 vendor hotspot ready`, `ap0` is up at `192.168.43.1`, and firmware credentials are present
- verified stop returns `ap0` to `DOWN`

---

> [!NOTE]
> **Medium Risk**
> Changes the only K900 hotspot path used for phone gallery sync; incorrect readiness or cleanup could block sync or leave the AP running, though behavior is scoped to K900 and mirrors the supported firmware API.
> 
> **Overview**
> **Mentra Live gallery sync** on K900 now brings up the hotspot through the firmware SmartXY `ap_start` broadcast instead of Android `LocalOnlyHotspot`, which current firmware cannot complete reliably.
> 
> `K900NetworkManager` sends `ap_start` to system UI, polls until `ap0` (or equivalent) has a gateway **and** `Settings.Global` exposes `xy_ssid` / `xy_pwd`, then publishes SSID, password, and gateway to the phone. Stop and failed start cancel readiness polling and send `ap_start` with `enable=false`. Tethering broadcasts stay ignored so readiness is not announced before credentials exist.
> 
> `AsgConstants` adds the SmartXY setting keys; `mentra-live-spec.md` documents the vendor hotspot path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba10bff96c6afb28d92152fac0483eb9b1b52dea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
+++ b/asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
@@ -4,11 +4,11 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
-import android.net.wifi.WifiConfiguration;
 import android.net.wifi.WifiManager;
 import android.os.Handler;
 import android.os.Looper;
 import android.os.SystemClock;
+import android.provider.Settings;
 import android.util.Log;
 import com.mentra.asg_client.AsgConstants;
 import com.mentra.asg_client.io.network.core.BaseNetworkManager;
@@ -39,15 +39,12 @@ public class K900NetworkManager extends BaseNetworkManager {
     private BroadcastReceiver wifiStateReceiver;
     private final boolean isSystemApp;
 
-    private final Handler localHotspotHandler = new Handler(Looper.getMainLooper());
-    private final Object localHotspotLock = new Object();
-    private WifiManager.LocalOnlyHotspotReservation localHotspotReservation;
-    private Runnable pendingLocalHotspotReadiness;
-    private boolean localHotspotStarting;
-    private int localHotspotGeneration;
-    private long localHotspotReadinessDeadlineMs;
-    private String pendingLocalHotspotSsid = "";
-    private String pendingLocalHotspotPassword = "";
+    private final Handler mHotspotHandler = new Handler(Looper.getMainLooper());
+    private final Object mHotspotLock = new Object();
+    private Runnable mPendingHotspotReadiness;
+    private boolean mHotspotStarting;
+    private int mHotspotGeneration;
+    private long mHotspotReadinessDeadlineMs;
 
     /**
      * Create a new K900NetworkManager
@@ -113,8 +110,8 @@ public class K900NetworkManager extends BaseNetworkManager {
 
     @Override
     protected boolean shouldMonitorTetheringBroadcasts() {
-        // LocalOnlyHotspot owns its lifecycle through LocalOnlyHotspotCallback. Treating its
-        // interface as a tethered AP can publish static credentials before the LOHS callback.
+        // The K900 vendor intent owns readiness. A framework tethering echo can arrive before
+        // ap0 and the SmartXY credentials are ready, so do not publish from that broadcast.
         return false;
     }
 
@@ -190,26 +187,25 @@ public class K900NetworkManager extends BaseNetworkManager {
 
     @Override
     public void startHotspot() {
-        Log.d(TAG, "🔥 =========================================");
-        Log.d(TAG, "🔥 START K900 LOCAL-ONLY HOTSPOT");
-        Log.d(TAG, "🔥 =========================================");
-
         final int generation;
-        synchronized (localHotspotLock) {
-            if (isHotspotEnabled || localHotspotStarting || localHotspotReservation != null) {
-                Log.d(TAG, "🔥 Local-only hotspot is already active or starting");
+        synchronized (mHotspotLock) {
+            if (isHotspotEnabled || mHotspotStarting) {
+                Log.d(TAG, "🔥 K900 vendor hotspot is already active or starting");
                 return;
             }
-            localHotspotStarting = true;
-            generation = ++localHotspotGeneration;
+            mHotspotStarting = true;
+            generation = ++mHotspotGeneration;
+            mHotspotReadinessDeadlineMs =
+                    SystemClock.elapsedRealtime()
+                            + AsgConstants.LOCAL_HOTSPOT_READINESS_TIMEOUT_MS;
         }
 
-        requestLocalOnlyHotspot(generation);
+        requestVendorHotspot(generation);
     }
 
-    private void requestLocalOnlyHotspot(int generation) {
-        synchronized (localHotspotLock) {
-            if (generation != localHotspotGeneration || !localHotspotStarting) {
+    private void requestVendorHotspot(int generation) {
+        synchronized (mHotspotLock) {
+            if (generation != mHotspotGeneration || !mHotspotStarting) {
                 Log.d(TAG, "🔥 Ignoring hotspot request from a stale generation");
                 return;
             }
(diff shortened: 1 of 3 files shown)


aisraelov:
@codex review

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
```diff
@@ -190,146 +187,103 @@ public static void enableScan5GWifi(Context context, boolean bEnable) {
...
-                        public void onFailed(int reason) {
-                            failLocalHotspotStartup(
-                                    generation,
-                                    "Local-only hotspot failed with reason " + reason);
-                        }
-                    },
-                    localHotspotHandler);
-            Log.i(TAG, "🔥 Local-only hotspot start requested");
+            sendVendorHotspotState(true);
+            checkVendorHotspotReadiness(generation);
```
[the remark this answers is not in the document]
aisraelov: This is a possible stale-cache edge case, but it assumes an unobserved external shutdown of a healthy SmartXY AP. All supported app stop/failure/inactivity paths already call onHotspotStopped(), and we have no device evidence of spontaneous firmware teardown. Continuous AP monitoring or state validation would add lifecycle complexity outside the reported LocalOnlyHotspot incompatibility, so this PR intentionally stays scoped to restoring the device-supported SmartXY path.

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
```diff
@@ -190,146 +187,103 @@ public static void enableScan5GWifi(Context context, boolean bEnable) {
...
-            if (isHotspotEnabled || localHotspotStarting || localHotspotReservation != null) {
-                Log.d(TAG, "🔥 Local-only hotspot is already active or starting");
+        synchronized (mHotspotLock) {
+            if (isHotspotEnabled || mHotspotStarting) {
+                Log.d(TAG, "🔥 K900 vendor hotspot is already active or starting");
                 return;
             }
-            localHotspotStarting = true;
-            generation = ++localHotspotGeneration;
+            mHotspotStarting = true;
```
PhilippeFerreiraDeSousa, line 196: **[P1] Make an in-flight start cancellable through the command path**

Until readiness succeeds, `mHotspotStarting` is true but `isHotspotEnabled()` remains false. The unchanged `WifiCommandHandler` therefore treats a disable request as “already disabled” and never calls `stopHotspot()`. The queued vendor start continues and can publish an enabled AP after the phone/user was told it was disabled. Expose the transition to command handling (or otherwise route disable to `stopHotspot()` while starting) and cover start→cancel deterministically.

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
```diff
@@ -382,78 +336,57 @@ static boolean isLocalHotspotAddress(String interfaceName, String address) {
...
+            cancelHotspotReadinessLocked();
         }
-        if (reservation != null) {
-            reservation.close();
+        try {
+            sendVendorHotspotState(false);
+        } catch (Exception e) {
+            Log.e(TAG, "🔥 Error stopping K900 vendor hotspot", e);
         }
         onHotspotStopped();
```
PhilippeFerreiraDeSousa, line 391: **[P1] Do not acknowledge stop before SmartXY teardown completes**

`sendBroadcast()` only queues the vendor stop, but this immediately clears public state and lets the phone request another start. While the old `ap0` and `xy_*` settings are still present, the new generation's first readiness poll can accept those stale values and announce success; the queued stop may then take that AP down. Keep a stopping transition, wait with a bound until the physical AP is down, and only then acknowledge/execute a queued restart. Add a stop→start ordering test.

PhilippeFerreiraDeSousa (changes requested):
This is a much better architectural direction than the closed hybrid PR: K900 now has one device-supported hotspot path, and readiness is delayed until both the AP gateway and firmware credentials exist. That should address the observed LocalOnlyHotspot startup failure.

However, the implementation currently treats an asynchronous firmware broadcast like a synchronous lifecycle transition. That leaves two supported command sequences broken (inline comments): a start cannot be cancelled while readiness is pending, and a stop is acknowledged before `ap0` is actually down, allowing a restart to accept stale readiness from the previous generation. The generation counter protects Java callbacks, but it does not serialize the external SmartXY AP state.

So my verdict is: good core design, but incomplete lifecycle handling rather than merge-ready. It likely fixes the original start failure, while creating regressions around cancellation/restart. The resolved stale-state thread points at the same underlying issue: cached app state can diverge from the physical AP. Continuous monitoring may not be necessary, but transitions need observable completion or validation before cached state is authoritative.

Validation I ran at this exact head:
- The PR's Gradle command passes and `:app:assembleDebug` succeeds.
- `./scripts/check-android-compile.sh asg` passes.
- `git diff --check` passes.
- The test report contains only the two unchanged `K900NetworkManagerGatewayTest` cases. The advertised `WifiCommandHandlerHotspotTest` does not exist, so none of these lifecycle sequences is tested.
- CI is green apart from one third-party review still pending.
- No Mentra Live is attached here, so I could not independently repeat the ADB/device claims or verify end-to-end iPhone gallery sync and station-WiFi recovery.

Please model starting/stopping explicitly, serialize stop→start against physical AP teardown, make disable cancel an in-flight start, and add deterministic tests for start→cancel and stop→start ordering.

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
aisraelov: This sequence is not newly broken by this PR: the staging LocalOnlyHotspot path also keeps isHotspotEnabled=false while startup is pending, behind the same WifiCommandHandler short-circuit. In addition, both MentraBluetoothSDK implementations reject a second hotspot operation while the first is pending with request_in_flight, so start→cancel is not currently a supported command sequence at the transport API. I agree it could be modeled in a broader lifecycle redesign, but it is not a regression from restoring SmartXY and is outside this targeted fix.

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
aisraelov: The original SmartXY implementation also cleared/notified state immediately after queuing enable=false, so this is not a regression introduced by this PR. The current generation gate is stricter than the old path for in-process readiness callbacks. A stop→immediate-start physical teardown race is theoretically possible, but we have no report or device repro showing SmartXY leaves the final AP down; adding a queued transition state machine would recreate the complexity this staging fix is intentionally removing. I would keep that lifecycle redesign separate if evidence shows the sequence matters.

aisraelov:
Thanks for catching the validation typo: `WifiCommandHandlerHotspotTest` does not exist, and Gradle silently ignored that filter. I corrected the PR body to list only the real gateway test plus assemble and device validation. I replied inline on the lifecycle points with the staging/historical behavior and the iOS/Android SDK single-flight constraint; I do not think those hypothetical rapid-toggle sequences justify expanding this targeted SmartXY restoration into another state machine.

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
```diff
@@ -190,146 +187,103 @@ public static void enableScan5GWifi(Context context, boolean bEnable) {
...
-                        @Override
-                        public void onFailed(int reason) {
-                            failLocalHotspotStartup(
-                                    generation,
-                                    "Local-only hotspot failed with reason " + reason);
-                        }
-                    },
-                    localHotspotHandler);
-            Log.i(TAG, "🔥 Local-only hotspot start requested");
+            sendVendorHotspotState(true);
```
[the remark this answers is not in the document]
aisraelov: Fixed in ba10bff96c. Vendor enable/disable broadcasts now execute under mHotspotLock after generation validation, so their order matches the accepted lifecycle order and a stale start cannot send after stop.

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
```diff
@@ -190,146 +187,103 @@ public static void enableScan5GWifi(Context context, boolean bEnable) {
...
             if (!wifiManager.isWifiEnabled()) {
-                Log.i(TAG, "🔥 WiFi radio is off; enabling it before LocalOnlyHotspot startup");
+                Log.i(TAG, "🔥 WiFi radio is off; enabling it before K900 hotspot startup");
                 if (!wifiManager.setWifiEnabled(true)) {
-                    failLocalHotspotStartup(generation, "Failed to enable WiFi for local hotspot");
+                    failVendorHotspotStartup(generation, "Failed to enable WiFi for hotspot");
                     return;
                 }
-                localHotspotHandler.postDelayed(
+                mHotspotHandler.postDelayed(
```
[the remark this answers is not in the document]
aisraelov: Fixed in ba10bff96c. The hotspot handler is dedicated to this lifecycle, so stop/failure now removes all queued callbacks (including the Wi-Fi-enable delay) and clears the readiness reference.

aisraelov:
@codex review

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
PhilippeFerreiraDeSousa: Rereviewed at `ba10bff96c`. I agree this is not a new regression in this ASG patch: both native SDKs single-flight hotspot operations, so a second command is rejected before reaching `WifiCommandHandler`. The current gallery `cancelSync()` does attempt that disable and can leave the AP running after `request_in_flight`, but that is an existing phone/SDK lifecycle bug and should be handled separately. Withdrawing this as a blocker for this PR.

asg_client/app/src/main/java/com/mentra/asg_client/io/network/managers/K900NetworkManager.java
PhilippeFerreiraDeSousa: Rereviewed at `ba10bff96c`. The new lock fixes the narrower stale-Java-callback ordering bug and correctly cancels the delayed Wi-Fi-enable runnable. It does not resolve this finding as written: `stopHotspot()` still calls `onHotspotStopped()` immediately after merely enqueueing `ap_start=false`, so the public disable resolves before `ap0` is physically down. A sequential enable can then read the previous `ap0` and `xy_*` values and announce readiness before SystemUI has completed teardown/restart. Locking `sendBroadcast()` orders app-side submission, not physical firmware transition. Please gate restart/readiness on observable AP teardown, or provide device evidence that immediate stop→start plus the phone connection delay is reliable.

PhilippeFerreiraDeSousa (changes requested):
Rereviewed at `ba10bff96c`.

The follow-up is a real improvement: generation validation and vendor-command submission now share the lock, and stop/failure cancels the delayed Wi-Fi-enable callback. That fixes the narrower stale-callback ordering issue identified by Cubic.

I also withdrew my start→cancel thread as a blocker for this ASG-only PR after verifying that both native Bluetooth SDKs single-flight hotspot requests. The gallery cancellation behavior around `request_in_flight` is real, but pre-existing and better handled separately on the phone/SDK side.

I am still requesting changes for two lifecycle-ownership problems:

1. The original stop→restart thread is not fully fixed. `stopHotspot()` acknowledges disabled before the asynchronous SmartXY AP is physically down. A sequential, supported enable can therefore accept the previous `ap0` and `xy_*` values as readiness before teardown/restart completes. The new lock orders submission from this process; it does not establish physical completion.
2. Unlike `LocalOnlyHotspot`, the SystemUI-owned AP survives an `asg_client` crash or process kill. The inactivity timer and cached state disappear, and `initialize()` neither adopts nor disables an already-running AP. That can leave the glasses hotspot active indefinitely after an update/crash. The newly opened process-restart thread is valid and is a regression of moving lifecycle ownership out of the app reservation.

The architecture is still the correct direction—one K900-native SmartXY path—but lifecycle reconciliation remains incomplete. A bounded physical teardown/restart transition plus startup reconciliation (adopt with a fresh idle timer, or disable) would keep this targeted without reviving the hybrid design.

Validation at this head:
- Focused gateway unit tests and `:app:assembleDebug`: pass.
- `./scripts/check-android-compile.sh asg`: pass when run serially.
- `git diff --check`: pass.
- Remote CI/checks: complete and green.
- No new lifecycle tests were added; the test report still contains only the two gateway helper cases.
- No Mentra Live is attached here, so I could not independently run stop→immediate-start or process-kill hardware checks.

PR #3725, opened 2026-08-20
Keep blob share exports alive for async consumers

githubdoramon:
Recorder exports must remain readable after the share chooser returns because recipient apps can consume the file asynchronously. The original implementation deleted the renamed copy in `finally`; Android react-native-share resolves when a target is selected, which confirms the premature-deletion race described in the report.

Current dev already retains and reuses these cache copies. This update preserves that implementation and fixes its remaining lifetime gaps: every handoff renews a persisted 24-hour retention timestamp, and cleanup skips copies while any share chooser using them is still open. The timestamp is renewed again when the chooser finishes. Cancellations retain prepared copies; failed preparation is cleaned up. Expired copies are pruned on a later share. Storage remains best-effort app cache and can be evicted by the OS.

Validation:
- Actual-source delayed-read harness: the original implementation loses the exported file after handoff; the original PR and current dev preserve its bytes.
- The new regression suite fails on unchanged dev for renewed retention and overlapping chooser lifetimes, and passes with this update.
- Focused sharing and filename suites: 21 tests passed, 40 assertions.
- Full engine test command passed: 515 Bun tests plus 10 Node tests.
- Engine TypeScript check (`tsc --noEmit`) passed.
- Gmail attachment and WhatsApp audio behavior have not been tested on a physical phone; the verified result is the file-lifetime fix, not recipient-specific end-to-end validation.


PhilippeFerreiraDeSousa:
Review and verification summary (posted after merge):

I checked the code against the original bug report rather than relying on the PR description. The original implementation deleted the renamed export in `finally` after `Share.open` returned. Android react-native-share resolves when the chooser target is selected, before the recipient necessarily finishes reading. An actual-source delayed-read harness confirmed that the original implementation loses the exported file after handoff.

Current dev already retained and reused share cache copies, so the original immediate-deletion fix was already present. The remaining problems were that re-sharing an older copy did not renew its retention age, and cleanup could remove a copy while another share chooser was still using it. The merged update preserves cache reuse, persists a fresh 24-hour timestamp at each handoff and chooser completion, and skips copies while any chooser using them remains active. Expired copies are pruned on a later share; OS cache eviction remains possible.

Validation: the two new lifetime regressions failed against unchanged dev and passed with the fix. All 515 Bun engine tests plus 10 Node tests passed, along with the engine TypeScript check. Gmail attachment and WhatsApp audio behavior were not tested on a physical phone; the confirmed result is the file-lifetime fix, not recipient-specific end-to-end validation.

The PR description was corrected to match the final code and evidence. At merge time, GitHub's gate was red because the fork's Android, iOS, mobile quality, and OEM boundary workflows were `action_required` and had not run. The maintainer-requested admin bypass was used; remote CI was not verified. Merged as `1bcec19da9a1b0e2f161991c3eecfb76c2595153`.

Issue #179, opened 2025-01-27
Notification App keeps crashing


Streetlamp123:
The notification app seems to be constantly crashing. 
Running on GrapheneOS, android 15 (Build 2025011500)

See attached logs

[Show Notifications on Glasses log 0b17f00c8c50.txt](https://github.com/user-attachments/files/18553208/Show.Notifications.on.Glasses.log.0b17f00c8c50.txt)

CaydenPierce:
If you run the latest main on both the notifications repo and the AugmentOS repo, it should work for you.

blathers123:
Is there a build available on the notification repo? I didn't see one. 

Also when I download from the built in app store it downloads "Downloading Notify" as a bin file

Streetlamp123:
Is there a build available on the notification repo? I didn't see one.

Also when I download from the built in app store it downloads "Downloading Notify" as a bin file

PR #3777, opened 2026-08-25
Add advisory empty-scan diagnostics and pairing guidance

CarlKho-Minerva:
When a scan finishes empty and the phone reports a matching connected device, offer an optional `device_connected_on_phone` advisory. The Mentra App pairing screen displays “Glasses may be in use” with instructions to disconnect from another app if applicable and retry.

This preserves successful empty completion. It replaces the original `onError("device_held_by_other_app")` design because phone connection state cannot identify the owning app or prove why advertising was absent.

- Android adds opt-in `ScanDiagnosticCallback` / `MentraBluetoothScanCallback.onDiagnostic` while preserving the original `ScanCallback` interface and existing callers. Matching uses saved identity or existing model name patterns.
- iOS adds a scan overload with `onDiagnostic`, using the existing CoreBluetooth manager and service-filtered connected peripherals. Existing scan signatures remain available.
- React Native exposes `ScanOptions.onDiagnostic`, `useBluetoothScan().diagnostic`, and `useMentraBluetooth().scan.diagnostic`. Older native binaries omit the advisory. Cancellation and retry suppress stale diagnostics and wait for a delayed native start to stop.
- Diagnostics run only for empty natural completion, before completion is delivered. Cancelled/nonempty scans, simulated devices, known SDK connections/connection attempts, and unavailable Bluetooth access remain silent.
- The Mentra App shows the advisory after its 15-second empty scan timeout, including non-Live models, and clears it on retry, results, connection, and leaving the screen. Updated SDK documentation covers all three interfaces.

[Companion Starter Kit PR #179](https://github.com/Mentra-Community/Mentra-Bluetooth-SDK-Starter-Kit/pull/179) adds the hint to native Android, native iOS, and React Native example pickers. That draft targets `dev` and requires the SDK release containing this API before its dependency pins can be updated and it can merge.

Validation:
- Public Android AAR compile and 155 Android SDK unit tests passed, including matcher and callback-completion coverage.
- Original-shape Java `ScanCallback` binary compatibility fixture passed against the additive Android API.
- Swift SDK build and 23 tests passed.
- Eight React Native scan-session tests passed, including empty/nonempty completion, cancellation during lookup, delayed native start, unavailable diagnostics, and start failure.
- SDK and engine TypeScript checks passed; 21 Mentra App pairing-screen tests passed, including hint display/retry and stale lookup suppression.
- Companion native Android debug build, unsigned native iOS iPhone build, and React Native type check passed against local SDK artifacts. The iOS simulator link is blocked by the installed ARM64 GStreamer device binary.

Physical phone/glasses validation, full Mentra App iOS/Expo build, and UI screenshots are still pending. Local checks do not establish remote CI or hardware success. Keep this PR unmerged for design review; no auto-merge is enabled.


PhilippeFerreiraDeSousa:
Design review for `cda6fca00dbbd74ab222fda09032f69bb5b45fc2` — reworked and intentionally left unmerged for maintainer review.

The underlying idea is useful: a same-phone GATT connection can help explain an empty scan. I do not recommend the original design, which sent `onError` and then `onComplete`, and claimed another app held the connection. Android [documents that this query reports profile-wide connection state, not application-specific state](https://developer.android.com/reference/android/bluetooth/BluetoothManager#getConnectedDevices(int)). It does not establish ownership, exclusivity, or the cause of missing advertisements.

The proposed API keeps the signal advisory:

```text
Empty completed scan + matching connected device:
    onDiagnostic(code = "device_connected_on_phone")
    onComplete(emptyList())

Ordinary completed/cancelled scan:
    onComplete(results)

Failure to start:
    onError(error), preserving existing behavior
```

`ScanDiagnosticCallback` is a separate opt-in interface; `MentraBluetoothScanCallback` supplies a no-op diagnostic method that clients can override. Existing `ScanCallback` implementations and scan signatures stay unchanged. The message suggests disconnecting in another app *if* that app is using the glasses. Identity matching rejects blank saved identities and prefers known addresses over conflicting names. Matching by advertised-name pattern remains a model-level heuristic, not proof of identity.

Why this design: retries alone cannot remedy a device that is not advertising; turning this heuristic into a terminal error misrepresents a completed scan; inserting connected peripherals into scan results or automatically attaching to them would change connection behavior without proof that concurrent protocol clients work safely. An optional explanation is the smallest useful change for the reported native Android consumer. React Native/iOS parity is outside this PR.

Validation: public-mode Android SDK AAR build passed; all 155 Android SDK unit tests passed, including 12 matcher tests and 5 completion/diagnostic tests. A compatibility harness loaded old-shape Java callback binaries against the rebuilt SDK and verified one completion for both direct callbacks and adapter subclasses. The two-app hardware scenario and new diagnostic have not been independently verified on a physical phone.

GitHub's current gate is red because Android, iOS, mobile quality, and OEM boundary workflows require fork approval (`action_required`); remote CI is not verified. Auto-merge is off. This PR remains open pending approval of the new design.

PR #3907, opened 2026-08-31
fix(display): measured Cyrillic glyph widths for the G2 profile

PopaSpinka:
G2 inherited G1's uniform 18px Cyrillic width, causing the host to wrap Russian text earlier than the proportional widths measured in the official simulator. Add 66 Russian Cyrillic glyph advances to the G2 profile, keeping G1, existing Latin/digit metrics, and unmapped-script fallbacks unchanged. The existing glyph-map precedence and render formula support these values without changes to the measasurer or wrapper.

All 66 advances were independently reproduced with `@evenrealities/evenhub-simulator@0.9.5` and `@evenrealities/even_hub_sdk@0.0.15`: render each glyph 8 and 16 times, capture `/api/screenshot/glasses`, and divide the difference in alpha-bounds widths by 8. There were no mismatches. The simulator documentation explicitly notes that font rendering can differ from hardware; physical G2 verification reported by the contributor has not been independently reproduced.

The wrap claim is now precise: 52 copies of `о` at 11px fit in 576 host pixels; 51 fit in 568 usable pixels after native G2's 4px padding on each side. Scene processing currently uses the full box width. This PR changes glyph metrics only and does not resolve that existing host/native padding mismatch or guarantee every physical wrap boundary.

Validation:
- Expanded regression suite: 5 tests, 415 assertions, including both sides of wrap boundaries and the actual scene-processing path.
- Against unchanged dev, four tests fail, including the reported-length Russian scene; all pass with the new table.
- Engine TypeScript check passed.
- Full engine test command passed: 520 Bun tests plus 10 Node tests.


PopaSpinka:
Added a test that runs the real `TextMeasurer`/`TextWrapper` pipeline against both profiles: Cyrillic becomes proportional on G2 (a full 576px line now fits 51 chars instead of 32) while G1 keeps its uniform 18px, and Latin/digits/CJK stay byte-identical. All passing under `bun test`:

```
4 pass / 0 fail — 14 expect() calls
```

PhilippeFerreiraDeSousa:
Pre-merge review of `70a60910b165e50bc126ecafa8773dffb1525f4d`:

The code checks out for the early host-wrapping bug. I independently measured every one of the 66 Russian Cyrillic advances in the official simulator 0.9.5 using SDK 0.0.15. For each glyph, I rendered 8 and 16 copies, captured the glasses framebuffer, and calculated `(alphaBounds16 - alphaBounds8) / 8`. All 66 matched the submitted table. The G2-only glyph map uses the existing render formula correctly, including odd advances; it does not mutate G1.

I corrected the source comments, PR description, and tests where they overstated the evidence. The official simulator [documents that font rendering can differ from hardware](https://www.npmjs.com/package/@evenrealities/evenhub-simulator), so these measurements are not physical-device proof. Also, 52 copies of an 11px glyph fit 576 host pixels; 51 fit 568 usable pixels after native G2's 4px padding on each side. The existing scene processor uses the full box width, so this narrow metrics fix does not resolve that host/native padding mismatch or establish all physical wrap boundaries.

Validation: expanded regression coverage passes (5 tests, 415 assertions), including both sides of wrap boundaries, the real scene-processing path, every inherited G1 glyph, and unmapped Cyrillic/CJK/Korean fallbacks. Four tests fail against unchanged dev, including the Russian scene that is prematurely truncated. The full engine suite passes: 520 Bun tests plus 10 Node tests. Engine TypeScript checking and scoped formatting/whitespace checks pass.

GitHub's gate is red because the fork Android, iOS, mobile quality, and OEM boundary workflows are `action_required`; they have not run. Local verification is complete, but remote CI is not verified. Proceeding with the maintainer-requested admin bypass, with these findings documented before merging.

PR #3938, opened 2026-09-05
fix(ota): keep release notes readable on small screens

PhilippeFerreiraDeSousa:
## Problem
The completion screen put the heading above a changelog card inside a fixed-height, non-scrolling page. On short windows or with enlarged text, the notes viewport could disappear. The issue was reproduced with both current source components in a browser harness: at 390x600 and 2x text, the Engine notes viewport was 7px and the Starter Kit viewport was 0px.

## Change
Keep the same expanding card and pinned action footer, but allow the page above the actions to scroll when it overflows. Give the inner notes ScrollView a definite starting height of 120 and flexGrow: 1, with a 200 minimum card height. The card still grows into spare space on normal screens; large titles can grow the card instead of squeezing the notes away. A definite inner height is important: flex alone inside the outer ScrollView can expand to the entire Markdown height and break independent scrolling.

Per Philippe: remove Scroll for more, its reserved row, three state variables, and all hint-related layout/scroll callbacks. Keep the native vertical indicator; persistentScrollbar also prevents it fading on Android. iOS retains its system-managed indicator behavior.

No viewport breakpoints, new layout mode, measurement state, or OTA lifecycle changes. The matching Starter Kit change targets staging too. After landing, propagate by merging staging into dev, not by duplicating the fix.

## Validation
- Mobile TypeScript check: passed.
- All four mobile OTA presentation/progress suites: 80 tests passed, including completion/up_to_date content scrolling and Done outside the scroll region.
- Existing Markdown rendering regression still passes.
- 12 Yoga layout cases using the actual style objects: minimum notes height and bounded inner scrolling preserved.
- 22 browser scenarios using the actual source components, mocked OTA state, and plain-text Markdown renderer: phone/short/desktop sizes, 1x/2x/3x text, complete/up_to_date/empty notes; both scroll regions and Done exercised. Screenshots inspected.
- Companion Starter Kit TypeScript and 23 OTA tests passed.
- No physical-device OTA or native-app rebuild was performed for this layout-only change.

---

> [!NOTE]
> **Low Risk**
> Layout-only OTA UI changes with no update lifecycle or networking impact; risk is limited to presentation on completion screens.
> 
> **Overview**
> Fixes OTA **complete** and **up to date** screens where long headings, large text, or short viewports could collapse the release-notes area to almost nothing.
> 
> **FlowPage** now puts icon, title, and body in an outer `ScrollView` (`ota-page-scroll`) while **Done** stays in a fixed footer, so the whole content column can scroll when it overflows. **ChangelogList** drops the custom “Scroll for more” hint, its reserved row, and layout/scroll measurement state; the notes `ScrollView` keeps a visible scrollbar (`showsVerticalScrollIndicator`, `persistentScrollbar` on Android) with a **120px** inner height, **200px** minimum card height, and `flexGrow` so nested scrolling still works inside the page scroll.
> 
> English copy removes `ota:scrollForMore` from the engine defaults and `en.ts`. Tests assert the new layout styles and that **complete** / **up_to_date** flows scroll page content while **Done** remains outside the scroll region.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed076e80c1b46827e2ac62a729b308f22d983231. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mobile/modules/engine/src/react/MentraLiveOtaFlow.tsx
+++ b/mobile/modules/engine/src/react/MentraLiveOtaFlow.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react-native/no-raw-text -- BodyText and PercentText are local Text wrappers. */
-import React, {useMemo, useState} from "react"
+import React, {useMemo} from "react"
 import {
   ActivityIndicator,
   Pressable,
@@ -99,7 +99,6 @@ const ENGLISH_COPY: Record<string, string> = {
   "ota:updateLater": "Later",
   "ota:updateComplete": "Update complete",
   "ota:whatsNew": "What's new",
-  "ota:scrollForMore": "Scroll for more",
   "ota:upToDate": "Up To Date",
   "ota:devBuild": "Development Build",
   "ota:devBuildNoOta":
@@ -323,12 +322,7 @@ function OtaFlowContent({
             {translate("ota:updatedToVersion", {version: state.releaseTransition.toVersion})}
           </BodyText>
         ) : null}
-        <ChangelogList
-          changelogs={state.changelogs}
-          colors={colors}
-          scrollHint={translate("ota:scrollForMore")}
-          title={translate("ota:whatsNew")}
-        />
+        <ChangelogList changelogs={state.changelogs} colors={colors} title={translate("ota:whatsNew")} />
       </FlowPage>
     )
   }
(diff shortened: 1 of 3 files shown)


PhilippeFerreiraDeSousa:
Matching Starter Kit staging fix: https://github.com/Mentra-Community/Mentra-Bluetooth-SDK-Starter-Kit/pull/141. Both use the same layout rules. The first flex-basis-only attempt was rejected during validation because native Yoga still expanded the notes to their full content height; the final change bounds the inner ScrollView itself, and all 12 Yoga cases and 22 browser scenarios now pass.

PhilippeFerreiraDeSousa:
Additional verification: re-ran all 22 browser scenarios with the real react-native-marked useMarkdown renderer (not the earlier plain-text stub). All passed, including both scrolling regions and Done at 320x400 with 3x text. OTA state is still mocked and native platform views use React Native Web; this is layout verification, not a physical OTA test. The 12 Yoga cases also pass.

PR #3949, opened 2026-09-08
Report the glasses client package so OTA skips sideloaded builds

aisraelov:
## Why

From [Cayden's hotspot OTA report](https://mentra-labs.slack.com/archives/C096YCLEBFW/p1788822194596509): a glasses OTA that prompted, "installed," and then prompted again, forever. He fixed it by removing his third-party ASG client.

That's the whole bug. An ASG build made without Mentra's release keystore gets `applicationIdSuffix ".thirdparty"` (`asg_client/app/build.gradle:319,328`), so it installs as `com.mentra.asg_client.thirdparty` and **coexists** with the stock system app — deliberate, since Android won't let a differently-signed APK replace a system package.

But `sendVersionInfo()` reports `build_number` from whichever client is holding the BLE link, and the phone compares that against `versionJson.apps["com.mentra.asg_client"]`. Nothing compared the two identities. So:

1. `.thirdparty` client answers `version_info_1` with build 120
2. manifest pin says 137 → "update available"
3. phone pushes the APK, which installs **`com.mentra.asg_client`** (stock)
4. `.thirdparty` is still the running client, still reports 120
5. → loop

## What this does

The glasses report their runtime package identity, and both OTA check paths refuse to answer for a client they don't manage.

- **`asg_client`** — `package_name` in `version_info_1`, from `getPackageName()`.
- **Bluetooth SDK** (Android + iOS + TS types) — parses it into `packageName` on `GlassesStatus` / `VersionInfoResult`.
- **Engine `OtaUpdateCheckService`** — skips with a new `unofficial_client` reason, carrying the package name, before the manifest is even fetched.
- **Public SDK `checkForOtaUpdate()`** — rejects with an `unofficial_client` error, alongside the existing `glasses_not_connected` / `missing_glasses_version` guards.
- **UI** — new `unofficial_client` screen: *"Your glasses are running a sideloaded client (com.mentra.asg_client.thirdparty), so updates are blocked. Restore the stock client to update them."* The ambient prompt in `OtaUpdateChecker` already returns on any `skippedReason`, so the nag stops too.

`package_name` deliberately sits in **chunk 1**, next to `build_number`: the phone's check waits on `build_number` and then decides immediately, so identity arriving in `version_info_3` would leave a window where the check runs and assumes stock. (Same class of bug as the `wire_caps` ordering issue already commented in `MentraLive.kt`.)

An **absent** `package_name` means glasses whose client predates the field — those keep the existing behavior, so nothing in the field regresses.

## Design notes

One field, not three. I'd initially also proposed a signing-cert check and a build-channel flag; both are redundant given the suffix mechanism. `.thirdparty` is applied by `!useReleaseKey` on *both* buildTypes, so "no Mentra key" ⇒ different package, always. A cert check would only catch a fork that kept the stock package name while being differently signed — which can't install over the system app in the first place.

This is self-asserted over BLE and a determined fork can lie. That's fine: it's a footgun guard and a support signal, not DRM.

## Not in this PR

- **A phone-side loop breaker.** There is currently no "the install didn't take" detection anywhere (`otaAutoChain.ts`, `OtaUpdateCheckService.ts`, the install coordinator — nothing). `package_name` won't reach a fork built off an older commit, so "APK install completed but `buildNumber` didn't move ⇒ stop, don't re-prompt" is what actually guarantees nobody repeats Cayden's hour. Worth its own PR.
- **A stale gradle comment.** `asg_client/app/build.gradle:75` claims the `-dev` versionName suffix makes the phone skip OTA. That code does not exist — `appVersion` is never read in the OTA path. Unpinned Mentra-internal builds are only *accidentally* safe because their versionCode is `baseline + epoch_seconds`, and downgrades are off (`DOWNGRADE_FLOOR_VERSION_CODE = 0`). Raise that floor and every Mentra dev's local build starts getting rolled back to the pinned release.
- Cayden's other findings (speaker buzz, restart at the wrong time, "Restarting Mentra Live" never restarting, the 0→100 progress bar cycling with no indication it's multiple artifacts) are unrelated to this.

## Test evidence

- `bunx jest` — 97 suites, 801 passed / 2 skipped
- `bunx tsc --noEmit -p mobile` — clean
- `./scripts/check-android-compile.sh` (asg + bluetooth-sdk) — BUILD SUCCESSFUL
- `swiftc -parse` on all four touched Swift files — clean
- New cases in `otaUpdateCheckService.test.ts`: sideloaded package skips without fetching the manifest; stock package checks normally; **absent** package (pre-field glasses) checks normally

Not hardware-tested — the end-to-end confirmation is installing a `.thirdparty` build on a Mentra Live and seeing the update prompt replaced by the blocked screen.

---

> [!NOTE]
> **Medium Risk**
> Changes the OTA decision path on glasses connect, but only adds a fail-closed guard for non-stock packages while preserving behavior when `package_name` is absent.
> 
> **Overview**
> Fixes the **infinite OTA re-prompt** when glasses run a sideloaded ASG client (`com.mentra.asg_client.thirdparty`) that coexists with the stock app: the phone was comparing that client’s `build_number` to the stock manifest pin and installing an APK that never replaced the running client.
> 
> The glasses now send **`package_name` in `version_info_1`** (next to `build_number`) from `getPackageName()`. The Bluetooth SDK on Android, iOS, and TypeScript parses it into **`packageName`**, omits it on chunks that don’t carry it so later `version_info` responses can’t wipe a known sideload identity, and **clears it with `buildNumber` at each BLE session boundary**.
> 
> **Engine** and public **`checkForOtaUpdate()`** skip or reject with **`unofficial_client`** before fetching the manifest when a non-stock package is reported; missing `package_name` still means pre-field stock glasses. The Mentra Live OTA UI adds an **“Updates blocked”** screen naming the sideloaded package, and docs/tests cover the new skip reason and session-staleness regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c65e00ed2f9591b019910315b3aac45a428d19aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Stops the OTA re-prompt loop when glasses run a sideloaded ASG client. A build made without Mentra's release keystore installs as `com.mentra.asg_client.thirdparty` and coexists with the stock app, so its build number was compared against the stock manifest pin — the phone installed the stock APK, saw the same version from the sideloaded client, and re-prompted forever. Glasses now report their package name, and both OTA check paths refuse to serve anything that isn't the stock client.

- `package_name` sits in `version_info_1` next to `build_number`, so identity arrives before the check decides.
- Android/iOS/TS SDKs parse it into `packageName` on `GlassesStatus` and `VersionInfoResult`, including the public `version_info` event, and omit it when the chunk doesn't carry it so a later resolution can't blank a known sideloaded identity.
- Engine's OTA check and public `checkForOtaUpdate()` skip with a new `unofficial_client` reason/error before fetching the manifest.
- UI shows an "updates blocked" screen naming the sideloaded package.
- Identity is session-bound: cleared together with `buildNumber` only at the native session boundary — the engine store deliberately keeps the pair through disconnects so a sideloaded build can't read as stock mid-session.
- Missing `package_name` means pre-field glasses, which keep existing behavior.

Not addressed: no phone-side loop breaker for "install completed but the reported version didn't change," which is what actually guarantees this can't recur.

<sup>Written for commit c65e00ed2f9591b019910315b3aac45a428d19aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/3949?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aisraelov:
Bots found a real cluster in `782a81c` — 8 cubic violations, 2 Codex, 1 Bugbot, all reducing to three root causes. All three were valid; fixed.

**1. `packageName` was never cleared at a session boundary** (Bugbot medium, Codex P2, cubic ×4)

Both platforms already clear `buildNumber` / `appVersion` / `besFirmwareVersion` / `mtkFirmwareVersion` / `otaVersionUrl` on a fresh session, for exactly this reason — I added a field to that set and didn't add it to the clear list. So a `.thirdparty` unit seen once would keep blocking OTA for the *next* glasses, including a restored stock client. That's the specific thing the new screen instructs the user to do, so the bug would have made the remedy look broken.

Cleared alongside `buildNumber` in `MentraLive.kt` (connected branch), iOS `handleGlassesReady()`, and on disconnect in the engine store next to `hotspotOtaVersion`.

This also resolves cubic's P1 race (cached build, no package yet). Clearing the two *together* is what makes the guard sound: they travel in the same `version_info_1`, and the transport reassembles a chunked message atomically before dispatch, so a present build number now implies this session's identity. There's no window where one is set without the other.

**2. `packageName` dropped from the public `version_info` event** (Codex P1, cubic P2)

`Bridge.sendVersionInfo` builds an explicit body on both platforms and I didn't add the field, so `requestVersionInfo()` returned an empty package. Android's `getFreshGlassesStatus()` also merges field-by-field and didn't carry it — so the SDK's own `checkForOtaUpdate()` would read a sideloaded client as legacy and let it through. Added to both bridges and the Android merge. (iOS merges `versionInfo.dictionary` wholesale, so it follows once the bridge carries it.)

**3. Stock renderer had no copy** (cubic P2)

`MentraLiveOtaFlow`'s `ENGLISH_COPY` was missing the three new keys, so any host using the default `translate` would render the literal `ota:unofficialClient`. Added, matching `en.ts`.

Regression test added for the session-boundary case: `.thirdparty` unit → disconnect → stock unit checks normally.

Local: `bunx jest` 97 suites / 802 passed, `tsc --noEmit` clean, `check-android-compile.sh` (asg + bluetooth-sdk) BUILD SUCCESSFUL, `swiftc -parse` clean on all touched Swift.

PR #3973, opened 2026-09-09
Report host build facts in Bluetooth SDK analytics and re-enable them in the Mentra App

PhilippeFerreiraDeSousa:
## Why

Glasses WAU is measured from the Bluetooth SDK's PostHog events (see `agents/wau-telemetry-restoration-plan.md`, Phase 1). Two gaps block a trustworthy number:

1. **Nothing separates production from everything else.** Store installs, TestFlight, Play testing tracks, sideloaded dev builds, and simulators all look identical in PostHog.
2. **The Mentra App has SDK analytics turned off on staging.** `app.config.ts` passed `analytics: false` to the plugin since 0f2a2f29e8 (Cloud V2 support profiles, 2026-08-11), on the theory that support profiles duplicate it. They do not: support profiles are per-account and carry no glasses identity. Every TestFlight beta since 105 has shipped dark, and the next store release from staging would take `com.mentra.mentra` WAU to zero.

## What

**SDK events (Android + iOS), additive properties on every event**

| Property | Values |
|---|---|
| `app_version`, `app_build` | host app version name / build number |
| `app_build_type` | `debug` / `release` (Android `FLAG_DEBUGGABLE`, iOS `#if DEBUG`) |
| `app_install_source` | `play_store`, `app_store`, `testflight`, `adhoc_or_dev`, `simulator`, `sideload`, `amazon_appstore`, `galaxy_store`, `huawei_appgallery`, `xiaomi_getapps`, `other_store` |
| `app_installer_package` | raw Android installer package when present |
| `app_environment` | optional host-declared lane, validated `[a-z0-9][a-z0-9_-]{0,31}` |

iOS ordering: simulator → embedded provisioning profile (`adhoc_or_dev`) → receipt name (`sandboxReceipt` → `testflight`, `receipt` → `app_store`) → `unknown`. The profile check comes first because TestFlight and App Store builds never carry one; missing or unrecognized evidence is reported as `unknown`, never guessed. Android cannot tell Play production from Play testing tracks (both `com.android.vending`); `app_environment` is the host's way to add that, and a failed installer lookup is `unknown`.

**Production cohort rule** (documented in the README and the plan): `app_install_source ∈ {app_store, play_store}` AND `app_build_type = release` AND, for hosts that declare a lane, `app_environment = prod`. Store installs of packages without a lane are an unclassified cohort, not assumed production.

**Connection + identification events:** `glasses_is_simulated`. **Identification only:** `glasses_firmware_version`, `glasses_bes_firmware_version`, `glasses_mtk_firmware_version`, `glasses_android_version`, `glasses_app_version`, `glasses_build_number`, so a missing or malformed serial can be tied to firmware.

**iOS `sdk_version` is always present** (`"unknown"` for unstamped source builds) instead of being omitted; the shipped 3.1 iOS build reports null today.

**Plugin:** `analytics.environment` option (shared resolver in `plugin/src/analyticsProps.ts`), stamped as `com.mentra.bluetoothsdk.analytics.environment` metadata / `MentraBluetoothSdkAnalyticsEnvironment` Info.plist key, cleared when unset.

**Mentra App:** `analytics: {environment: EXPO_PUBLIC_BUILD_ENV}` replaces `analytics: false`.

**Docs:** SDK README, `api-reference.mdx` (adds the identification event and the new properties, and drops a stale claim that the SDK never uploads a serial), `production-checklist.mdx`, `changelogs/3.1.0.md`.

**Release gate (iOS):** `release-ios.mjs` now fails when the IPA's Info.plist lacks the stamped SDK version, has the wrong analytics lane for `EXPO_PUBLIC_BUILD_ENV`, or has SDK analytics disabled (`scripts/release-sdk-metadata.mjs`, unit-tested incl. a real binary-plist IPA fixture on macOS). Runtime keeps `sdk_version="unknown"` as a fallback; the gate is what stops the 3.1 iOS gap from recurring.

Host facts are resolved once, lazily, on the analytics executor / transport queue, so no PackageManager or Bundle lookups run on the caller thread. No PII added: install source and build type are properties of the binary, not the person.

## Review round

- Codex (local, Astra) and the GitHub bots: `unknown` install source, production-cohort definition, firmware-fields wording, environment pattern wording, `event_kind` table, tautological iOS test. All addressed above.
- Firmware fields on identification cannot diagnose a *missing* serial (identification needs one). Rejected the suggested fallback identification event: an unidentified pair must not enter the identification stream. Coverage is measured as connections without identification, per model and SDK version.
- Simulated-glasses ordering (Android publishes connected before `deviceModel`, so `glasses_is_simulated` is wrong on that first event): fixed generically in #3974, whose tracker defers the connected event until the model is known. Merge both before a release; a fresh simulated session has no serial and never enters the identification stream, so WAU itself is not affected.

## Test evidence

- `./scripts/check-android-compile.sh bluetooth-sdk` → BUILD SUCCESSFUL
- `:mentra-bluetooth-sdk:testDebugUnitTest --tests BluetoothSdkAnalyticsHostTest -PmentraPublicSdk=true` → passes
- `xcodebuild -scheme MentraBluetoothSDK -destination 'iOS Simulator,iPhone 17 Pro' test -only-testing:…BluetoothSdkAnalyticsHostTests` → 6/6, TEST SUCCEEDED
- Plugin Jest (`analyticsProps.test.ts`, `withIos.test.ts`) → 7/7
- `node --test mobile/scripts/release-sdk-metadata.test.mjs` → 4/4 (IPA fixture test runs on macOS, skipped elsewhere)
- After the review round: Android `BluetoothSdkAnalyticsHostTest` 4/4, iOS `BluetoothSdkAnalyticsHostTests` 7/7

## Still to do on device (Phase 4 of the plan)

Mentra Live 023B + iPhone TestFlight + Z Fold Play internal: confirm `app_install_source` reads `testflight` / `play_store` with `app_environment=staging`, and that identification carries the firmware fields.


PhilippeFerreiraDeSousa:

**Verdict: good design, with a few targeted corrections before merge.** Keep the native host-facts resolver, shared plugin configuration, and SDK-owned WAU event. This is a sensible, low-complexity separation of responsibilities. I would not add fallback identification events or firmware-waiting logic just to satisfy the reviews.

Reviewed head `d945641`, matching GitHub, using refs only.

Changes I would make, in importance order:

1. **Preserve “unknown” install source instead of guessing production.**  
   In `BluetoothSdkAnalyticsHost.swift:38–41`, no profile plus a missing/unrecognized receipt becomes `app_store`; the test explicitly enshrines the missing-receipt case. Missing evidence should return `unknown`, particularly when this property controls production inclusion. Apple documents that the receipt URL can be nil. Android likewise maps lookup failures/missing installers to `sideload`, although null has other documented causes. Keep the classifier small; add an unknown outcome rather than more detection machinery. [Apple documentation](https://developer.apple.com/documentation/foundation/bundle/appstorereceipturl?changes=_3&language=objc), [Android documentation](https://developer.android.com/reference/android/content/pm/InstallSourceInfo#getInstallingPackageName())

2. **Make the production cohort contract explicit in the plan and docs.**  
   The PR correctly acknowledges that Play production and testing tracks both report `play_store`. But the plan’s proposed filter uses only store source and simulation status, so it would include Play internal builds. For the Mentra App, require the declared production environment alongside store source and release build type. For third-party hosts without a lane, retain an unclassified cohort. Also, `app_environment=prod` describes the binary’s configuration—it cannot distinguish testing versus production distribution of the *same binary*. Resolve that limitation explicitly rather than inventing a `play_store_internal` detector.

3. **Verify actual release metadata, especially iOS SDK version.**  
   `sdk_version ?? "unknown"` preserves event delivery, which is appropriate, but does not fulfill the plan’s version-provenance requirement. The existing plugin test verifies a helper writes the key; it does not prove the key survives the Mentra App’s complete prebuild/archive process. Add a release-artifact check for the expected SDK version and environment. Missing metadata should fail release validation, not crash SDK initialization on somebody’s phone.

4. **Fix connection/model ordering once, in the lifecycle work.**  
   Android constructs `Simulated`, which publishes connected status, before `initSGC` publishes its model. The new flag therefore records false on the first connection event. Establish the model before publishing connected state, or implement the shared model-ready emission rule planned for #3974. Cover the real connection sequence. Avoid simulation-specific retries, timers, or corrective events. If #3974 owns this, make that dependency explicit before shipping the combined restoration.

My assessment of the existing comments:

| Comment | Recommendation |
|---|---|
| [Codex: simulated flag ordering](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973669469) | **Act on the ordering bug.** The claimed WAU pollution is overstated: a fresh simulated session has no serial and therefore emits no `glasses_identified` event. The demonstrated corruption is in connection metadata. |
| Cubic: firmware missing when serial is invalid ([Android](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973657974), [iOS](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973657983)) | **Valid observation; reject the proposed identification workaround.** Correct the claim that this metadata diagnoses missing serials. An unidentified device must not enter the identification event stream. Separate coverage diagnostics can follow if needed. |
| Cubic: environment syntax, README, and `event_kind` documentation | **Worth fixing as small documentation edits.** No validator redesign needed. |
| [Cubic: tautological iOS assertions](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973658016) | **Nonblocking test-quality nit.** Testing extracted bundle values would help; deleting assertions merely to appease the comment adds little value. |

**Remaining reliability limits:** firmware properties are a snapshot when the serial first becomes available; later version updates are not captured again. The Mentra Live handlers do write firmware before serial within the same parsed response, so I do not see a reason to delay identification globally. Existing fire-and-forget delivery and connection-only identification can still undercount offline usage and connections spanning weeks; those belong in the planned #3974 reliability work.

I found no new obvious threading or Bluetooth-blocking regression in the host-property collection. `git diff --check` passed. I did not rerun builds/tests or perform device verification; the PR reports those build/unit checks, with device validation still pending. No files were modified.

PhilippeFerreiraDeSousa:
## Local Codex review (gpt-6-astra, medium) — pass 2

_Read-only run of `codex exec` after the first-pass fixes._

**The design is sound. I would approve this revision and merge into `staging` once the remaining CI build passes.** I found no blocking source issue at `621cd9b`. One build was still running when I checked.

The approach stays simple: report host facts on the existing SDK events, keep production-cohort policy in the dashboard, and validate release metadata in the exported IPA. The shared plugin resolver and cached native lookups are appropriate. The first-pass fixes improve reliability without introducing another telemetry mechanism.

What I would change, in importance order—neither blocks this staging merge:

1. **Make the Mentra App release gate require an explicit lane.** [`expectedSdkAnalyticsMetadata()`](https://github.com/Mentra-Community/MentraOS/blob/621cd9ba540e4a0778f37baeaeb759d234394885/mobile/scripts/release-sdk-metadata.mjs#L17) returns no expected environment when the variable is absent, so validation skips that field. A small required-value check would make this release-specific contract complete. Coordinated release CI already supplies and validates the lane, so this is hardening for other release invocations, not a demonstrated production-CI failure.
2. **Finish the wording cleanup.** The API reference should match the README: specifically `app_store`/`play_store`, with undeclared lanes kept unclassified. The PR description and native firmware helper comments still claim firmware fields diagnose missing serials, despite the “Review round” correctly explaining that they cannot.

My assessment of the review comments:

| Comment | Assessment |
|---|---|
| [Missing release lane](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973819328) | Worth the small hardening above. Its claim that a `dev` lane masquerades as production is wrong; that lane excludes the event from production. |
| [Simulated-glasses ordering](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973669469) | Real correctness issue for connection/coverage data. Keep the generic ordering fix in #3974; do not add a simulator-specific workaround here. Ship both together. |
| [Firmware on unidentified glasses](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973657974) | Correct observation, unsuitable proposed fix. Do not emit identification without a valid serial or delay valid identification waiting for optional firmware. Correct the remaining misleading comments. |
| [Exact production store set](https://github.com/Mentra-Community/MentraOS/pull/3973#discussion_r3973819332) | Valid documentation nit. |
| Environment syntax, missing `event_kind`, tautological iOS assertions | Addressed adequately. Further churn is noise. |

The remaining reliability limits are clear:

- **Play installation source cannot distinguish testing tracks.** The lane separates staging-configured builds, but a `prod`-configured binary installed through Play internal testing still qualifies under this rule. That is an operational limitation of the metric definition.
- **Firmware properties are best-effort snapshots.** Versions arriving after the one-shot identification event will be absent. They are not complete firmware diagnostics.
- **Delivery loss and long-lived connections remain Phase 2 work.** This PR alone does not complete WAU restoration.
- **Physical distribution checks remain outstanding.** Helper tests prove the classification logic, not what TestFlight and Play installations actually report.

I read the live discussion and PR-head files through refs; no checkout or file modifications. `git diff --check` passed. I reviewed the reported test evidence but did not rerun native builds or device tests.

PR #3974, opened 2026-09-09
Close Bluetooth SDK analytics coverage gaps: G2 serial, model timing, heartbeat, retry

PhilippeFerreiraDeSousa:
Stacked on #3973 (retarget to `staging` once that merges). Phase 2 of `agents/wau-telemetry-restoration-plan.md`.

## Why

Glasses WAU = distinct manufacturing serials on `bluetooth_sdk_glasses_identified`. Last week's production data (project 130586, 7 d) shows four undercounts:

| Gap | Evidence |
|---|---|
| G2 never reports a serial | 0 / 148 connected G2 installs identified. The code decodes the serial from the advertisement but only stores it in the reconnection name slot. |
| Android connections with no model | 104 connected installs with `glasses_model` empty: the connected flag flips before the model is set. |
| No week-boundary re-identification | Identification fires once per connection. A pair connected across Sunday night is invisible in the next week. |
| Failed uploads are dropped | No retry on either platform. |

## What

- **G2 serial** (Android + iOS): also write the advertisement serial to `glasses.serialNumber` at auth completion. On cached reconnects (address/UUID based, after a process restart) the scan that fills the serial map never runs, so `G2SerialResolution` reuses the serial persisted as the SDK device name at the previous auth completion, only when it equals the id this connection was asked for. Unit-tested on both platforms, including the pair-switch case.
- **Pure tracker** (`BluetoothSdkAnalyticsTracker`, both platforms): all connect/identify/heartbeat decisions in one unit-tested, I/O-free type. The sender only maps `GlassesStatus` to a snapshot and captures what the tracker returns.
  - Connected waits for the model; if the connection ends first it is sent without one and `glasses_model_unresolved=true`.
  - Identification once per connection per serial; `event_kind=glasses_heartbeat` re-emission on the first status update of each new **`America/Los_Angeles` calendar day** (WAU weeks are Monday-start Pacific; a Sunday-evening to Monday-morning Pacific session is one UTC day and would otherwise miss Monday). `count(DISTINCT glasses_device_id)` stays idempotent; the dashboard counts both kinds. Heartbeats ride on status updates; a connection whose status never changes for a day produces none.
- **Retry queue** (`BluetoothSdkAnalyticsQueue`, both platforms): JSON-lines file, 100 entries, 7-day TTL, oldest first. Send outcomes are delivered / retry / discard: 4xx other than 408/429 is dropped so one permanently rejected payload cannot block the rest; network errors, 5xx, 408, 429 stop the drain and retry later. Drained after each successful send and on SDK start. Payloads carry their own `uuid` + capture `timestamp`, so PostHog dedups retries and attributes them to the original time.
- iOS send is synchronous on the transport queue (already off the caller thread) so one request is in flight at a time. The completion handler owns the outcome behind a lock; a timed-out task is cancelled rather than left to write after the return.
- Docs: SDK README, `api-reference.mdx`, `changelogs/3.1.0.md` (+ regenerated catalogs).

## Review round

- Cursor + GitHub Codex + local Codex (Astra): the same cached-reconnect G2 bug; fixed once at the serial-resolution boundary. Local Codex additionally: Pacific-day heartbeat bucketing, iOS transport ownership, retry vs discard. All addressed.
- Second pass: delivery (executor + retry file) is now owned process-wide by `BluetoothSdkAnalyticsTransport` on both platforms, so two overlapping SDK instances cannot race on the retry file; `shutdown()` no longer tears down delivery. Regression-tested on both platforms.
- Known limits kept explicit: heartbeats need a status change; delivery is best effort (a crash before a failed send is persisted loses it).

## Test evidence

- `./scripts/check-android-compile.sh bluetooth-sdk` → BUILD SUCCESSFUL
- Android unit tests (`BluetoothSdkAnalytics*`, `G2SerialResolutionTest`, `-PmentraPublicSdk=true`) → Host 4/4, Tracker 11/11, Queue 5/5, Transport 1/1, G2 3/3
- `xcodebuild -scheme MentraBluetoothSDK … test` (Host, Tracker, Queue, Transport, G2 suites) → 25/25, TEST SUCCEEDED

## Not in this PR

Vuzix Z100 / Mach1 report no serial today; whether one is reachable is a separate question. Hardware verification (Phase 4) still applies: G2 identification on both platforms is the thing to confirm on a real pair.


PhilippeFerreiraDeSousa:

**Good overall design, but I would request changes before merging.** The pure tracker, small bounded retry queue, and fixed event identity/timestamp are sensible. Keep that structure; the problems are specific gaps in identity and delivery semantics, not a reason for a larger analytics framework.

Reviewed head `42e106a` against the requested base and the WAU plan.

Changes I would make, in importance order:

1. **Make G2 identity survive cached reconnects on both platforms.**  
   The new serial write depends on `deviceNameToSerialNumber`, populated during scanning. After process restart, Android’s address reconnect and iOS’s UUID reconnect bypass scanning. Consequently, ordinary returning users can remain completely absent from G2 WAU.

   Resolve the manufacturing serial once for the selected connection, whether discovered or restored, then publish it at auth completion. Reuse the persisted full serial associated with those cached peripherals; don’t blindly treat an arbitrary search string as a manufacturing serial. Add coverage for fresh-process cached reconnects, including a device switch. [Changed G2 code](https://github.com/Mentra-Community/MentraOS/blob/42e106aefbad8872cb59b269659acafdf7be6f0a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/sgcs/G2.kt#L1803)

2. **Align heartbeat bucketing with the WAU calendar.**  
   UTC-day deduplication does **not** guarantee coverage across a Monday-starting Los Angeles week. Example: a heartbeat Sunday at 5 p.m. PDT, continued use Monday at 1 a.m., then disconnection. Both observations fall on the same UTC day, so Monday gets no identification event.

   The simplest correction is to use `America/Los_Angeles` calendar days for heartbeat deduplication, retaining UTC event timestamps. Test the actual Sunday/Monday boundary and DST. This corrects a mismatch in the plan itself; the PR follows its UTC-day instruction faithfully. [Tracker logic](https://github.com/Mentra-Community/MentraOS/blob/42e106aefbad8872cb59b269659acafdf7be6f0a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/BluetoothSdkAnalyticsTracker.kt#L81)

3. **Fix the iOS transport’s timeout/completion ownership.**  
   The semaphore wait times out after six seconds, but its result is ignored and the URLSession task remains running. Its callback can subsequently mutate `succeeded` concurrently with the return path. The sender can also begin another request while the first remains active, contradicting the intended serialized transport.

   Prefer one completion-driven drain with one request in flight. If retaining the synchronous wrapper, synchronize the result and handle timeout/cancellation explicitly. Add a late-completion test. [iOS sender](https://github.com/Mentra-Community/MentraOS/blob/42e106aefbad8872cb59b269659acafdf7be6f0a/mobile/modules/bluetooth-sdk/ios/Source/internal/BluetoothSdkAnalytics.swift#L120)

4. **Distinguish permanent rejection from retryable failure.**  
   Every non-2xx response currently becomes `false`, and draining stops at the first failure. One permanently rejected payload can therefore block otherwise deliverable queued identifications until expiry or eviction.

   A small result type—delivered, retry later, discard—is sufficient. Continue past permanently invalid payloads; stop for transient failures. No elaborate retry scheduler is needed. [Queue drain](https://github.com/Mentra-Community/MentraOS/blob/42e106aefbad8872cb59b269659acafdf7be6f0a/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/BluetoothSdkAnalyticsQueue.kt#L33)

**The existing review comments are worth acting on.** [Cursor’s comment](https://github.com/Mentra-Community/MentraOS/pull/3974#discussion_r3973720080) and [Codex’s comment](https://github.com/Mentra-Community/MentraOS/pull/3974#discussion_r3973721749) describe the same real cached-reconnect bug. Fix it once at the connection-identity boundary. Neither is noise; they are duplicate findings. Codex’s proposed fallback needs the full-serial guarantee described above.

Other reliability limits to keep explicit:

- Heartbeats run on **changed glasses-store values**, not a periodic tick; unchanged updates are suppressed. Quiet connections therefore have no guaranteed heartbeat.
- Model deferral improves connected events, but identification can still fire before the model arrives and is not enriched when it arrives later.
- Delivery remains best effort: termination before a failed request is persisted can lose it. That is acceptable for a bounded analytics implementation, but it is not guaranteed delivery.

I would keep production/build classification in the existing host contract and dashboard filtering. No additional machinery belongs in this PR for that.

No files changed or branches checked out. I inspected the tests; I did not rerun builds or hardware validation.

PhilippeFerreiraDeSousa:
## Local Codex review (gpt-6-astra, medium) — pass 2

_Read-only run of `codex exec` after the first-pass fixes._

**The design is sound, but I would not merge this head yet.** I’d fix one remaining queue-ownership issue, then merge without expanding the scope.

Reviewed `2a3f20ac63` against `621cd9ba54`, including the latest discussion and review comments.

The pure tracker, bounded retry queue, and thin transport are sensible boundaries. The first-pass fixes are well targeted:

- G2’s normal cached reconnect now resolves the persisted serial, with a guard against using a different selected identity.
- Pacific calendar days correctly match the WAU week boundary, with midnight and DST coverage.
- iOS completion state is protected and timed-out requests are cancelled.
- Permanent rejection versus retry is handled consistently on both platforms.

**Before merge: one change, P2 — give the retry file one process-wide transport owner.**

Each analytics instance creates its own serial executor/dispatch queue, but all instances use the same file. Android’s `shutdown()` lets existing tasks continue; SDK creation immediately permits another instance. This can occur during module destruction/recreation, without deliberately running two SDKs.

A concrete failure sequence:

1. Old instance reads pending events and starts draining.
2. New instance persists another failed event.
3. Old instance finishes and rewrites—or deletes—the file using its earlier snapshot.
4. The newly persisted event disappears.

Atomic file replacement does not prevent that lost update. Both [Android’s drain](https://github.com/Mentra-Community/MentraOS/blob/2a3f20ac63/mobile/modules/bluetooth-sdk/android/src/main/java/com/mentra/bluetoothsdk/BluetoothSdkAnalyticsQueue.kt#L50) and [iOS’s drain](https://github.com/Mentra-Community/MentraOS/blob/2a3f20ac63/mobile/modules/bluetooth-sdk/ios/Source/internal/BluetoothSdkAnalyticsQueue.swift#L63) have this structure.

The clean fix is one shared transport executor/queue owning the retry file across SDK instances. Keep connection tracking per instance. Add a regression test covering overlapping drain/enqueue across instance replacement. This establishes a clear ownership rule without adding sleeps, blocking SDK close, or layering locks around individual reads and writes.

**My assessment of the existing comments:**

| Comment | Assessment |
|---|---|
| Cursor/Codex/Cubic: cached G2 reconnect lacks serial | **Valid, addressed for the normal reconnect path.** Resolve the duplicates; retain real-device verification before rollout. |
| Cubic: concurrent SDK instances overwrite retries | **Valid; act before merge**, as above. |
| Cubic: permanent 4xx blocks the queue | **Valid, addressed.** The outcome classification is sufficient. |
| Cubic: use `$insert_id` instead of top-level `uuid` | **Incorrect as stated.** PostHog documents eventual deduplication using matching `uuid`, event, timestamp, and distinct ID; these payloads preserve all four. [PostHog documentation](https://github.com/PostHog/posthog.com/blob/master/contents/docs/data/events.mdx#event-deduplication) |
| Cubic: queued events must precede live events | **Unnecessary for this metric.** Arrival order does not affect distinct serials or capture-time week attribution. |
| Cubic: identification must follow the connected event | **Unnecessary for WAU.** Identification is independently qualifying. Holding a known serial solely to enforce event ordering adds state and potential loss. |
| Cubic: unresolved connection uses disconnected `fully_booted` | **Real diagnostic imprecision, nonblocking.** It does not alter serial WAU. Preserve the pending snapshot if that field needs historical accuracy; don’t introduce a larger event-ordering mechanism. |
| Cubic: read failure is treated as an empty queue | **Valid hardening follow-up.** An unreadable existing file should ideally remain untouched. I would not block this best-effort telemetry PR solely on that rare I/O failure case. |

**Remaining reliability limits:** heartbeats require status updates, retries require another successful send or SDK start, and a crash before persistence can lose an event. These can undercount WAU; they are documented tradeoffs, not reasons to build a background delivery service here. Identification can also precede model/firmware arrival, leaving those diagnostic properties incomplete until a later event.

The host/build properties remain attached to captured payloads through retries, preserving the base PR’s production-versus-development classification.

I reviewed source and tests; `git diff --check` passed. I did not rerun builds/tests in this read-only session or verify G2 hardware. The PR reports passing platform checks. No files were modified.

PhilippeFerreiraDeSousa:
## Local Codex review (gpt-6-astra, medium) — pass 3 (confirmation)

_Read-only run of `codex exec` after the process-wide transport fix._

**I would merge now. No remaining code blockers found at `939048b474`.** The design is appropriately simple: pure connection tracking per SDK instance, one process-wide delivery owner, and a bounded retry file.

The second-pass P2 is addressed correctly. Both platforms serialize delivery and retry-file operations through `BluetoothSdkAnalyticsTransport`; Android shutdown no longer tears that owner down. The new tests exercise an old drain overlapping a new enqueue and verify that the new event survives. This fixes the ownership rule directly.

**Before merge:** no required changes. Two optional follow-ups, ordered by value:

1. **Preserve an unreadable retry file.** Currently, read failure looks like an empty queue, so a subsequent enqueue can overwrite pending events. A distinct read-failure result would prevent avoidable loss. Worth hardening, but nonblocking for this best-effort telemetry.
2. **Preserve the pending connection snapshot if diagnostic accuracy matters.** The unresolved-model fallback uses the disconnected snapshot’s `fully_booted`. That can misdescribe the earlier connection, but does not change serial WAU.

My assessment of the existing comments:

| Comment | Assessment |
|---|---|
| Cursor/Codex/Cubic: G2 cached reconnect misses serial | Valid and addressed; duplicate findings. Keep hardware verification before rollout. |
| Cubic: overlapping instances overwrite retries | Valid and now addressed by the shared transport. |
| Cubic: permanent 4xx blocks retries | Valid and addressed by delivered/retry/discard classification. |
| Cubic: replace `uuid` with `$insert_id` | Noise as stated. PostHog documents eventual deduplication using matching UUID, event, timestamp, and distinct ID; retries preserve these. [Documentation](https://github.com/PostHog/posthog.com/blob/master/contents/docs/data/events.mdx#event-deduplication) |
| Cubic: queued events must precede live events | Unnecessary for distinct-serial, capture-time WAU. |
| Cubic: identification must follow connected | Unnecessary for WAU; delaying a known identity adds state and loss opportunities. |
| Cubic: unreadable queue / disconnected snapshot | Real but nonblocking follow-ups described above. |

The remaining reliability limits are explicit and acceptable for this scope:

- Heartbeats require a status update; a quiet connection can still miss a reporting week.
- Retries need another successful send or SDK start. Process termination before persistence, queue eviction, expiry, or disk failure can lose events.
- Identification can precede model/firmware arrival, leaving diagnostic properties incomplete.
- Host package and build-classification properties remain preserved in queued payloads, supporting the base PR’s production/dev separation.

I reviewed the source, tests, and current GitHub discussion. `git diff --check` passed. The PR reports passing Android and iOS checks; I did not rerun builds/tests or verify G2 hardware in this read-only pass. No files changed or branches checked out.

PR #3975, opened 2026-09-09
Remove the Mentra App's unused PostHog client

PhilippeFerreiraDeSousa:
Phase 3 of `agents/wau-telemetry-restoration-plan.md`. No shared files with #3973 / #3974, so it merges independently. Note that this PR alone does not restore WAU: `app.config.ts` still carries `analytics: false` on staging, and #3973 is what replaces it with SDK analytics plus the build-lane label.

## Why

The app's own `posthog-react-native` integration has never shipped enabled: the key is empty in `.env.example`, `prepare-mobile-release-env.mjs` injects Sentry and Mapbox but never PostHog, and `AllProviders.tsx` drops the provider when the key is missing. The only `Application Opened` events in the prod project are developer builds (unreleased 3.2.0 build numbers) and legacy 2.x.

Glasses usage analytics come from the Bluetooth SDK's embedded PostHog events (the WAU source of truth, re-enabled for this host in #3973). Per-account diagnostics come from the Cloud V2 support profile. A second client adds nothing and has been a source of confusion in Cayden's report.

## What

- Remove the `PostHogProvider` wrapper, `PostHogIdentityBridge`, and `isAnonymousDistinctId` from `AllProviders.tsx`; trim the imports that only they used.
- Remove `posthog-react-native` from `mobile/package.json` (+ lockfile; the diff is PostHog entries only).
- Remove `EXPO_PUBLIC_POSTHOG_API_KEY` from `.env.example` and from `RELEASE_BUNDLE_ENV_KEYS`.
- Remove the Sentry global-handler filter whose only case was a PostHog uuidv7 `RangeError`, and the dev-log filter for `Error while flushing PostHog`.
- Remove the license entry.
- **Guard:** `validateReleaseArchive` now fails if a release JS bundle contains a PostHog project token (`phc_…`). The SDK's token lives in native code, so a match means an app-level client crept back in. It is a tripwire on the validated IPA/APK bundles, not proof of absence: the Play AAB is not passed through this validator today (pre-existing gap).

## Verification

- `eslint` on the three edited TS files: no new findings (the `isProd` unused-var error in `SentrySetup.tsx` predates this PR).
- `node --test mobile/scripts/release-bundle-config.test.mjs` → 6/6, including two new cases for `assertNoAppAnalyticsToken` (clean and binary-wrapped bundles pass, an embedded `phc_…` token fails without leaking the token).
- `grep -ri posthog mobile` (excluding the SDK module and node_modules): no remaining references.
- Local `bun compile` reports one error in `src/services/Migrations.test.ts`, untouched here and caused by duplicate `typesafe-ts` copies on my machine; CI typecheck is authoritative.


PhilippeFerreiraDeSousa:

**Verdict: good design; I would approve the cleanup.** Removing the unused app client, identity bridge, and associated crash workaround reduces complexity. SDK-owned glasses telemetry is the right boundary for serial-based WAU across host apps. I found no blocking runtime bug in commit `d4a10e0`.

Before merge, in importance order:

1. **Make the release dependency on #3973 explicit.** This PR still has `analytics: false` in `mobile/app.config.ts`. [#3973](https://github.com/Mentra-Community/MentraOS/pull/3973) replaces that with SDK analytics and the build-environment label. Independent merging is fine, but #3975 alone does not restore WAU or production/TestFlight separation. Keep that implementation in #3973; duplicating it here would add unnecessary overlap.

2. **Add small automated coverage for the new release guard.** `assertNoAppAnalyticsToken` can block publication, but the committed tests never exercise it. Clean bytes passing and token-containing binary bytes failing are sufficient. No elaborate test framework or native-token allowlist is needed: this scans JavaScript, while the SDK token lives in native code.

3. **Remove the “Generated with Claude Code” footer from the PR description.** The repository explicitly prohibits it. Administrative cleanup only.

**Review comments:** At inspection, GitHub returned **zero inline comments and zero review threads**. The discussion contains a [Cursor low-risk summary](https://github.com/Mentra-Community/MentraOS/pull/3975#issuecomment-5609795706), build status, and a still-running Codex review. There are no concrete review requests to act on or dismiss. I agree with Cursor’s low-risk assessment, but its claim that an app client “can’t slip back in” overstates the guard.

**Correctness and reliability risks:**

- The token scan is a useful tripwire, not proof that no analytics client exists. It detects contiguous `phc_…` strings; a dynamically supplied token could bypass it, and an inert matching string could fail the build. These are limitations to document, not reasons to build a sophisticated scanner.
- Android validation checks the APK; the separately built Play AAB is not passed through this validator. That is an existing validation gap, not a regression here, but the new guarantee should not be described as covering every published artifact.
- Removing the PostHog-specific global error handler is appropriate because its dependency is removed too. I found no remaining app runtime PostHog imports and no change to SDK event identity or transport.

I ran the four existing release-config tests plus clean/binary-token checks against the PR-ref source; all passed. `git diff --check` passed. I did not build release artifacts or verify device telemetry. No files were modified.

PR #3978, opened 2026-09-10
Gate Android releases and the app config on Bluetooth SDK analytics being enabled

PhilippeFerreiraDeSousa:
## What could recur

From 2026-08-11 until #3973, `app.config.ts` passed `analytics: false` to the Bluetooth SDK plugin, so every dev/beta build shipped silent. #3973 added an iOS release gate (Info.plist must carry the SDK version and lane, and must not have analytics disabled). Android had no gate, and nothing checked the config at PR time.

## Guards after this PR

| When | What | Where |
|---|---|---|
| Every PR (mobile script tests) | `expo config --type prebuild` must show the SDK plugin with analytics enabled and a valid lane | `mobile/scripts/app-config-analytics.test.mjs` |
| iOS release | IPA Info.plist: SDK version present, lane matches `EXPO_PUBLIC_BUILD_ENV`, analytics not disabled | `release-ios.mjs` (from #3973) |
| Android release | APK manifest via aapt2: analytics not disabled, lane matches `EXPO_PUBLIC_BUILD_ENV` | `release-android.mjs` → `validateAndroidSdkAnalyticsMetadata` (new) |

## Review round (local Codex, Astra)

P1 found and fixed: build-tools 36 prints boolean meta-data bare (`=true`), which the first parser turned into an empty string the validator accepted. The parser now handles every aapt2 value form, and the validator fails on anything but an explicit `"false"` for the disabled key. Regression tests cover a bare `true` with a correct lane and unrecognized values.

## Verification

- `node --test mobile/scripts/app-config-analytics.test.mjs` → 2/2; with `analytics: false` temporarily reintroduced → 0/2 (the guard fires).
- `node --test mobile/scripts/release-sdk-metadata.test.mjs` → 9/9, including `readApkManifestMetaData` against last night's real CI APK (build 53407874), which carries `com.mentra.bluetoothsdk.analytics.environment=dev`.
- The Android gate needs `aapt2` under `ANDROID_HOME/build-tools`; the coordinated workflow already installs build-tools 36.0.0 and uses aapt/apksigner from there.

Not covered by any gate: a partner's app disabling analytics in its own config, which is their right. This is about our own builds.

### DIFF
--- a/mobile/scripts/app-config-analytics.test.mjs
+++ b/mobile/scripts/app-config-analytics.test.mjs
@@ -9,12 +9,28 @@ import test from "node:test"
 // `expo config`, the same path prebuild uses, so what is asserted is what ships.
 
 function bluetoothSdkPluginProps() {
-  const json = execFileSync("npx", ["expo", "config", "--json", "--type", "prebuild"], {
-    encoding: "utf8",
-    stdio: ["ignore", "pipe", "ignore"],
-    env: {...process.env, EXPO_NO_TELEMETRY: "1", CI: "1"},
-    maxBuffer: 16 * 1024 * 1024,
-  })
+  let json
+  try {
+    json = execFileSync("npx", ["expo", "config", "--json", "--type", "prebuild"], {
+      encoding: "utf8",
+      stdio: ["ignore", "pipe", "pipe"],
+      env: {
+        ...process.env,
+        EXPO_NO_TELEMETRY: "1",
+        // app.config.ts refuses to build without a Mapbox token in CI. The token
+        // has nothing to do with this check, so supply the same placeholder
+        // .env.example ships when the environment has none.
+        EXPO_PUBLIC_MAPBOX_ACCESS_TOKEN: process.env.EXPO_PUBLIC_MAPBOX_ACCESS_TOKEN || "ci-dummy-not-a-real-token",
+      },
+      maxBuffer: 16 * 1024 * 1024,
+    })
+  } catch (error) {
+    throw new Error(
+      `expo config failed: ${String(error.stderr || error.message)
+        .trim()
+        .slice(0, 2000)}`,
+    )
+  }
   const config = JSON.parse(json)
   const entry = (config.plugins ?? []).find(
     (p) => Array.isArray(p) && String(p[0]).includes("modules/bluetooth-sdk/app.plugin"),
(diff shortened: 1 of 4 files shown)


PhilippeFerreiraDeSousa:

**I would not merge now.** The high-level design is good, but the Android gate has one concrete bug that defeats its purpose.

- **[P1: The parser misses actual `aapt2` booleans](https://github.com/Mentra-Community/MentraOS/blob/8729fcdad9ae3813889fdfcb5e6f84368ea3ab8e/mobile/scripts/release-sdk-metadata.mjs#L83).** Build-tools 36.0.0 prints boolean metadata as `android:value(...)=true`. The parser handles quoted strings and `(type 0x12)0x...`, but silently converts bare `true` to `""`. The validator then accepts it. I reproduced this using the PR’s functions and boolean output from a real local APK.
- The test fixture uses **quoted `"true"`**, so it misses this failure. The reported real-APK test checks only that the environment exists; it doesn’t prove disabled analytics are rejected.

The fix should stay small: parse the actual boolean format and add a regression test proving that a correct lane plus bare `true` fails. For these two analytics keys, an unrecognized present value should also fail validation.

Otherwise, I like the approach: resolve Expo configuration at PR time, then validate the built artifact before publication. The existing CI glob already picks up the new test. **No redesign or extra framework is needed.**

I checked comments and inline reviews: there were no substantive review findings yet; Cursor supplied a summary, and Codex review was still running.

Used the requested diff. `gh pr view 3978 --comments` failed to connect, so I retrieved comments through the GitHub connector. Nothing was checked out or modified.

PhilippeFerreiraDeSousa:
## Local Codex review (gpt-6-astra, medium) — pass 2 (confirmation)

**Yes—I would merge #3978 at `e775c3f`, subject to required CI passing. No remaining blocking findings.**

The design is appropriately small: resolve Expo configuration at PR time, then validate the built APK before publication, alongside the existing iOS gate. The existing CI test glob picks up the new config test.

The P1 is fixed properly: bare `true` now parses correctly, and present disabled values must be explicitly `"false"`. Regression coverage proves a correct lane cannot hide disabled analytics. This addresses the failure directly; no redesign or additional framework is warranted.

I ran the seven read-only metadata tests: **7 passed**. Artifact integration tests were skipped; I did not rerun Expo configuration or a release build.

I used the requested diff. `gh pr view 3978 --comments` failed to connect, so I checked comments through the GitHub connector: no additional substantive findings appeared. Nothing was checked out or modified.

PR #3986, opened 2026-09-10
Publish stable packages from the promoted beta after production promotion

PhilippeFerreiraDeSousa:
## Design

**Trigger.** A new manually dispatched workflow, `production-release-packages.yml`, run on `main` with two inputs: `beta_identity` (the promoted beta, e.g. `3.1.0-beta.192`) and `phase` (`publish` | `release`). The CLI wraps it as `scripts/production-release.mjs packages --beta X.Y.Z-beta.N --phase publish|release`.

**Keyed on the frozen beta plan, not on a promotion state.** The `load` job downloads the beta's `release-plan.json` + manifest from `mentra-builds-vX.Y.Z`, re-verifies the beta is complete, checks its `sourceCommit` is an ancestor of `origin/main`, checks out that exact commit into `release-source/`, and derives a deterministic production plan from it (`production-packages.mjs plan`): channel `production`, release identity `X.Y.Z`, the beta's OTA manifest pin as `promotion.otaManifest`. Every reusable job builds from that `sourceCommit`, never from `main` HEAD.

**Which promotion states it may run from.** Any time after `promote` merged the beta into `main`: before `start`, in any live promotion state, or after rollout. It does not need a promotion attempt to exist at all. This matters today: there is no `mentra-production-promotion-v3.1.0-*` container on GitHub (3.1.0 shipped outside the state machine), so keying on `release_identity + attempt` would have blocked 3.1.0 packages on a provenance bootstrap. The promotion chain is **never written** by these phases, so they cannot race a Cloud or mobile transition. It is read once: a live attempt for the identity that froze a different beta or source is a hard stop, and an allocated attempt without a state record fails closed. `nextAction` and every mobile transition are untouched. This satisfies requirement 2: package publication is not a state transition and neither blocks nor is blocked by `stores-submitted → store-review-approved → stores-approved`.

**Staged vs. public, and what a human approves.**

| Phase | Gate | npm | Maven Central | SwiftPM |
| --- | --- | --- | --- | --- |
| `publish` | environment `production-packages` | plain `X.Y.Z` published with provenance under dist-tag `candidate-X.Y.Z` via the unchanged reusable npm workflow (`npm_tag` input); `latest` untouched | `USER_MANAGED` Sonatype deployment; the phase only succeeds once Sonatype reports it `VALIDATED`, nothing published | export committed and pushed to mirror branch `release/X.Y.Z`; no tag |
| `release` | environment `production-packages-release` | preflights every target first (npm dry run, Sonatype `inspect`, SwiftPM archive commit id via `git get-tar-commit-id`), then `promote-npm-latest.mjs` decides for the whole family before moving any `latest` (refuses if `latest` is already newer), reads back, retires the candidate tag; needs `NPM_TOKEN` because OIDC only covers `npm publish` | `sonatype-central-deployment.mjs publish` on the persisted record, waits for `PUBLISHED` | tag `X.Y.Z` pushed at the staged commit; mirror `main` fast-forwarded only when possible |

Current npm behaviour was kept: `publish-release-npm.mjs` does a direct `npm publish --provenance` (no `npm stage publish` in this code path), so the candidate publish behaves exactly like a beta publish and inherits the #3983 read-back wait and 120-minute job budget.

**Evidence.** Both phases write content-addressed JSON into the stable release `mentra-vX.Y.Z` (allocated with the same payload `production-release-rollout.yml` uses, asserted by a contract test so the two cannot drift; a container already published by hand is accepted). The native SDK job already persists the Sonatype record and SwiftPM archive there.

## Implementation

- `production-release-packages.yml` (new): `load` → `approve` → reusable `npm` + `sdk-native` → `record-publication`; or `load` → `release`.
- `reusable-coordinated-sdk-native.yml`: Sonatype tooling now runs from the workflow revision (`github.sha`) rather than the frozen package source, so a beta cut before this change still stages correctly; on the production channel the Maven upload is `USER_MANAGED`, verified and awaited to `VALIDATED`, and the SwiftPM job pushes `release/X.Y.Z` instead of `main` + tag (idempotent on rerun via the staged branch). Dev/beta paths unchanged.
- `production-promotion-state.mjs`: read-only `requirePromotionMatchesPackages` guard and CLI `packages-guard`; no chain changes.
- `production-promotion-assets.mjs`: `latestPromotionContainer` + CLI `download-latest-attempt` (fails closed on an allocated attempt without a state record). Also fixes a pre-existing defect: listing releases through `gh api` overflowed Node's 1 MiB subprocess buffer once the build containers carried ~700 assets, which would have broken every promotion command that lists releases; listings now project only the needed fields and stream as JSON lines.
- `production-packages.mjs` (new): plan derivation, candidate tag, stable container allocate/reuse.
- `promote-npm-latest.mjs` (new): latest flip with downgrade protection and read-back.
- `sonatype-central-deployment.mjs`: `--publishing-type` (AUTOMATIC | USER_MANAGED), recorded on the persisted deployment, plus `wait-validated`.
- `prepare-production-promotion.mjs`: shared `validateSelectedBeta` (no behaviour change).
- `scripts/production-release.mjs`: `packages` command.
- Runbook: new "Stable packages" section and two environment rows.

Cloud and mobile promotion behaviour is unchanged.

## Test evidence

```
node --test .github/scripts/*.test.mjs scripts/production-release.test.mjs \
  mobile/modules/bluetooth-sdk/scripts/*.test.mjs mobile/modules/engine/scripts/*.test.mjs \
  mobile/scripts/app-store-connect-build.test.mjs
# tests 241  # pass 241  # fail 0        (staging baseline: 165 in .github/scripts)
node .github/scripts/release-family-cli.mjs validate --require-version-mirrors   # OK
actionlint on both workflows: only the pre-existing `queue: max` warning
```

New coverage: the read-only promotion guard across every state and its beta/source mismatch failures, latest-attempt lookup, deterministic plan derivation and beta validation failures, stable-container reuse/mismatch, latest-flip decisions, the full fake-registry flow and the no-mutation-on-partial-failure case, USER_MANAGED upload and `wait-validated`, release-phase preflight ordering, workflow contract for the new file, and the CLI command.

## Before first use

- Create the `production-packages` and `production-packages-release` environments (required reviewers).
- The release phase needs `NPM_TOKEN`, `MAVEN_CENTRAL_TOKEN_BASE64`, and `MENTRA_BLUETOOTH_SDK_IOS_PUSH_TOKEN` as repository secrets (the reusable jobs get them via `secrets: inherit`).
- If npm trusted publishing is scoped per workflow file, register `production-release-packages.yml` for each family package; otherwise `NPM_TOKEN` covers the publish.
- Once phase `publish` has put `X.Y.Z` on npm, that identity is spent; an aborted promotion cannot be replaced by a different source under the same version (documented as a stop condition).

## Review

Reviewed locally by Codex (gpt-6-astra, medium reasoning), posting as the `mentra-release-coordinator` app. Pass 1 requested changes (partial npm flips, Sonatype validation timing, frozen-source tooling, chain races, published container) and pass 3 requested changes for the no-attempt guard path; both reworks are design changes rather than local patches, smoke-tested against the real repository. Final formal review: **approved at `90081cfa88`**.

### DIFF
--- a/.github/production-release/README.md
+++ b/.github/production-release/README.md
@@ -403,8 +403,8 @@ SDK jobs as the coordinated beta on the production channel:
   `candidate-X.Y.Z`; `latest` is untouched. Publication uses provenance, so if
   npm trusted publishing is scoped per workflow file, register
   `production-release-packages.yml` for each package or keep `NPM_TOKEN` set.
-- Maven Central receives a `USER_MANAGED` Sonatype deployment. It is validated
-  but nothing is public until phase 2.
+- Maven Central receives a `USER_MANAGED` Sonatype deployment. The phase only
+  succeeds once Sonatype reports it validated; nothing is public until phase 2.
 - The SwiftPM export is committed and pushed to the mirror branch
   `release/X.Y.Z` of `mentra-bluetooth-sdk-ios`; no tag is created.
 
@@ -414,23 +414,24 @@ Phase 2 makes them public after `production-packages-release` approval:
 ./scripts/production-release.mjs packages --beta X.Y.Z-beta.N --phase release
 ```
 
-It moves npm `latest` to `X.Y.Z` for every member (refusing if `latest`
-already points at a newer version), retires the candidate dist-tag, requests
-the Sonatype publication and waits for `PUBLISHED`, and pushes the SwiftPM tag
-`X.Y.Z` at the staged commit after verifying it matches the archived export.
-Moving a dist-tag requires the `NPM_TOKEN` automation secret; trusted-publisher
-OIDC only covers `npm publish`.
+It first checks all three targets without changing anything: every npm
+member is published and its `latest` is not already newer, the Sonatype
+deployment is validated, and the staged SwiftPM commit is the one recorded in
+the archived export. Only then does it move npm `latest` to `X.Y.Z` for every
+member and retire the candidate dist-tag, request the Sonatype publication and
+wait for `PUBLISHED`, and push the SwiftPM tag `X.Y.Z`. Moving a dist-tag
+requires the `NPM_TOKEN` automation secret; trusted-publisher OIDC only covers
+`npm publish`.
 
 Both phases are idempotent: a rerun reuses versions, deployments, and mirror
 commits that already exist and refuses anything that exists with different
-bytes. Evidence is stored as content-addressed assets in the stable draft
-release `mentra-vX.Y.Z` (the same draft the rollout finalization stages the
-canonical records into). When a live promotion attempt for `X.Y.Z` exists and
-selected the same beta, the evidence is also appended to its chain as an
-additive `production-packages-publication` or `production-packages-release`
-record without changing its state. A promotion at the `finalizing` checkpoint,
-or one that is completed or aborted, keeps the evidence in the stable draft
-only. A live attempt that froze a different beta or source is a hard stop.
+bytes. Evidence is stored as content-addressed assets in the stable release
+`mentra-vX.Y.Z` (the same draft the rollout finalization stages the canonical
+records into; it is also accepted after that release has been published by
+hand). The promotion chain is never written by these phases, so they cannot
+race a Cloud or mobile transition. It is read once: a live attempt for `X.Y.Z`
+that froze a different beta or source is a hard stop, and an allocated attempt
+without a state record must be resumed or aborted first.
 
 Stop conditions specific to packages:
 
--- a/.github/production-release/README.md
+++ b/.github/production-release/README.md
@@ -65,6 +65,8 @@ Configure these GitHub environments:
 | `production-mobile-candidates` | Production-signed candidate uploads | Required reviewer; no public release       |
 | `production-store-submission`  | App review submission               | Required reviewer                          |
 | `production-store-release`     | Public release and rollout evidence | Required reviewer different from initiator |
+| `production-packages`          | Stage plain package versions        | Required reviewer                          |
+| `production-packages-release`  | npm latest, Maven Central, SwiftPM  | Required reviewer different from initiator |
 
 Required secrets are the existing Porter, App Store Connect, Google Play,
 Android upload-signing, Apple Match, Doppler, Mapbox, and Sentry credentials
@@ -113,6 +115,7 @@ git pull --ff-only origin main
 `status` is the source of truth for the current state and next action. Use
 `--json` for machine-readable output. Use `--attempt N` when inspecting an older
 attempt. Use `status --refresh` to dispatch the read-only store status workflow.
+Stable package publication is not part of `next`; see "Stable packages" below.
 
 Mutating commands require typing the release identity, or `--yes` in an already
 reviewed non-interactive procedure. The CLI never reads production credentials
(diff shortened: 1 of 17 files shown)


PhilippeFerreiraDeSousa (commented):
Reviewed locally by Codex (gpt-6-astra, medium reasoning).

Request changes. The independent, frozen-beta package workflow is the right direction, but the defects below prevent reliable publication and can disrupt mobile promotion.

Checked: PR description, full origin/staging...HEAD diff at b2bd911, relevant complete scripts/workflows/runbook, and all 17 inline review comments plus review summaries. `node --test .github/scripts/*.test.mjs scripts/production-release.test.mjs`: 188 passed, 0 failed. In-memory probes reproduced partial npm mutation and duplicate-sequence rejection. No files changed; no production workflows dispatched.

Valid existing comments and required changes:
- #3983598732 / #3983649308: package appends race promotion transitions (`.github/workflows/production-release-packages.yml:40,322-365,552-595`). Different state filenames permit two records at the same sequence, breaking subsequent mobile/Cloud loads. Prefer keeping package evidence solely in the stable container, without modifying the promotion chain. If retaining appends, serialize the entire read/append/write operation with promotion writers without holding that lock during package builds or approvals.
- #3983621707: Maven executes the uploader from the frozen beta checkout (`.github/workflows/reusable-coordinated-sdk-native.yml:96-99,349-354`). The pre-PR uploader silently ignores the new flag and selects AUTOMATIC. Execute publication tooling from the pinned workflow revision, separately from frozen package sources; verify USER_MANAGED for production and cover a pre-change beta.
- #3983598740 / #3983649302 and #3983598745 / #3983649340: preflight all npm members before mutations (`.github/scripts/promote-npm-latest.mjs:93-118`), and wait for Sonatype VALIDATED before staging succeeds (native workflow:349-403). These are real partial-release defects, although the npm comment overstates this as an individual-package downgrade: the demonstrated failure leaves mixed family versions.
- #3983598752 / #3983649378: published GitHub releases are rejected (`.github/scripts/production-packages.mjs:84`; native workflow:79). Accept matching published containers throughout the package path to preserve the documented ability to publish packages after rollout, without changing mobile release order.
- #3983649317: an allocated but empty attempt is not absent (`.github/scripts/production-promotion-assets.mjs:297`); its body already reserves a selection. Validate that reservation or fail explicitly instead of skipping the mismatch guard.
- #3983649356: package concurrency uses beta identity despite sharing stable coordinates (package workflow:40). Serialize by stable identity, or use one package-only concurrency group.

Additional finding: release changes npm latest before even checking that the Maven record and SwiftPM archive/commit exist (package workflow:416-480). A partially failed publish phase can therefore expose npm before discovering missing native outputs. Require complete, source-matching staging evidence and preflight all three targets before the first public mutation; add failure-path coverage. Cross-registry atomicity is unnecessary, but predictable precondition failures must precede publication.

Comments I would not block on:
- #3983621721 / #3983649327: cross-runner tar differences are a theoretical portability concern, not a demonstrated failure. Both commands archive the same commit; [Git documents commit-derived timestamps](https://git-scm.com/docs/git-archive). Checking the archive before public mutations addresses the immediate ordering risk; normalize contents only if needed.
- #3983649371: container allocation races are a real but retryable availability edge case; re-read after conflict is reasonable, not a separate design blocker.
- #3983649349: confirming the base identity is consistent with the runbook; exact beta confirmation is optional UX polish.
- #3983649363 / #3983649383: strict prose-body equality and duplicate-import cleanup are noise for release correctness.

Preserve the separate package approvals, exact beta sourceCommit/OTA pin, plain npm versions, stable Maven coordinates and SwiftPM tag, and existing Cloud/mobile transitions.

PhilippeFerreiraDeSousa (commented):
Approve.
Reviewed locally by Codex (gpt-6-astra, medium reasoning) at 91c6737972.

Checked the previous review, rework commit, full diff against origin/staging, release scripts/workflows/runbook, and all 17 existing inline comments. `node --test .github/scripts/*.test.mjs scripts/production-release.test.mjs`: 189 passed, 0 failed. No files edited or production workflows dispatched.

The design now keeps package evidence in the stable container and leaves Cloud/mobile transitions unchanged. Package approvals are independent of store review; builds use the beta's frozen sourceCommit and OTA pin, producing plain X.Y.Z npm versions, stable Maven coordinates, and the X.Y.Z SwiftPM tag.

Previous required changes are resolved:
- Chain races and shared-coordinate concurrency (#3983598732, #3983649308, #3983649356): no package chain writes; one package lock (`.github/workflows/production-release-packages.yml:42,310`; `.github/scripts/production-promotion-state.mjs:354`).
- Frozen-beta uploader (#3983621707): Sonatype tooling comes from github.sha separately from package sources, with publishing-type verification (`.github/workflows/reusable-coordinated-sdk-native.yml:110,362,368`).
- npm family preflight (#3983598740, #3983649302): all decisions precede mutations (`.github/scripts/promote-npm-latest.mjs:95-109`). The original individual-package downgrade claim was overstated; mixed family versions were the actual defect.
- Sonatype validation (#3983598745, #3983649340): new/resumed production deployments must validate before staging succeeds (`.github/workflows/reusable-coordinated-sdk-native.yml:395-404`).
- Published containers (#3983598752, #3983649378) and empty attempts (#3983649317): accepted matching published releases; allocated attempts without state fail closed (`.github/scripts/production-packages.mjs:82-94`; `.github/scripts/production-promotion-assets.mjs:297-303`).
- My additional release-ordering finding: npm, Maven, and SwiftPM preflights now precede the first public mutation (`.github/workflows/production-release-packages.yml:364-449`).

Other comments:
- #3983621721/#3983649327 are addressed by checking the archive's commit ID before publication (`production-release-packages.yml:430-433`).
- #3983649371 is addressed by re-reading after container allocation failure (`production-packages.mjs:124-130`).
- #3983649383's duplicate import is removed.
- I would ignore #3983649349 and #3983649363 as blockers: confirming X.Y.Z follows the documented CLI convention, and prose-body equality adds no source-identity protection. Leaving those unchanged is reasonable.

No existing blocking comment remains valid at this head. One non-blocking cleanup: remove the obsolete append-design comment/unused constants at `.github/scripts/production-promotion-state.mjs:82-90`, and refresh the PR description's stale append/cmp claims. No further required changes. External registry publication remains untested locally.

PR #4000, opened 2026-09-11
Let the first coordinated promotion freeze the store-observed Mentra App

PhilippeFerreiraDeSousa:
## Why

`start --beta 3.1.0-beta.205` failed in `production-release-prepare.yml` at "Download current production provenance": no published `mentra-vX.Y.Z` release exists because this is the first coordinated promotion. The runbook asked for a one-time hand-written provenance bootstrap, but that cannot describe the legacy 3.0 release honestly:

- the legacy release scripts derived build numbers from the wall clock separately per platform, so every 3.0 build has a different iOS build number and Android versionCode, while `validateCurrentMentraApp` requires them to be one integer;
- the compatibility lab (Phase 2) rebuilds the current app from its source commit with the coordinated mobile workflow, and the commits that produced 3.0 predate `.github/release-family.json` and the coordinated scripts.

## What

When no published `mentra-vX.Y.Z` release exists, preparation now:

- freezes the public Mentra App exactly as both stores serve it: `coordinates.currentMentraApp` gets `provenance: "store-observed"`, `sourceCommit: null`, `provenanceUrl: null`, the App Store marketing version and build number for iOS, and the Play production version code for Android;
- sets `coordinates.compatibilityLab` to `null` and allocates only one new build number (the candidate);
- creates the initial record directly in `staging-compatible`, so `next` goes to the Cloud preflight. There is no source to build a lab app from, so Phase 2 does not exist for the attempt;
- publishes the store inventory into the attempt container as `current-production-store-inventory.json` (in both modes) and skips the frozen `current-production-release-*.json` assets when there is nothing to freeze;
- fingerprints the selection without the absent previous release.

Records with a coordinated current app are unchanged apart from the explicit `provenance: "coordinated"` field. The state machine rejects a store-observed record in `selected`, a store-observed app carrying a commit or lab coordinates, and a coordinated app without lab coordinates. Phase 5 (`production-mobile-n-compatibility`) still attests the real store app against production Cloud N+1 using the frozen store coordinates. The next promotion finds the `mentra-vX.Y.Z` release this one publishes and runs Phase 2 as before.

Targets `main` because the promotion workflows and scripts run from `main`; it must be reconciled into `staging` and `dev` afterwards.

## Test evidence

- `node --test` over `.github/scripts/*.test.mjs`, `mobile/scripts/app-store-connect-build.test.mjs`, `scripts/production-release.test.mjs`: 237 passed, 0 failed.
- Dry run of `prepare-production-promotion.mjs` with beta.205-shaped inputs, a store inventory of the real 3.0 numbers (iOS 51180073, Android 51180031), and no `--previous-manifest`: produces `production-promotion-3.1.0-attempt-1-00-staging-compatible.json`; `production-promotion-state.mjs validate` accepts it and `next` returns the `production-release-cloud.yml` preflight.
- Edited workflow parses as YAML (16 steps).

### DIFF
--- a/.github/production-release/README.md
+++ b/.github/production-release/README.md
@@ -56,7 +56,8 @@ The required order is:
 
 Mobile N+1 compatibility with Cloud N is not a normal gate. Customers will
 inevitably keep Mobile N after Cloud N+1 is deployed, so Mobile N with Cloud N+1
-is mandatory.
+is mandatory. Step 1 needs the current app's source provenance; the first
+coordinated promotion has none and keeps only step 3 (see "Phase 1").
 
 ## One-time repository and account setup
 
@@ -155,8 +156,9 @@ release this one publishes and runs Phase 2 normally.
 If preparation stops before `status` can find an initial state record, rerun
 `start` with the same beta. That interrupted bootstrap may leave an empty draft
 attempt, but it has not deployed Cloud, uploaded an app, or consumed a store
-build coordinate. Once `status` returns `selected`, resume that attempt with
-`next` rather than starting another one.
+build coordinate. Once `status` returns `selected` (or `staging-compatible` for
+a first coordinated promotion), resume that attempt with `next` rather than
+starting another one.
 
 ## Phase 2 - Mobile N against staging Cloud N+1
 
--- a/.github/production-release/README.md
+++ b/.github/production-release/README.md
@@ -138,9 +138,19 @@ Preparation fails if:
 - its source is not contained in `main`;
 - the public Mentra App does not match the previous production manifest;
 - store build numbers cannot be allocated monotonically; or
-- the current production release lacks provenance. For a one-time provenance
-  bootstrap, create and review an accurate immutable current-production record;
-  do not invent coordinates in the workflow.
+- a published `mentra-vX.Y.Z` release exists but its record does not describe
+  the public Mentra App.
+
+When no published `mentra-vX.Y.Z` release exists yet (the first coordinated
+promotion), the public Mentra App predates this system and cannot be rebuilt
+from provenance. Preparation then freezes the app exactly as both stores serve
+it (`currentMentraApp.provenance` is `store-observed`, with the App Store build
+number and the Play production version code), publishes the store inventory
+into the attempt container, and creates the initial record directly in
+`staging-compatible`: Phase 2 does not exist for that attempt because there is
+no source to build a lab app from. Phase 5 still verifies the real store app
+against production Cloud N+1. Every later promotion finds the `mentra-vX.Y.Z`
+release this one publishes and runs Phase 2 normally.
 
 If preparation stops before `status` can find an initial state record, rerun
 `start` with the same beta. That interrupted bootstrap may leave an empty draft
@@ -149,6 +159,10 @@ build coordinate. Once `status` returns `selected`, resume that attempt with
 `next` rather than starting another one.
 
 ## Phase 2 - Mobile N against staging Cloud N+1
+
+This phase only exists when the current production app has coordinated
+provenance. `status` shows a first coordinated promotion already in
+`staging-compatible`; continue with Phase 3.
 
 Run the next action and watch it to completion:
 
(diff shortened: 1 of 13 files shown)


PhilippeFerreiraDeSousa:
Review verdict: REQUEST CHANGES for head 4ef54df13b641a3099301ed721448f2ab206d178.

GitHub rejected `gh pr review --request-changes` because the authenticated account owns this PR. This comment records the verdict; it is not a formal submitted review.

Defects:

1. **[P1] First-promotion finalization still requires nonexistent Phase 2 evidence.** The new initial state in `.github/scripts/production-promotion-state.mjs:323-326` skips the lab, but `.github/scripts/finalize-production-promotion.mjs:50-64` unconditionally requires `staging-mobile-n-compatibility-lab` and `staging-mobile-n-compatibility`. I reproduced a valid store-observed chain through all subsequent transitions to `finalizing` (sequence 10). The finalizer throws `Cannot finalize production promotion: promotion chain is missing staging-mobile-n-compatibility-lab evidence`. This occurs after the durable 100% rollout checkpoint, preventing the canonical manifest and completed state; retrying cannot fix it. Use the shared provenance predicate to require the two staging evidence kinds only for coordinated current apps, keep every later gate, and cover store-observed finalization. Do not fabricate lab evidence or weaken coordinated validation.

2. **[P2] A bounded release search cannot establish first-promotion eligibility.** `.github/workflows/production-release-prepare.yml:87-96` searches only 100 non-draft/non-prerelease releases, then treats no matching tag as proof no coordinated release exists. With 100 newer unrelated stable releases and an older `mentra-v3.1.0`, the exact jq query returns null on the limited list and finds the production release on the complete list. The new fallback would silently classify a coordinated app as store-observed and skip Phase 2; previously this failed closed. Paginate release discovery before allowing this exception and test the case. This is a future-history failure: the current filtered listing has 47 releases.

Checked:

- PR description and full diff against main; requested state/preparation/assets scripts, prepare/lab/attest workflows, and complete runbook.
- Cloud preflight/deploy, production Mobile N attestation, candidate builds, store-submit, store-release, rollout, finalization, and CLI/store status. Source/candidate lookups tolerate null old-app source/lab coordinates; the finalizer evidence list is the exception above.
- Selection fingerprints, conditional previous assets, inventory publication, contiguous transitions, frozen fields, and append-only evidence. Provenance/coordinates cannot change within a valid chain. Coordinated preparation still starts at selected and requires lab build evidence followed by staging acceptance.
- Shell: gh currently returns a newline for no match; command substitution strips it and the empty-or-null guard works. Empty array expansion under nounset fails on this Mac's Bash 3.2, but is supported by the modern Bash on this Ubuntu workflow runner. See [runner inventory](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) and [Bash manual](https://www.gnu.org/s/bash/manual/html_node/The-Set-Builtin.html). This is not a workflow defect.
- Docs describe the intended first-promotion exception, but completion and subsequent normal promotion are blocked by the defects above. Minor wording: the top-level required release order still presents Phase 2 unconditionally, and the resume paragraph only mentions selected, not staging-compatible.
- `node --test .github/scripts/*.test.mjs mobile/scripts/app-store-connect-build.test.mjs scripts/production-release.test.mjs`, with shell-expanded explicit file arguments: **237 passed, 0 failed** on local Node 20.20.1. The additional finalization reproduction ran in memory with no file edits.

Existing comments assessed:

- Paginated review-comments/reviews endpoints and GraphQL review threads were empty when rechecked. No inline findings or cubic review were available to agree/disagree with.
- Agree with Cursor Bugbot's issue-level summary of preparation, state selection, build allocation, and production Mobile N attestation. Its test/validation reassurance does not establish end-to-end readiness: finalization fails despite the passing suite.
- Codex connector's issue comment reports Running and contains no technical findings or approval to accept/reject.
- GitHub Actions' helper reports waiting builds and a checkout command; it contains no substantive code-review conclusion.

No repository files edited or pushed, and no production workflows dispatched.

PR #4020, opened 2026-09-12
Map glasses OTA failure codes to translated copy on the failed screen

PhilippeFerreiraDeSousa:
## Scope

The Mentra Live OTA failed screen rendered the raw error string from the glasses. `getOtaErrorMessage` in the engine passed any code outside its switch straight through, so a failed downgrade handoff showed the literal `downgrade_handoff_failed` (seen on a Galaxy Z Fold on 2026-09-11).

This PR:

- Keeps the English copy for every glasses-reported OTA failure code in one table in the engine's `OtaErrorMapping.ts`, keyed by i18n copy keys (`ota:errorDowngradeHandoffFailed` etc.). The known codes are the ones the ASG client's `OtaHelper` sends via `sendProgressToPhone(..., "FAILED", errorCode)`: the download classification codes, `install_failed`, the verify codes, and the three downgrade codes (`downgrade_handoff_refused`, `downgrade_handoff_failed`, `downgrade_transaction_stalled`).
- Exposes `copyKey` and `glassesCode` on the hook's `MentraLiveOtaError` state alongside the existing English `message`. Precedence is unchanged: BES restart instruction, then phone-side watchdog/preflight text (English-only, no key), then the glasses code mapped to copy.
- Has the flow screen translate the copy key through the host's `translate` prop and print the raw code in a subdued secondary line (`Error code: downgrade_handoff_failed`, testID `ota-error-code`) so support can still identify it. The screen title now goes through `ota:updateFailed` too.
- Falls back to generic glasses-error copy for unknown codes instead of echoing the code. `getOtaErrorMessage` keeps its signature but no longer returns the raw code for unmapped input.
- For `downgrade_handoff_failed` the message says the recovery service on the glasses did not respond and suggests restarting the glasses and retrying.
- Mirrors the new keys in the Mentra App's `ota` i18n namespace in `mobile/src/i18n/en.ts` (other locales fall back to English).

## Review rework (second commit)

- `copyKey` and `glassesCode` are optional on `MentraLiveOtaError` so host code that builds the type by hand keeps compiling; the hook always populates them.
- `apk_restart_guard_not_persisted` (OtaHelper, APK restart guard could not be persisted) is now mapped with its own copy.
- Code lookups use own properties, so a code naming an inherited Object member (`constructor`, `__proto__`) is unknown rather than a mapped key. The engine's default translator is hardened the same way.
- A producer-coverage test scans the ASG client sources (literal `FAILED` codes, downgrade watchdog codes, `classifyDownloadError` returns, `FirmwareDownloadException` constants, `AsgConstants` OTA codes) and asserts every code is mapped with English copy.

## Test evidence

Jest (`mobile`): `otaErrorMapping.test.ts`, `ota/__tests__/progress.test.tsx`, `ota/__tests__/shared-flow.test.tsx`, `otaInstallCoordinator.test.ts`: 4 suites, 163 tests passed. New cases cover the downgrade handoff copy, the raw-code secondary line, the unknown-code fallback, and that a phone-side watchdog message shows no code line.

Bun (`mobile/modules/engine`): `useMentraLiveOta.test.tsx`: 14 tests passed, including a new case asserting `copyKey` and `glassesCode` for a glasses failure code.

`tsc --noEmit` clean for `mobile` and for the engine module, and the engine's `test:public-ota-consumer` fixture still compiles.

### DIFF
--- a/mobile/modules/engine/src/index.ts
+++ b/mobile/modules/engine/src/index.ts
@@ -119,7 +119,13 @@ export {
 export type {OtaAutoChainAdvanceResult} from "./services/OtaAutoChain"
 export {
   BES_INSTALL_RESTART_MESSAGE,
+  OTA_ERROR_BES_RESTART_REQUIRED_COPY_KEY,
+  OTA_ERROR_ENGLISH_COPY,
+  OTA_ERROR_GENERIC_COPY_KEY,
+  OTA_ERROR_UNKNOWN_GLASSES_COPY_KEY,
+  OTA_GLASSES_ERROR_COPY_KEYS,
   getOtaErrorMessage,
+  otaErrorCopyKey,
   shouldRequireGlassesRebootForBesFailure,
   shouldShowChangeWifiForOtaDownloadFailure,
 } from "./services/OtaErrorMapping"
--- a/mobile/modules/engine/src/react/MentraLiveOtaFlow.tsx
+++ b/mobile/modules/engine/src/react/MentraLiveOtaFlow.tsx
@@ -14,10 +14,12 @@ import {useMarkdown, type MarkedStyles, type useMarkdownHookOptions} from "react
 import {SafeAreaView} from "react-native-safe-area-context"
 import Svg, {Path, Rect} from "react-native-svg"
 
+import {OTA_ERROR_ENGLISH_COPY} from "../services/OtaErrorMapping"
 import {
   MINIMUM_OTA_BATTERY_LEVEL,
   useMentraLiveOta,
   type MentraLiveOtaController,
+  type MentraLiveOtaError,
   type MentraLiveOtaFlowPage,
 } from "./useMentraLiveOta"
 
@@ -142,6 +144,8 @@ const ENGLISH_COPY: Record<string, string> = {
   "ota:versionChangeFirmwarePassComplete": "Firmware updated",
   "ota:versionChangeFirmwarePassCompleteMessage":
     "Your glasses restarted with new firmware. One more step: they'll now continue to the required version.",
+  "ota:updateFailed": "Update Failed",
+  ...OTA_ERROR_ENGLISH_COPY,
 }
 
 const componentCopyKey = {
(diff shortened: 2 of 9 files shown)


PhilippeFerreiraDeSousa (commented):
Request changes.

Reviewed head 3aa47ca92f0ed34d23c7b52f0cace4588a60ec85 against dev.

Concrete defects:
1. [P2] Preserve source compatibility of the exported MentraLiveOtaError type (mobile/modules/engine/src/react/useMentraLiveOta.ts:57-72). Both new fields are required. A previously valid host value, const error: MentraLiveOtaError = {code: "install_failed", message: "Install failed"}, now fails TypeScript with missing copyKey and glassesCode. Make the added metadata optional while continuing to populate it in the hook. The public consumer fixture passes but never constructs this error type, so it does not cover this compatibility case.
2. [P2] Complete the known-code table (mobile/modules/engine/src/services/OtaErrorMapping.ts:15-28). OtaHelper.java:1044-1046 already emits apk_restart_guard_not_persisted when persisting the APK restart guard fails; it is present in the PR base too. This code is missing and currently receives the unknown-error copy. Add an explicit mapping and corresponding English/app copy (or deliberately share appropriate existing copy). Add producer-coverage validation rather than only iterating entries already present in the table. insufficient_storage is correctly retained.
3. [P2] Make unknown-code lookup use own entries (OtaErrorMapping.ts:62-65). The normal object lookup treats inherited names as mapped codes. Reproduced: otaErrorCopyKey("constructor") returns Object, and getOtaErrorMessage("constructor"), getOtaErrorMessage("toString"), and getOtaErrorMessage("__proto__") return undefined instead of the promised generic string. This leaves default failed-screen copy empty and can pass a non-string to a host translator. An own-property check or Map fixes the shared lookup without special-case patches.

Checks completed:
- Read the full PR diff/title/body, surrounding hook/flow/coordinator/host code, ASG failure producers, and all currently available comments and reviews. Cursor's summary is broadly accurate about the normal screen path but misses the defects above. Codex has only a running-status comment for an older head; there is no completed finding to endorse. Both inline-comment and review APIs are empty; no cubic review is currently present.
- Traced the primary failed-screen message: known/ordinary unknown glasses codes are mapped; raw code stays on the secondary support line. Coordinator errorMsg producers use phone-side messages, so they do not bypass mapping with raw glasses codes.
- Compared the previous precedence: BES restart instruction still outranks phone watchdog/preflight text, which outranks mapped glasses codes. Restart/change-WiFi policy helpers are unchanged.
- Compared all OTA_ERROR_ENGLISH_COPY entries with the app's en.ts: zero mismatches; the flow spreads that table into ENGLISH_COPY.
- Checked public exports and getOtaErrorMessage: existing call signatures remain accepted (null is additionally accepted); unknown-string pass-through is intentionally changed, so hosts relying on that behavior will receive generic copy. Existing known-code English messages are retained.
- Confirmed progress tests render the real route/shared flow and actual mapping/coordinator through the test setup, not a mocked replacement screen.
- Ran four Jest suites (mapping, rendered progress, shared flow, coordinator): 158 tests passed. Engine hook Bun suite: 14 tests passed. Public OTA consumer TypeScript fixture passed. Independently reproduced the old host error construction failure with an in-memory TypeScript fixture and the inherited-key fallback failure.

No source files edited or commits pushed.

PR #4021, opened 2026-09-12
Narrate the downgrade detour as a reinstall, not a verification

PhilippeFerreiraDeSousa:
## Scope

Copy and phase-mapping fix for the Mentra Live version-change (downgrade) detour in the OTA flow. No protocol changes; the recovery worker still reports no progress.

**Observed on hardware (2026-09-11):** after ASG hands the downgrade to `com.mentra.recovery`, the worker uninstalls and reinstalls ASG silently while the BES keeps the BLE link up. `deriveVersionChangePhase` split the detour on `connected`, so the phone showed a spinner with **"Verifying your glasses…"** for the whole operation (30–60 s in the good case, up to the 180 s watchdog in the bad case). Nothing was being verified yet.

## Change

`OtaInstallCoordinator`:
- The detour wait (install latched, not converged) is now always the **`reinstalling`** phase, independent of link state.
- **`verifying`** is reserved for the window between a *physical* BLE reconnect inside the detour and the build number the phone then checks. It is cleared by the next reported build number (target → converged, passive factory build → back to `reinstalling`), by a further disconnect, and when the detour is re-latched after a retry.
- `glasses_session_changed` deliberately does not enter `verifying`: the SDK bridge detects the sid change inside the same `version_info` message that already applied the build number (`MentraLive.kt` `handleGlassesSessionId` is called from the version_info parsing path), so the check has already run by the time that signal arrives, and latching on it would leave "Verifying" stuck on for the target install that follows a factory-build flicker.
- `VersionChangePhase` is exported alongside `OtaInstallSnapshot`.

`useMentraLiveOta` / `MentraLiveOtaFlow`:
- Public screen union gains `reinstalling` (fixture `public-ota-consumer.tsx` updated). The host firmware-restarting overlay keeps its previous trigger (detour wait while the link is actually down).
- Copy (engine defaults and host `en.ts`; no other locale carries these keys):
  - `ota:versionChangeReinstalling`: "Reinstalling glasses software…"
  - `ota:versionChangeReinstallingMessage`: "This can take a few minutes and your glasses will restart on their own. Keep them nearby and leave this screen open."
  - `ota:versionChangeVerifying`: "Verifying your glasses…" (unchanged)
  - `ota:versionChangeVerifyingMessage`: "Checking the installed version. Keep your glasses nearby and connected."
  - Removed `ota:versionChangeRestarting` ("Installing a different version…") and `ota:versionChangeKeepNearby`; the detour page no longer repeats the "restart twice / up to 2 minutes" line, which contradicted the new duration copy and the 180 s watchdog.

## Tests

- `otaInstallCoordinator.test.ts`: new describe block covering the connected detour wait → `reinstalling` (the hardware regression), sid change not entering `verifying`, physical reconnect → `verifying` → non-target build → `reinstalling` → reconnect → target → converged/null, and a reconnect that goes dark again before reporting a version.
- `useMentraLiveOta.test.tsx`: `reinstalling` phase maps to its own screen with finish blocked; `verifying` follows on the next snapshot.
- `bun run test -- otaInstallCoordinator useMentraLiveOta shared-flow progress`: all green. `tsc --noEmit` clean for `mobile`.
- Not verified on hardware in this PR (copy/mapping only); the good-case sequence on unit 023B should now read "Reinstalling glasses software…" until the target build reports, then "Version Change Complete".

### DIFF
--- a/mobile/modules/engine/scripts/fixtures/public-ota-consumer.tsx
+++ b/mobile/modules/engine/scripts/fixtures/public-ota-consumer.tsx
@@ -41,6 +41,8 @@ function screenLabel(screen: MentraLiveOtaScreen): string {
       return "Updating"
     case "restarting":
       return "Restarting"
+    case "reinstalling":
+      return "Reinstalling"
     case "verifying":
       return "Verifying"
     case "complete":
--- a/mobile/modules/engine/src/facades/ota.ts
+++ b/mobile/modules/engine/src/facades/ota.ts
@@ -19,7 +19,11 @@ import type {
 } from "@mentra/bluetooth-sdk"
 import {isGlassesConnected, isGlassesReady, useGlassesStore} from "../stores/glasses"
 import {resolveOtaManifestUrl} from "../services/otaManifestUrl"
-import {otaInstallCoordinator, type OtaInstallSnapshot} from "../services/OtaInstallCoordinator"
+import {
+  otaInstallCoordinator,
+  type OtaInstallSnapshot,
+  type VersionChangePhase,
+} from "../services/OtaInstallCoordinator"
 import {startGlassesStatusProjection} from "../services/GlassesStatusProjection"
 import {startOtaService} from "../services/OtaService"
 import {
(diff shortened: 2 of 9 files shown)


PhilippeFerreiraDeSousa:
Sorry I don't want that

PR #4027, opened 2026-09-12
Relay managed WebRTC through phones on Android and iOS

aisraelov:
Managed WebRTC streams currently send Cloudflare's ingest URL directly to the glasses, requiring glasses internet Wi-Fi. This adds Android and iOS phone publishing so `ingest: "whip"` uses:

`Glasses → SoftAP/WebRTC → phone receiver → phone WHIP publisher → Cloudflare`

Both publishers adapt ASG's full-ICE WHIP negotiation and feed WebRTC decoded glasses video/audio instead of phone camera/microphone capture. The existing coordinator owns provisioning and subscribers across both connections. Retry tears down both peers and the hotspot, preserves the live input, and isolates stale attempt callbacks. Last-subscriber stop cancels setup/reconnect, retains failed cleanup for retry, and defers unreachable glasses shutdown until BLE reconnects. ACS and managed relay reserve the same hotspot; Android exposes cellular alongside the scoped network only for relay sessions.

SRT/RTMP retain direct glasses publishing. WHIP does not silently fall back to those protocols. Bundled Livestreamer **1.0.20** skips glasses Wi-Fi setup only for managed WebRTC; platform restreams and direct streams keep that check. The bundle targets the production backend. External source: [Livestreamer commit 076679e](https://github.com/Mentra-Community/Livestreamer-Miniapp/commit/076679e).

This also fixes the confirmed iOS call startup race in incident `rep_01M2BRR8HR3H61X7DGCZ9KQFQF`: the glasses SSID became visible before DHCP replaced the phone's old router address. Both ACS and managed relay now carry the BLE-reported gateway into the shared hotspot join and wait for a valid client address on that subnet; retries refresh the gateway. The incident's subsequent WHIP HTTP 500 was not reproduced. ACS state and diagnostic logs now retain the native receiver's rejection reason for the device retest.

Validation:
- Incident follow-up: 168 ACS/SoftAP/relay service tests and 17 shared-media Swift tests pass; engine build and both arm64 iOS device/simulator native builds pass. A simulator probe of the actual WHIP HTTP receiver negotiates send-only video/audio with full ICE and returns HTTP 201.
- 57 relay/coordinator tests and 93 existing SoftAP lifecycle/barrier tests pass; engine TypeScript build passes.
- 518 Android shared-media/ACS unit tests pass, plus the Bluetooth SDK compile check using the public SDK dependency configuration.
- 40 Swift media/audio-policy tests pass. Shared iPhone native module and ACS simulator consumer compile successfully; full Android and iOS app compilation also passes in CI.
- Livestreamer: 243 tests, backend/miniapp typechecks, production packaging and ZIP verification pass.
- Mobile and SDK frozen lockfile checks, release-family validation and 18 affected release-tooling tests pass.

Draft pending real Android/iPhone + glasses qualification: simultaneous local Wi-Fi/cellular media, audio/video sync, startup cancellation, both-leg reconnect, last-subscriber stop, screen-off/background streaming, and sustained thermal behavior. Existing background Bluetooth/SoftAP support is reused; the new outgoing media pipeline still needs this hardware validation. No Android device is attached to the development machine.


aisraelov:
bugbot run

aisraelov:
@cubic-dev-ai review this PR

aisraelov:
bugbot run

aisraelov:
bugbot run

aisraelov:
bugbot run

PR #4045, opened 2026-09-14
asg_client: make stream telemetry a per-stream runtime opt-in

yuriylevyt:
## Summary
- `ENABLE_PIPELINE_FPS_TELEMETRY` (7f6fe05182) is a compile-time constant, so the 1 Hz `[STREAM_QUALITY]` logs and `stream_status.stats` (the only BLE channel that carries `temperatureC`) cannot be turned on without a custom build. This PR keeps the constant as the build-time **default** and routes every gate through a new `StreamTelemetryPolicy`, which the phone sets per stream with `start_stream.telemetry` (compact key `tl`).
- Each `start_stream` re-evaluates the flag. A stream that does not send the key falls back to the default, so it never inherits telemetry from the previous stream.
- Gates switched: `WhipStreamingService` (inside `statsSweepEnabled()` and the stats sweep), `PeriodicStreamMetricsReporter`, `RtmpStreamingService`, `SrtStreamingService`, `MediaManager.onStreamMetrics`. `ENABLE_CALL_PIPELINE_DIAGNOSTICS` still keeps the WHIP sweep running on its own, as before.
- The flag is applied after the `controllerProbeVersion` gate and logged as `[STREAM_STARTUP] telemetry=<bool>`.
- Default stays off, so phones that do not send the key see no behaviour change.

## Why
The temperature is what we use to watch the glasses thermal plateau during long sessions. Debug and QA builds of our app want it, production builds do not. A per-stream opt-in lets the phone decide without a firmware rebuild.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest` on this branch (dev): 1095 tests, 0 failures, including 8 new in `StreamTelemetryPolicyTest` (default matches the constant, set/reset, absent key, full key, compact key, full key wins over compact, null payload, absent key resets to default)
- [x] On Mentra Live hardware, with the same change on a `staging`-based `asg_client.thirdparty` build: `start_stream` with `"telemetry": true` logs `[STREAM_STARTUP] telemetry=true`, `[STREAM_QUALITY]` runs at 1 Hz, and the phone receives `stream_status.stats.temperatureC` (69.9–70.5 °C over a 35 min 1080p/5 fps WHIP session)
- [ ] Not tested on hardware from this exact dev-based build (the dev build needs `controllerProbeVersion`, which our phone SDK does not send yet)
- [ ] Not tested: RTMP/SRT on hardware

https://claude.ai/code/session_014LMeRXP469kwBVz58qc2v1

---
## Summary by cubic
Makes 1 Hz stream telemetry (`[STREAM_QUALITY]` logs and BLE `stream_status.stats`, which carries SoC temperature) a per-stream runtime opt-in surfaced through the Bluetooth SDK, so phones can turn it on without a custom firmware build. The build default stays off.

- Each `start_stream` re-evaluates the flag via `StreamTelemetryPolicy`; the opt-in key is `telemetry` (compact `tl`), and streams that don't send it fall back to the build default instead of inheriting from the previous stream.
- The flag is applied after the `controllerProbeVersion` gate and logged as `[STREAM_STARTUP] telemetry=<bool>`; `ENABLE_CALL_PIPELINE_DIAGNOSTICS` still runs the WHIP stats sweep on its own.
- SDK request models (TypeScript, Kotlin, iOS) expose `telemetry`, forward explicit true/false over the wire, omit it when unset, and prefer the full key over the compact key.
- The TypeScript native bridge now guards `startStream` via `bindNativeMethod`, drops a nonexistent `startExternallyManagedStream` wrap that crashed module load, and forwards `authToken`.
- Added unit tests on the glasses and SDK sides covering omission, explicit values, key precedence, null payload, and reset.
- Only WHIP was validated on hardware; RTMP/SRT paths were not exercised.

<sup>Written for commit 20e1cdbdfea8d73fd52957f681732639e39b3af5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/4045?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aisraelov:
Could you also expose `telemetry?: boolean` in the public Bluetooth SDK stream request, including the TypeScript, Kotlin, and Swift models? Preserve omission so the glasses use their build default, and forward explicit `true`/`false`. Please add serialization tests covering all three cases. That would let debug/QA callers opt in through the normal API.

PhilippeFerreiraDeSousa (approved):

Verdict: Approve.

Checked:
- Full 18-file diff at 20e1cdbdfea8d73fd52957f681732639e39b3af5 against origin/dev; root/mobile/ASG AGENTS.md and CLAUDE.md; Mentra Live spec, thermal characterization, relevant notes/agents/docs, and base-branch telemetry/lifecycle history.
- ASG policy, command admission/replacement, WHIP/RTMP/SRT scheduling and cleanup, MediaManager emission; TypeScript bridge, Kotlin/Swift request serialization, BLE forwarding, and added tests.
- Commands serialize on the main thread and stop existing publishers before applying the next policy. Volatile reads provide cross-thread visibility. Omission resets the next accepted start to the build default; lifecycle statuses and independent WHIP diagnostics remain intact. Native controllers still add the required probe fields.
- Ran bun test ./mobile/modules/bluetooth-sdk/src/__tests__/streamRequestPayload.test.ts: 6 passed. git diff --check origin/dev...HEAD passed. Reviewed Java/Kotlin/Swift tests without running them. Native builds were skipped: StreamPackLite is uninitialized and mobile dependencies are absent. No local hardware testing; the description's staging-based WHIP result does not qualify this exact head or RTMP/SRT. Checkout unchanged.

Existing comments:
- aisraelov #5667992470: agree that the public SDK must expose optional telemetry and preserve absent/true/false. Addressed by the TypeScript, Kotlin and Swift models and serialization tests.
- Cursor inline #4011662756 and its review at fe89948: agree the nonexistent startExternallyManagedStream binding could crash import. Addressed: that wrapper is absent and startStream uses bindNativeMethod.
- Cursor summary #5667216294: agree with default-off compatibility and the global-state caveat. The current serialized, single-publisher lifecycle makes the shared switch reasonable.
- No Codex connector, cubic, or github-actions comments/reviews were returned by the paginated comment endpoints and reviews query. Cubic's description summary broadly matches the head.

Defects: No blocking defects found. Coverage is strongest for parsing/serialization; actual reporter scheduling, stream replacement, and all-protocol device behavior are not established by the new unit tests.

## Is this change needed?

Yes. Existing WHIP thermal characterization documents a real long-session measurement need. The compile-time gate requires custom glasses builds to expose temperature over BLE; a phone-selected optional boolean reuses the existing stats channel without dependencies or another reporting system. Base commits 7f6fe05182 and 5e59fdc5f6 deliberately suppress routine BLE telemetry and separate logcat diagnostics; this change preserves those defaults. Engine's separate cloud/miniapp stats filter also remains in place, so this enables SDK consumers rather than automatically enabling downstream fanout. A firmware-only change would leave normal SDK callers unable to opt in.

Approval is warranted: the scope is useful, additive and proportionate, and the reported import defect is fixed. Native/device qualification remains a validation limitation, not a claimed result.

PR #4074, opened 2026-09-16
Ship Mentra Call 2.1.13 SoftAP recovery and ACS diagnostics

nic-olo:
## Summary
- Hold ACS across a glasses-hotspot drop, fence SoftAP attempts, and wait for leave + ingest + hotspot_off before the host reports settled.
- Apply the SoftAP video policy (min 1.2 / start 2.0 / max 2.5 Mbps) and origin-tagged ACS/WHIP quality traces.
- Bundle Mentra Call 2.1.13, including Wi-Fi-off preflight and the recovery-aware leaving path. Companion: Mentra-Community/Mentra-Call#32.

## Test plan
- [ ] Start a SoftAP call with phone Wi-Fi off — structured `SOFTAP_WIFI_DISABLED`, no hotspot start
- [ ] Start and Join a live Teams meeting; confirm `softap_video_policy` and `acs_bwe_sample` traces
- [ ] Leave during connecting — teardown barrier completes, no leftover default route
- [ ] Drop SoftAP mid-call — ACS stays up, ingest reconnects
- [ ] Confirm bundled zip is `com.mentra.call-2.1.13.zip`


PhilippeFerreiraDeSousa (changes requested):

Request changes.

### Checked
- Supplied checkout 6426a52f96559ad6f86e680215ceb4c097269191; full `git diff origin/dev...HEAD` (four files), root/mobile AGENTS.md and CLAUDE.md, surrounding runtime/transport/ACS/WHIP lifecycle, bundled Call 2.1.13 background code and manifest, relevant notes and recent dev commits.
- GitHub reports this PR already merged at a53b63c (merge 545ed70); origin/dev contains that merge. This verdict covers the supplied local follow-up delta, not an assertion that GitHub's merged head contains these fixes.
- Separate Bun suites: MiniappLiveness 7, SoftapCallTransport 118, LocalMiniappRuntime.softap 27, AcsMeetingService 82: 234 passed. Diff whitespace check passed. In-memory probes reproduced both findings below. No native build/device soak; existing tests do not cover native rollback or the lobby recovery completion contract.

### Findings
- **P2 — Restoring the listener does not restore recoverability:** `mobile/modules/glasses-media/ios/LocalWhipIngestSource.swift:88-96` restores the old listener but still rejects. That rejection reaches `SoftapCallTransport.ts:841-842`, leaves scoped Wi-Fi, turns off the hotspot and sets phase to failed; a subsequent recover rejects NOT_RECOVERABLE (:514). The runtime then emits a terminal error (:4313). The probe confirmed both outcomes. This does not fix the reported mid-call failure. Make recovery transactional across the native source and transport: retry transient binds within the existing recovery budget, fence retired-listener callbacks, and report success only after a usable receiver is republished and a fresh frame arrives. Add failed-bind-then-success and cancellation coverage.
- **P2 — Send an explicit recovery completion before removing metadata:** `mobile/modules/engine/src/services/LocalMiniappRuntime.ts:4287-4294` clears the timestamp, so both completion events omit recovery; finally clears recoveryActive without publishing it. Bundled `background/index.js:6951-6952` preserves its previous recovery unless ACS is connected AND media is live. Media can recover while ACS is still in the lobby (`AcsMeetingService.ts:829-831`), so the miniapp retains active:true and reconnecting state until admission. An extracted-consumer probe confirms this. Publish recovery.active=false for the completed generation before dropping the deadline/metadata; test completion in both lobby and connected states.

### Existing comments
- github-actions inline 4027434016 (duplicate replayStreamControlReady): valid historical compile defect, fixed; current SGCManager has one requirement and one default implementation.
- Cursor inline 4027483849 (iOS failed rebind): agree with the failure outcome; the local rollback remains incomplete as detailed above.
- Cursor inline 4027483873 (whole-call ping hold): agree; addressed by joinInFlight becoming false after transport.start and by the liveness test.
- Cursor inline 4027483901 (stale recovery deadline): agree metadata should be retired, but disagree that active:false plus a future deadline alone keeps this bundled UI reconnecting; it checks active. The replacement omission introduces the lobby completion problem above.
- Cursor summary 5699455795 and its formal review: describe the earlier broad change and repeat those three findings; risk merits lifecycle checks, not automatic rejection.
- github-actions orchestrator 5699465706 and review summary 5699542173: agree the duplicate declaration is resolved; their blanket three-open-findings verdict is stale for the liveness fix. The standards approval explicitly inspected an empty diff, so it does not validate this delta.
- github-actions build helper 5699453585 and build-ready notice 5699491003: informational evidence for a53b63c, not native/device validation of 6426a52.
- github-actions triggers 5699474927 and 5699784642, and its empty formal review: no technical assertion to assess. No Codex connector or cubic comments were returned.

### Is this change needed?
Yes: bounded join liveness and correct media recovery fit the existing shared SoftAP transport and established iOS background/gallery architecture. The old iOS-impossibility spike is explicitly superseded; recent dev already ships Call 2.1.13 recovery and diagnostics. This small follow-up adds no dependency and should repair that implementation. Keep one recovery owner and an explicit completion contract; the listener-only patch is insufficient. The older device-soak plan still expects terminal hotspot loss and should be aligned with recovery behavior.

Verdict: Request changes for the two reproducible lifecycle/consumer-contract defects, not merely unresolved bot comments.

PR #4079, opened 2026-09-16
Fix miniapp-cli dev watcher for Windows path separators

### LINKED ISSUE #3888: miniapp-cli dev: background edits never respawn the JSContext on Windows (path separators)
## Summary

On Windows, `miniapp-cli dev` never respawns the background JSContext. Every edit under
`src/background/` broadcasts `reload` (WebView only) instead of `respawn-bg`, so the phone keeps
running the previously installed bundle while the terminal reports activity and `dist/` is correctly
rebuilt.

The failure is silent and looks like a working dev loop. It cost us about an hour of debugging a
stale bundle before we noticed the terminal line was `reload src\background` rather than
`respawn-bg src/background/index.ts`.

## Cause

`src/dev-server.ts` decides which layer changed with forward-slash string tests:

```ts
const touchedBackground =
  filename.startsWith("src/background/") || filename.includes("/src/background/")
```

Windows `fs.watch` reports `src\background\index.ts`, so `touchedBackground` is **always false** and
`nextType` is always `"reload"`.

The same assumption breaks the watcher's skip list:

```ts
const isUnder = (filename: string, name: string): boolean =>
  filename === name || filename.startsWith(`${name}/`) || filename.includes(`/${name}/`)
```

so `isUnder(filename, "dist")` never matches either, and build output is not filtered out of the
watcher on Windows.

## Reproduction

1. On Windows, `bunx @mentra/miniapp-cli dev` in a two-layer project.
2. Install on a phone by scanning the QR.
3. Edit anything under `src/background/`, e.g. add a `console.log` at the top of the handler.
4. Observe the terminal prints `reload src\background` (note the backslash).
5. The new log line never appears in the forwarded `[MentraJS]` output — the phone is still running
   the old bundle, even though `dist/background/index.js` has a fresh mtime and contains the change.

## Suggested fix

Normalise the separator once at the watcher boundary, before any path test:

```ts
const norm = (p: string): string => p.replace(/\\/g, "/")

const isUnder = (filename: string, name: string): boolean => {
  const f = norm(filename)
  return f === name || f.startsWith(`${name}/`) || f.includes(`/${name}/`)
}

const watcher = watch(options.watchDir, {recursive: true}, (_event, rawFilename) => {
  const filename = rawFilename === null ? null : norm(rawFilename)
  if (!filename) return
  // ...unchanged
})
```

Verified against a local copy of `@mentra/miniapp-cli@0.1.0-dev.0`: after this change a background
edit prints `respawn-bg src/background/index.ts` and the phone picks up the new bundle without a
reinstall.

## Environment

- `@mentra/miniapp-cli` 0.1.0-dev.0, `@mentra/miniapp` 3.1.0-dev.27
- Host: Windows 11, Bun 1.x
- Device: Mentra Live, MentraOS app `staging.20260808.*`

## Note

The project README we inherited advised "delete the entry in Miniapp Developer Settings and rescan to
pick up background changes". That workaround is a correct observation of this bug, and it may be
worth checking whether the docs recommend it anywhere — on macOS/Linux `respawn-bg` works and the
advice is unnecessary.

Hayyan612:
Fixes #3888

### Scope

`miniapp-cli dev` decides whether a watcher event needs a background respawn, a WebView reload, or nothing, by matching the reported filename against `/`-separated paths. On Windows, `fs.watch` reports `src\background\index.ts`, so:

- edits under `src/background/` broadcast `reload` instead of `respawn-bg`, and the phone keeps running the old background bundle
- `isUnder(filename, "dist")` and the rest of the skip list never match, so build output is not filtered

### Change

- Move the existing matching into an exported `classifyWatchEvent(filename, separator = path.sep)` that returns `"ignore" | "reload" | "respawn-bg"`. The checks and their comments are unchanged.
- Normalize the platform separator to `/` before matching. When `path.sep` is `/`, the filename is passed through untouched, so macOS and Linux behave exactly as before.
- `onWatch` calls `classifyWatchEvent` and keeps its debounce and `respawn-bg` precedence logic as it was.

### Test evidence

New `classifyWatchEvent` tests in `src/dev-server.test.ts` cover POSIX and Windows paths (the separator is a parameter, so the Windows cases run on any OS), plus a check that a literal backslash is left alone where it is not the separator.

- Before the fix (classification moved into the function, no normalization): the POSIX cases and existing tests passed, and both Windows cases failed with the symptoms from the issue, `src\background\index.ts` classified as `reload` and `dist\background\index.js` as `reload` instead of `ignore`.
- After the fix: `bun test src/dev-server.test.ts` 12 pass, and the full `bun test src` in `sdk/miniapp-cli` 136 pass.
- `tsc --noEmit` reports no errors in `dev-server.ts` or its test. The two existing errors in `src/lan.test.ts` are the same on `dev` without this change.

I tested on macOS, not on a Windows machine or with a phone, so the Windows coverage is the separator-parameterized unit tests rather than a live `dev` session.

---
## Summary by cubic
Fixes #3888 so `miniapp-cli dev` handles Windows path separators in watcher events. On Windows, `fs.watch` reports `src\background\index.ts`, but the old matching only checked `/`-separated paths, so background edits triggered `reload` instead of `respawn-bg` and `dist/` build output was never filtered. The watcher now normalizes the platform separator to `/` before classifying events; macOS and Linux behavior is unchanged.

**Tests**
- Adds `classifyWatchEvent` unit tests covering POSIX and Windows separators, plus a check that a literal backslash is left alone on POSIX.

<sup>Written for commit f24a7769e30c8cab649ecef7232140b58d9787fb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Mentra-Community/MentraOS/pull/4079?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


PhilippeFerreiraDeSousa (approved):

- Checked: head f24a7769e30c8cab649ecef7232140b58d9787fb, PR description, full origin/dev...HEAD diff, sdk/miniapp-cli/src/dev-server.ts and dev-server.test.ts, dev.ts integration, package exports/config, AGENTS.md/CLAUDE.md, issue #3888, CLI README/public docs, notes/ and agents/ searches, and recent base history.
- Validation: git diff --check passed. Bun 1.4.0 bun test src: 111 passed; three suites failed to load because qrcode, typescript, and jszip are absent. Without installing dependencies, 51 in-memory assertions against the transpiled head classifier and base logic passed, covering both reported Windows failures, POSIX parity, default separator, nested paths, ignore rules, and literal POSIX backslashes. No Windows/phone integration test or compile build was run; the reported 136-test pass was not independently reproduced.
- Existing comments: agree with Cursor Bugbot summary #5703810374: normalization fixes classification while preserving debounce and respawn precedence; low risk is justified by the diff. Agree with the cubic summary embedded in the description for the same reasons. The paginated issue/inline endpoints and reviews list contain no other comments or reviews, including none from Codex connector or github-actions.
- Defects: none introduced found. dev-server.ts:134 normalizes once before every match; :264 delegates without changing pending-event precedence, suppression, cleanup, or error handling. The helper is pure, adds no shared mutable state or dependencies, and preserves POSIX behavior and the wire protocol. The new tests cover the changed behavior; no corrective patch is needed.

### Is this change needed?

Yes. Issue #3888 reports a real Windows stale-background failure, and the base logic demonstrably misclassifies backslash paths and fails to filter build output. Current docs and the two-layer architecture retain this watcher; recent base changes add optional hot-reload disabling and USB support rather than replace it. The macOS/Linux recommendation concerns mobile development and does not make this CLI fix unnecessary. A normalization-only edit would be shorter, but extracting the existing rules into one testable helper adds little maintenance cost and avoids duplicated matching.

Verdict: Approve. This is a small, coherent repair to the existing design with focused regression coverage and no identified compatibility regression.

PR #4086, opened 2026-09-17
Add staging-to-dev reconciliation skill

PhilippeFerreiraDeSousa:
Adds a repository skill for reconciling staging into dev while preserving staging behavior and existing dev-only changes. It directs agents to use the matching published dev prerelease when staging requires a newer API, regenerate dependency locks, and validate consumers without compatibility workarounds for obsolete pins.

The skill includes isolated-worktree guidance, publication evidence, and distinct prepared/merged/published outcomes. It preserves explicit invocation via `agents/openai.yaml` and uses portable inputs rather than tool-specific positional arguments or local paths.

Validation: skill-creator `quick_validate.py` and `git diff --check` pass. Documentation and skill metadata only; no application behavior changes.


PhilippeFerreiraDeSousa:
Actually I don't want that sorry

PR #4112, opened 2026-09-19
Install PR iPhone builds directly from Slack

PhilippeFerreiraDeSousa:
Downloading an IPA in Slack on an iPhone only saves a file. PR build notifications now include **Install on iPhone**, linking directly to Apple's installation confirmation for the existing ad hoc signed IPA. **Install via Safari** remains beside it as a fallback; the GitHub PR comment links to that HTTPS installation page.

Slack groups Android, iOS and macOS in matching rows above the glasses OTA details, with build logs together in the footer. Those rows use structured rich-text links: the webhook mrkdwn parser escaped the direct itms-services URL as literal text, while Slack rich-text links preserve it.

The publisher generates an XML manifest and mobile HTML page on the existing HTTPS artifact CDN. Both are included in the verified publication receipt and published before the receipt. Before advertising the links, the notifier downloads these small files and verifies their decoded byte lengths, SHA-256 hashes and MIME types, including when CDN compression omits or changes Content-Length. Publication retries preserve the original IPA and installation files; older receipts retain their download-only links.

iOS CI now uses setup-bun v2 with Bun 1.4.0, matching coordinated mobile builds. It reuses an installed version or downloads from GitHub Releases instead of the failing bun.sh latest alias. Signed archives retain their full signing commands and report public keychain metadata on failure.

Devices must already be included in the provisioning profile. No new certificate, signing profile, service or secret is required. This changes PR distribution only; coordinated release and TestFlight workflows are unchanged.

Validation:
- 40 focused tests passed for artifact generation, publication and notification, including actual Brotli HTTP responses. The final direct-link change passed all 19 affected artifact/notification tests, verifying the exact manifest URL, Safari fallback, failed-download suppression and legacy receipts.
- Actionlint and `git diff --check` passed. The page was visually checked at 390 × 844 with the install button visible before scrolling.
- Run 35418057812 completed and its ready Slack post includes the verified Safari installer. Both text assets were checked against their published receipt hashes.
- Philippe confirmed the Safari installation flow works, then confirmed the [direct-link test in Slack](https://mentra-labs.slack.com/archives/C0C1V60KLA0/p1789790030871569?thread_ts=1789788832.009709&cid=C0C1V60KLA0) works on iPhone (PR build 302018635). CI notifications use that tested direct URL format as the primary iPhone action. The layout update passed all 19 affected tests, including ready and unavailable platform rows.

Latest verification: [run 35424996241](https://github.com/Mentra-Community/MentraOS/actions/runs/35424996241) passed on final head `0de315d`. The signed archive succeeded on its first attempt; export, publication, and notification all passed. The IPA/ZIP availability and sizes, and the decoded manifest/page sizes and SHA-256 hashes, were checked against the receipt (native build **302018789**). All PR checks are green.

The [final Slack post](https://mentra-labs.slack.com/archives/C0C1V60KLA0/p1789797872760999) has matching Android, iOS and macOS rows and preserves the direct `itms-services` link as a link, rather than escaped text. The earlier framework-signing error did not reproduce on the two subsequent successful archive runs; its underlying cause is not established, and the signing diagnostics are retained.


PhilippeFerreiraDeSousa:
Fixed the incomplete Slack notification: Cloudflare served the XML manifest and HTML installation page with Brotli compression and no `Content-Length`. The notifier incorrectly treated that missing header as zero bytes. Both published files were intact.

The notifier now fetches these small text artifacts and verifies their decoded byte length, SHA-256 and MIME type against the receipt. IPA and ZIP availability still uses HEAD. All 40 focused tests pass, including actual Brotli HTTP responses with absent or compressed Content-Length, plus rejection of altered content.

Live verification against run 35416233583 passed for the 790-byte manifest and 2,859-byte page, with exact receipt hashes. The existing [Install on iPhone page](https://artifactscdn.mentraglass.com/Mentra-Community/MentraOS/releases/pr-builds/mentra-ios-install-pr-4112-f36660443891b904c978dced5cee00009cf7c92a-35416233583-1.html) is available now. A new CI run will exercise the corrected Slack notification. Installation on a physical iPhone remains to be verified.

PR #4114, opened 2026-09-19
Keep legacy OTA sessions open until the release is verified

PhilippeFerreiraDeSousa:
Legacy glasses can briefly report ASG 37 after the firmware reboot before switching to ASG 39. An empty recovery manifest, or an MTK-only offer removed by the pending-reboot filter, could make the app show “Update complete,” return home, and then offer the remaining ASG/BES updates after 39 selected the coordinated release manifest.

Keep the approved OTA flow open through that handoff and check the release pin before reporting completion. Make the completion decision after filtering updates. Await and share fresh version requests so concurrent checks cannot compare cached versions while the refresh is pending. Failed version verification remains retryable within the approved session.

The additional handoff wait is specific to legacy glasses because ASG builds below 39 use the device-reported recovery manifest (or the legacy production fallback), while builds 39 and above use the selected release pin. Finishing the recovery manifest only finishes the migration stage; the release manifest may still contain APK/BES updates. Modern glasses already check that release pin, so they do not need this extra manifest-switch wait. The guard runs only during an approved update chain with no effective updates left and a legacy manifest selected; deployments that explicitly disable legacy fallback bypass it. “Legacy” here refers to the glasses' ASG build, not the phone app version. The fresh-version request correction applies to both legacy and modern checks.

The supplied production logs show ASG 36 → 37, legacy MTK/BES updates, a no-update result at 21:07:53, ASG 39 at 21:08:01, and then the remaining APK/BES updates to release 3.1.1. The initial reported `3.2.1 → 3.1.1` screen is outside the retained production-app log window; its exact source remains unconfirmed.

Validation:

- 99 Jest tests across OTA checking, chaining, background checks, and screen presentation passed.
- 16 Engine OTA hook tests and 10 chain/manifest/release policy tests passed.
- The pending-MTK-reboot regressions reproduce the premature completion in both the service and the real hook before the fix; both pass afterward and verify automatic continuation at ASG 39.
- Full mobile TypeScript compilation, changed-file formatting, and `git diff --check` passed. Corrected fetch-mock typing that caused the previous CI test job to fail compilation.
- Old-pair hardware replay is still required: verify one uninterrupted flow across rescue reboot, automatic continuation through the release updates, and completion only after the final check.


PhilippeFerreiraDeSousa:
Addressed [review 5254973356](https://github.com/Mentra-Community/MentraOS/pull/4114#pullrequestreview-5254973356) in d7170e29c7.

The legacy handoff guard now uses the final effective update availability, after the pending-reboot MTK filter. An MTK-only offer filtered to empty keeps the approved flow on “Finishing” until ASG 39 arrives and the release manifest is checked. Added service and real-hook regressions that fail before this fix and pass afterward, including automatic continuation into the remaining APK/BES updates.

Also corrected the fetch-mock types that blocked the previous CI test job. Local validation passed: 99 Jest tests, 16 Engine hook tests, 10 OTA policy tests, full mobile TypeScript compilation, formatting, and diff checks. Old-pair hardware replay remains pending.

PR #4136, opened 2026-09-22
Add Mac PR installation and CI OTA results in admin

PhilippeFerreiraDeSousa:
PR builds need a reusable Mac installer, and hardware tests need to consume the same app and OTA artifacts that CI publishes. This adds a signed native installer to the existing company Mac download, a registered day-one OTA routine, and private recorded results in admin.

- Sign, notarize and staple the native installer; retain it for subsequent **Install on Mac** links. Verify the immutable publication receipt, archive, app signature and packaged OTA pin before replacement. Preserve the iOS app identity, account state and reusable permissions; first-use OS grants remain host setup.
- Share one durable app-ownership protocol across installers and registered routines. Active or unknown OTA operations, process interruption and persistence failures keep installers excluded. Only recovery of the same frozen run can reconcile pending work without resending it; settled completion releases the reservation.
- Freeze the PR app, target manifest, source profile, January artifacts and return artifacts. Default to full MTK OTA for setup/restoration, with an explicit January userdata wipe exposing bundled ASG27; flashing remains an explicit alternative. Record English chapters, screenshots and continuous app video, with separate setup, customer, recovery and return evidence.
- Export actual phase-aware MTK, BES, ASG and APK comparisons to authenticated admin results, with private media streaming and chapter seeking. Freeze terminal state and its exact journal prefix; retain the original worker result and publish later recovery as a linked immutable generation. Preserve original test failures after successful recovery. Fix the January compatibility issues found during supervised testing: stale parser bytes at BES raw-mode entry and ASG27's two-part version response.

Validation at `b13839c1a2`: 326 harness tests (2,854 assertions) and 11 admin viewer tests passed with CI's Bun 1.3.14; harness/admin typechecks, formatting and diff checks passed. The new regression uses actual request consumption and lifecycle recovery, then prepares and uploads both result generations through the real publisher and local ingest service. It verifies one dispatch, unchanged original failure, recovered return state, idempotent reupload and rejection of replacement under the original ID. Snapshot corruption, partial later recovery and interrupted snapshot publication are covered. These are offline checks; fresh GitHub checks are pending.

Earlier installer validation passed 50 native installer/distribution tests and compiled the shipping installer. The existing production ownership subprocess tests remain in the current harness suite. No firmware or installer code changed in this review follow-up.

Supervised hardware evidence separately verified the full January downgrade over Wi-Fi and active bundled ASG27. A later customer update reached the selected MTK/BES/ASG versions and exact APK hash, with all fourteen return checks passing. Its original customer recording failed during an intermediate reboot and remains failed. Subsequent fixture enrollment and its recording passed; authenticated Safari playback and chapter seeking of the actual historical recording now work inline.

**Draft: complete registered CI OTA qualification remains open.** The next run must consume artifacts from the current PR head, complete setup/customer OTA/return, and publish its own result to admin. No CI request is represented as a passing hardware run. Fresh CI is pending. Nightly activation and PR result comments remain follow-up work; raw evidence, firmware and credentials stay outside Git.

The reviewer confirmed the GitGuardian findings are synthetic test fixtures, not live credentials. Provider reclassification remains pending.


PhilippeFerreiraDeSousa:
Addressed both P2 findings in c3f22ae7c8.

- **App ownership:** native and repository installers now use the same atomic lease as the runner, held through quit, replacement, launch and cleanup. The existing Open button is covered too. Interrupted installs retain ownership for explicit recovery. Registered runs require the canonical lease, and the shared helper is included in package/source/evidence pins. Cross-language tests prove that an active worker prevents native installation and a native installer prevents a worker.
- **Firmware evidence:** registered exports and their sanitized summary now carry the four allowlisted firmware/APK comparisons, with expected/actual values, status and an optional structured phase in the canonical record. Admin displays that phase. Known source versions can be observed without changing the selected target or relaxing identity, idle or update eligibility checks. The regression uses the actual collector under a consumed lifecycle, then ingests and renders the result; failed final checks remain visible beside successful teardown/return, with private identifiers excluded.

Validation: 259 offline tests passed (124 runner, 16 JS installer, 48 Swift installer/distribution, 48 Python packaging/import, 9 admin viewer, 14 Core API), plus harness/admin TypeScript checks. Core API tests used a temporary local tsconfig alias for the checkout's existing broken workspace symlink. Independent review found no remaining blocker in these fixes.

The PR remains draft for the separately documented complete device qualification. This change does not claim a new hardware pass. Fresh CI is pending; the previous iOS build failed its cached Xcode attempt with exit 65 and no published underlying diagnostic.

